PR #8, opened 2025-06-30
[WIP][Stacked] Use go-ethereum via FFI to call Go functions to parse the raw transaction into SignablePayload in the new format

prasanna-anchorage:
# How to use this

`make -C src parser`
```
grpcurl -plaintext -d '{"unsigned_payload": "0xf86c808504a817c800825208943535353535353535353535353535353535353535880de0b6b3a76400008025a028ef61340bd939bc2195fe537567866003e1a15d3c71ff63e1590620aa636276a067cbe9d8997f761aecb703304b3800ccf555c9f3dc64214b297fb1966a3b6d83", "chain":"CHAIN_ETHEREUM"}'  localhost:44020 parser.ParserService/Parse
```

### DIFF
--- a/src/chain_parsers/visualsign-goethereum/build.rs
+++ b/src/chain_parsers/visualsign-goethereum/build.rs
@@ -109,24 +109,8 @@ fn main() {
 
 pub type GoInt = ::std::os::raw::c_longlong;
 
-#[repr(C)]
-pub struct EthTransaction {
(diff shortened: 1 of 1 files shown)


prasanna-anchorage:
this isn't relevant anymore with visualsign-ethereum, FFI is still an option but we'll consider this after binaries are split and in special cases

PR #134, opened 2025-12-24
[ wip ] feat: WASM bindings for Ethereum / Solana / Tron transaction parsing

lukehalasy-anchorage:


lukehalasy-anchorage:
> I don't think spending time on WASM bindings is a good idea right now with visualsign-parser. The main value proposition of visualsign-parser is that it's executed in a TEE, and wasm adds a lot of extra complexity and compile time risk here. You can already run it locally and point your client/extension to the local instance with exactly the same binary that you can reproducibly build. If we release WASM binaries out, then we have to also accept all the complexity of wasm/wasi, whatever. The client code already compiles to wasm with tinygo and riscv32 in this [PR](https://github.com/anchorageoss/visualsign-turnkeyclient/pull/2) and you can use it to verify what ran on TEE with absolutely minimal dependencies. By compiling to WASM you have to deal with every single quirk of upstream library and dependency.
> I'll point to this PR for future attempts at this on this repo because it doesn't fit with our goals with VisualSign. 

closing 👍

PR #169, opened 2026-02-06
Fix CLA workflow failing on comments with quotes

prasanna-anchorage:
## Summary
- Pass comment body via `env` var instead of direct `${{ }}` interpolation to avoid shell injection
- Replace `xargs` with `sed` for whitespace trimming since `xargs` treats quotes as special characters and fails on comments like `I'll`

Fixes https://github.com/anchorageoss/visualsign-parser/actions/runs/21718497730/job/62641178844

## Test plan
- [ ] Comment on a PR with a message containing single quotes (e.g. "I'll check this") and verify the workflow doesn't fail
- [ ] Comment `/approve-cla` on a PR and verify it still works correctly

### DIFF
--- a/.github/workflows/add-cla-user.yml
+++ b/.github/workflows/add-cla-user.yml
@@ -14,19 +14,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Debug event
+        env:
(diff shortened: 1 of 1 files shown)


lukehalasy-anchorage:
still need to verify this worked ( loading )

Issue #185, opened 2026-02-27
Split visualsign-solana binary


prasanna-anchorage:
### Describe the feature

Create Visualsign-Solana Binary as a separate deployable binary

PR #191, opened 2026-03-03
feat(images): add Containerfile for Solana parser_cli

### LINKED ISSUE #188: Add Containerfile for Solana parser CLI
## Summary

Add `images/parser_cli/Containerfile` using the same stagex hermetic build approach as existing images, targeting a Solana-only build of `parser_cli`.

## Approach

- Use `stagex/pallet-rust` base image, digest-pinned, matching `rust-toolchain.toml`
- Follow existing `cargo fetch` then `--network=none` build pattern
- Build with `--no-default-features --features solana` to produce a lean Solana-only binary

## Dependencies

- #186 Remove Ethereum-specific coupling from `parser_cli`
- #187 Refactor `parser_cli` to inline registry and add chain feature flags

shahan-khatchadourian-anchorage:
## Summary

- Add `images/parser_cli/Containerfile` using the same stagex hermetic build approach as existing images
- Targets `parser_cli` with `--features solana` — does not compile `visualsign-ethereum`
- Add `out/parser_cli` OCI Makefile target
- Add `parser_cli` to the `stagex.yml` CI matrix so the build is verified on every PR

## Verification

CI (`stagex` workflow) will build the image as part of this PR. The matrix entry runs the same `make out/parser_cli/index.json` → `docker load` → GHCR push pipeline as `parser_app` and `parser_host`.

## Notes

- `--no-default-features` is already in `CARGOFLAGS`; only `--features solana` is added at build time
- Final image uses `scratch` (fully static musl binary, consistent with `parser_host`)
- When #185 lands (dedicated `visualsign-solana` binary), the build target here would update from `parser/cli` to the new crate

Stacked on #190. Closes #188.

### DIFF
--- a/.github/workflows/stagex.yml
+++ b/.github/workflows/stagex.yml
@@ -36,6 +36,7 @@ jobs:
       matrix:
         target:
           - name: parser_app
+          - name: parser_cli
     steps:
       - name: Checkout sources
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,14 @@ out/parser_app/index.json: \
 	$(shell git ls-files images/parser_app src)
 	$(call build,parser_app)
 
+out/parser_cli/index.json: \
+	$(shell git ls-files images/parser_cli src)
+	$(call build,parser_cli)
+
 .PHONY: non-oci-docker-images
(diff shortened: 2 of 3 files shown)


prasanna-anchorage:
as discussed, this should be for parser_app and not CLI, I'm inclined to merge #190 and create a binary artifact that has everything but document how someone can make a smaller binary just focused to their chain.

PR #196, opened 2026-03-12
test: property-based and pipeline fuzz tests for Solana IDL parsing

shahan-khatchadourian-anchorage:
## Summary

- Adds `tests/fuzz_idl_parsing.rs`: property-based and roundtrip tests that call `parse_instruction_with_idl` directly, covering crash safety, discriminator dispatch, valid-input parse guarantees, defined struct types, and SizeGuard boundary enforcement
- Adds `tests/pipeline_integration.rs`: end-to-end tests through the full `transaction_to_visual_sign` stack, covering the IDL path, fallback path, named accounts, multi-instruction transactions, and pipeline-level property tests
- Adds `proptest = "1"` to `[dev-dependencies]` in `visualsign-solana/Cargo.toml`

### `fuzz_idl_parsing.rs` (17 tests)

**Property tests:**
- `fuzz_idl_parsing_never_panics` — 50/50 mix of valid-discriminator + random-arg-bytes vs fully random bytes; on `Ok`, asserts instruction name is non-empty
- `fuzz_valid_discriminator_random_args` — always uses a valid discriminator from the IDL; on `Ok`, asserts dispatch routed to the correct instruction by name
- `fuzz_decode_idl_data_arbitrary_input` — arbitrary strings must never panic in `decode_idl_data`
- `fuzz_defined_struct_types_never_panics` — IDLs with defined struct types in the `types` array; same 50/50 mix; on `Ok`, asserts correct dispatch
- `fuzz_valid_data_always_parses_ok` — the strongest assertion: generates borsh-correct bytes via `arb_bytes_for_type` and asserts `result.is_ok()` unconditionally; a failure here indicates a parser bug
- `fuzz_size_guard_vec_length_prefix` — valid discriminator followed by an arbitrary `u32` length prefix and a short trailing payload; verifies the SizeGuard budget (1232 × 24 = 29 568 bytes) is enforced cleanly

**Roundtrip tests (concrete):** no-arg, single u64, mixed primitives, `Option` Some/None, `Vec<u8>`, `Array<u32, 3>`, multiple instructions with distinct discriminators, defined struct arg

**SizeGuard concrete tests:** `Vec<u8>` with 10M-element claim, `Vec<u64>` with 100K-element claim — both must return `Err`, not panic

### `pipeline_integration.rs` (9 tests)

**Concrete tests:**
- `pipeline_idl_path_correct_data` — valid discriminator + correct args; asserts title contains `(IDL)` and arg values appear in condensed fields
- `pipeline_idl_discriminator_miss` — IDL registered but wrong discriminator; asserts IDL path attempted, expanded fields report parse failure
- `pipeline_no_idl_registered` — raw hex fallback; title is the program ID, no `(IDL)` marker
- `pipeline_named_accounts` — IDL account names appear with their pubkey values in expanded fields
- `pipeline_field_count_equals_instruction_count` — one `PreviewLayout` field per instruction, invariant holds
- `pipeline_multi_instruction_mixed_programs` — two instructions, one with IDL and one without; each takes the correct path independently

**Property tests:**
- `fuzz_pipeline_never_panics` — 50/50 valid/random discriminator mix through the full pipeline
- `fuzz_pipeline_field_count_invariant` — field count always equals instruction count regardless of discriminator validity
- `fuzz_pipeline_idl_path_taken_on_valid_discriminator` — when a valid discriminator is present, title always contains `(IDL)`

## Strategy design

`arb_idl_type()` generates primitives and single-level containers (`Vec`, `Option`, `Array`) plus nested combos (`Vec<Option<T>>`, `Option<Vec<T>>`), weighted 4:1:1:1:1:1.

`arb_bytes_for_type(ty, types)` recurses over `IdlType` to produce borsh-correct bytes. Vec is bounded to 0–2 elements and String/Bytes to 0–16 bytes so all generated payloads stay under `MAX_CURSOR_LENGTH` (1232 bytes) — keeping the `prop_assert!(result.is_ok())` guarantee unconditional.

Run with more iterations: `PROPTEST_CASES=5000 cargo test --test fuzz_idl_parsing`

### DIFF
--- a/src/chain_parsers/visualsign-solana/tests/fuzz_idl_parsing.rs
+++ b/src/chain_parsers/visualsign-solana/tests/fuzz_idl_parsing.rs
@@ -13,7 +13,9 @@
 //! More iterations: `PROPTEST_CASES=5000 cargo test --test fuzz_idl_parsing`
 
 use proptest::prelude::*;
+use solana_parser::solana::structs::{IdlInstruction, IdlType, IdlTypeDefinition, IdlTypeDefinitionType};
 use solana_parser::{decode_idl_data, parse_instruction_with_idl};
+use std::sync::Arc;
 
 const TEST_PROGRAM_ID: &str = "11111111111111111111111111111111";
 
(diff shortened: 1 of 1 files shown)


shahan-khatchadourian-anchorage:
@copilot is it possible to use Arbitrary for these tests?

PR #214, opened 2026-03-20
feat(networks): add fee-paying asset symbol to network definitions

pepe-anchor:
Add `fee_paying_asset_symbol` per network to the `define_networks!` macro and generate a `get_fee_paying_asset_symbol()` lookup function. Returns `Option<&'static str>` — `Some(symbol)` for known chains, `None` for unknown. Replaces hardcoded `"ETH"` / `"Test Asset"` in `convert_to_visual_sign_payload()` so non-Ethereum EVM chains display the correct symbol (e.g. POL, BNB, AVAX, MNT), and unknown chains cleanly omit the symbol.

## Breaking Changes (serialized output)
- **Known chains**: `AssetLabel` on AddressV2 fields changes from `"Test Asset"` to the correct chain symbol (e.g. `"ETH"`, `"POL"`, `"BNB"`)
- **Unknown chains**: `AssetLabel` is now omitted (was `"Test Asset"`), `Abbreviation` on AmountV2 is omitted (was `"ETH"`), and `FallbackText` on Value is the raw amount without a symbol suffix (was `"<amount> ETH"`)

These are intentional corrections — the previous values were placeholders or incorrect for non-Ethereum chains.

## Changes
- Extend `define_networks!` macro with `fee_paying_asset_symbol` field
- Add `get_fee_paying_asset_symbol(chain_id) -> Option<&'static str>` function
- Update `convert_to_visual_sign_payload` to use optional dynamic symbol
- Extract `format_ether(transaction.value())` to avoid redundant computation
- Update fixture expected outputs and integration test assertions

## Stack
| # | Branch | PR |
|---|--------|-----|
| 1 | `stack/wire-abi/01-native-asset-symbol` | #214 ← current |
| 2 | `stack/wire-abi/02-proto-abi-mappings` | #215 |
| 3 | `stack/wire-abi/03-abi-metadata` | #216 |
| 4 | `stack/wire-abi/04-cli-plugin` | #217 |
| 5 | `stack/wire-abi/05-remove-old-code` | #218 |


prasanna-anchorage (commented):
I don't think NativeAsset is a good term for this in 2026. Something more accurate would be FeePayingAsset? A network can have a very different asset that's "native" but not used for Fees because it uses stables for that or has more than one.

prasanna-anchorage (approved):

PR #224, opened 2026-03-26
test: add real-IDL structural validation tests

shahan-khatchadourian-anchorage:
## Summary

- Extracts `load_idl_from_env` from `fuzz_idl_parsing.rs` into `common/mod.rs` so all real-IDL test files can share it
- Adds `real_idl_validation.rs` with 4 deterministic structural invariant tests for production IDLs:
  - `real_idl_all_instructions_have_discriminators` — every instruction has an 8-byte discriminator
  - `real_idl_discriminators_are_unique` — no two instructions share a discriminator
  - `real_idl_instruction_names_are_unique` — no duplicate instruction names
  - `real_idl_idl_hash_is_stable` — `compute_idl_hash` is deterministic

These tests require no random input generation (unlike the proptest-based `real_idl_*` tests in `fuzz_idl_parsing.rs`). They assert invariants on the decoded IDL structure itself.

## Stacked on

- #204

## Test plan

- [ ] `IDL_FILE=<path> cargo test --test real_idl_validation` — 4 tests pass
- [ ] `cargo test --test fuzz_idl_parsing` — existing tests still pass (load_idl_from_env now imported from common)

### DIFF
--- a/.github/workflows/fuzz-solana.yml
+++ b/.github/workflows/fuzz-solana.yml
@@ -1,8 +1,11 @@
-name: Fuzz Testing
+name: "Fuzz Testing: Solana"
 
 on:
   pull_request:
     types: [opened, synchronize, reopened, labeled]
+
+env:
+  NIGHTLY_VERSION: nightly-2026-03-13
 
 jobs:
   fuzz:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
prasanna-anchorage: we lost the main branch here?

src/chain_parsers/visualsign-solana/src/core/instructions.rs
prasanna-anchorage: Out-of-bounds program_id_index silently skips the entire instruction. Out-of-bounds account indices silently omit accounts. The defensive logic is correct (no panic), but there's no tracing::warn! — operators get zero visibility into malformed input in  production. It'd be good to add some fields eventually for those but warning might be okay too for now

PR #226, opened 2026-03-27
ci: label-triggered stagex builds and cache/credential fixes

shahan-khatchadourian-anchorage:
## Summary

- **stagex.yml**: PR image builds now require the `stagex` label instead of auto-triggering on path changes. Push to main/tags with path filtering unchanged. GHCR credentials moved after build step so build runs without registry write access.
- **main.yml**: Disable `setup-rust-toolchain` built-in cache (`Swatinem/rust-cache`) which fails because `Cargo.toml` is in `src/`, not the repo root. The manual `actions/cache` step already handles caching correctly.

## Test plan

- [ ] Verify adding `stagex` label triggers the build on a PR
- [ ] Verify push to main with relevant path changes still triggers the build
- [ ] Verify GHCR push still works on push to main
- [ ] Verify `Continuous Integration` no longer shows `Cargo.toml` not found errors

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,7 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@fb51252c7ba57d633bc668f941da052e410add48 # v1.13.0
         with:
           components: clippy, rustfmt
+          cache: false
 
       - name: Cache Rust dependencies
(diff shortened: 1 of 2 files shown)


shahan-khatchadourian-anchorage:
Superseded by PR #232 which now incorporates these changes (commit 2bf5b2c).

PR #239, opened 2026-04-08
feat(solana): add Drift Protocol visualizer preset

prasanna-anchorage:
## Summary
- Adds IDL-based visualizer preset for Drift Protocol v2 (`dRiftyHA39MWEi3m9aunc5MzRF1JYuBsbn6VPcn33UH`)
- 249 instructions decoded via Anchor IDL from [drift-labs/protocol-v2](https://github.com/drift-labs/protocol-v2)
- Uses `LazyLock` to cache parsed 438KB IDL (parsed once, not on every call)
- Includes 4 unit tests: IDL loading, discriminator validation, unknown discriminator error, short data error

## Test plan
- [x] `cargo clippy -p visualsign-solana --all-targets -- -D warnings`
- [x] `cargo test -p visualsign-solana` (all 108 tests pass)
- [x] `make -C src test` (full workspace)

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/drift/config.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/drift/config.rs
@@ -1,0 +1,22 @@
+use super::DRIFT_PROGRAM_ID;
+use crate::core::{SolanaIntegrationConfig, SolanaIntegrationConfigData};
+use std::collections::HashMap;
+
+pub struct DriftConfig;
+
+impl SolanaIntegrationConfig for DriftConfig {
+    fn new() -> Self {
+        Self
+    }
+
+    fn data(&self) -> &SolanaIntegrationConfigData {
+        static DATA: std::sync::OnceLock<SolanaIntegrationConfigData> = std::sync::OnceLock::new();
+        DATA.get_or_init(|| {
+            let mut programs = HashMap::new();
+            let mut instructions = HashMap::new();
+            instructions.insert("*", vec!["*"]);
+            programs.insert(DRIFT_PROGRAM_ID, instructions);
+            SolanaIntegrationConfigData { programs }
+        })
+    }
+}
--- a/src/chain_parsers/visualsign-solana/src/presets/drift/drift.json
+++ b/src/chain_parsers/visualsign-solana/src/presets/drift/drift.json
@@ -1,0 +1,20139 @@
+{
+  "version": "2.162.0",
+  "name": "drift",
+  "instructions": [
+    {
+      "name": "initializeUser",
+      "accounts": [
+        {
+          "name": "user",
+          "isMut": true,
+          "isSigner": false
+        },
+        {
+          "name": "userStats",
+          "isMut": true,
+          "isSigner": false
+        },
+        {
+          "name": "state",
+          "isMut": true,
+          "isSigner": false
+        },
+        {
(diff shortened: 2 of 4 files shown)


shahan-khatchadourian-anchorage (changes requested):
Thanks for this — solid addition, and the structure is right. A few things to call out before it can land.

**What's good**
- Follows the IDL-preset pattern cleanly: `include_str!` + `LazyLock` to parse the 438KB IDL exactly once is the correct approach, and the file layout (`mod.rs` / `config.rs` / `drift.json`) matches the other presets.
- Graceful "Unknown Instruction" fallback on decode failure is consistent with the recent `spl_token` degrade-gracefully behaviour rather than erroring.
- Uses the `field_builders` helpers throughout instead of hand-building field structs, and the `Current`/`New` labeling on account/arg name collisions is a thoughtful touch.
- Sensible unit tests for IDL loading, discriminator shape (8 bytes), and the short-data / unknown-discriminator error paths.

**Blockers (all from the branch being ~168 commits behind `main`)**
This branch predates two refactors on `main`, so the test-plan claims hold against the old base but not current `main`. After rebasing:
1. **`config.rs` won't compile** — `SolanaIntegrationConfigData.programs` is now `BTreeMap`, not `HashMap` (changed for deterministic serialization). Switch the two `HashMap::new()`s to `BTreeMap`, matching `onre_app` / `exponent_finance`.
2. **Drop the `presets/mod.rs` edit entirely** — since #278 that file is auto-generated by `build.rs` from the directory listing (header says "do NOT edit"). Once rebased, `drift/` is auto-discovered and `DriftVisualizer` registers with no manual wiring.
3. **Test module needs `#[allow(clippy::unwrap_used, clippy::expect_used, clippy::panic)]`** — every other preset test module has it; without it `clippy -D warnings` fails CI.

**Worth folding in (overlaps with Copilot's open threads)**
- `named_accounts` as a `HashMap` gives non-deterministic field ordering in the output, which breaks stable metadata hashing — the fields aren't re-sorted downstream. Copilot also flagged this struct from a maintainability angle (it duplicates `SolanaParsedInstructionData::named_accounts`). Populating the existing `parsed.named_accounts` directly resolves both at once.
- A real end-to-end fixture test (decode one known instruction, e.g. `initializeUser` / `placePerpOrder`, to expected fields) would lock in the account-name mapping as the IDL evolves — currently nothing exercises the decode path against a real tx.

Also flagging for cross-PR awareness: #350/#351/#360 are other in-flight Solana presets built on current `main` (none touch `presets/mod.rs`), and #289 reworks the `solana-add-idl` scaffolding — worth rebasing onto and regenerating from, so this preset matches the current template.

Happy to take the mechanical fixes (1–3 + the `named_accounts` swap) on a rebased branch if useful.

shahan-khatchadourian-anchorage:
I've opened #366, which rebases this work onto current `main` and gets it building/linting/testing green. This branch is ~168 commits behind and no longer compiles against `main` (the `HashMap`→`BTreeMap` switch in `SolanaIntegrationConfigData`, the auto-generated `presets/mod.rs` from #278, and the removed `VisualizerContext::current_instruction()` API all broke it).

#366 keeps your implementation intact and applies only the mechanical fixes needed to land it — full details in its description. Leaving this PR open for now; feel free to review #366 or pull its changes back here, whichever you prefer.

PR #240, opened 2026-04-08
feat(solana): add Squads v4 Multisig visualizer preset

prasanna-anchorage:
## Summary
- Adds IDL-based visualizer for Squads v4 Multisig program (`SQDS4ep65T869zMMBKyuUq6aD6EgTu8psMjkvj52pCf`)
- Decodes nested inner instructions inside `vaultTransactionCreate` by parsing the `VaultTransactionMessage` wire format and dispatching to the full visualizer framework
- Fixes `data_len` field in `VaultTransactionMessage::deserialize` to read as u16 LE (matching the Squads v4 on-chain format)
- Caches parsed IDL with `LazyLock` to avoid re-parsing 86KB JSON on every call

## Test plan
- [x] `cargo clippy -p visualsign-solana --all-targets -- -D warnings`
- [x] `cargo test -p visualsign-solana` — 5 new squads tests (IDL loading, discriminators, error handling, vault transaction deserialization)
- [x] `make -C src test` — full workspace passes
- [ ] Parse real Squads multisig transaction with `parser_cli`

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/mod.rs
@@ -28,6 +28,15 @@ pub enum VisualizerKind {
     Payments(&'static str),
 }
 
+/// Maximum nesting depth for visualizing inner instructions.
+///
+/// Solana's runtime CPI cap is 4, so any visualization that hits this depth is
+/// either malformed or adversarial input. The cap prevents stack-overflow DoS
+/// through nested-instruction encodings (e.g., a `vaultTransactionCreate`
+/// containing another `vaultTransactionCreate`, or a swig instruction wrapping
+/// another swig instruction).
+pub const MAX_CALL_DEPTH: u8 = 4;
+
 /// Context for visualizing a Solana instruction.
 ///
 /// Holds all necessary information to visualize a specific command
@@ -43,10 +52,12 @@ pub struct VisualizerContext<'a> {
     instructions: &'a Vec<Instruction>,
     /// IDL registry for parsing unknown programs with Anchor IDLs
     idl_registry: &'a crate::idl::IdlRegistry,
+    /// Depth of nested inner-instruction visualization (0 for top-level).
+    call_depth: u8,
 }
 
 impl<'a> VisualizerContext<'a> {
-    /// Creates a new `VisualizerContext`.
+    /// Creates a new top-level `VisualizerContext` with `call_depth = 0`.
     pub fn new(
         sender: &'a SolanaAccount,
         instruction_index: usize,
@@ -58,7 +69,34 @@ impl<'a> VisualizerContext<'a> {
             instruction_index,
             instructions,
             idl_registry,
+            call_depth: 0,
         }
+    }
+
+    /// Creates a child context for visualizing a nested inner instruction.
+    ///
+    /// Returns `None` when incrementing would exceed [`MAX_CALL_DEPTH`]. Callers
+    /// should treat `None` as a signal to emit a "max depth exceeded" fallback
+    /// rather than recursing further.
+    pub fn for_nested_call<'b>(
+        &self,
+        sender: &'b SolanaAccount,
+        instruction_index: usize,
+        instructions: &'b Vec<Instruction>,
+    ) -> Option<VisualizerContext<'b>>
+    where
+        'a: 'b,
+    {
+        if self.call_depth >= MAX_CALL_DEPTH {
+            return None;
+        }
+        Some(VisualizerContext {
+            sender,
+            instruction_index,
+            instructions,
+            idl_registry: self.idl_registry,
+            call_depth: self.call_depth + 1,
+        })
     }
 
     /// Returns a reference to the IDL registry
@@ -84,6 +122,11 @@ impl<'a> VisualizerContext<'a> {
     /// Returns the current instruction being visualized.
     pub fn current_instruction(&self) -> Option<&Instruction> {
         self.instructions.get(self.instruction_index)
+    }
+
+    /// Returns the nesting depth of this context (0 at the top level).
+    pub fn call_depth(&self) -> u8 {
+        self.call_depth
     }
 }
 
@@ -180,4 +223,40 @@ pub fn visualize_with_any(
                 }),
         )
     })
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used, clippy::expect_used, clippy::panic)]
+mod tests {
+    use super::*;
+    use crate::idl::IdlRegistry;
+
+    #[test]
+    fn for_nested_call_caps_at_max_call_depth() {
+        let sender = SolanaAccount {
+            account_key: "11111111111111111111111111111111".to_string(),
+            signer: false,
+            writable: false,
+        };
+        let instructions: Vec<Instruction> = vec![];
+        let registry = IdlRegistry::new();
+        let root = VisualizerContext::new(&sender, 0, &instructions, &registry);
+        assert_eq!(root.call_depth(), 0);
+
+        let mut current = root;
+        for expected_depth in 1..=u32::from(MAX_CALL_DEPTH) {
+            let next = current
+                .for_nested_call(&sender, 0, &instructions)
+                .unwrap_or_else(|| {
+                    panic!("for_nested_call refused before reaching cap at depth {expected_depth}")
+                });
+            assert_eq!(u32::from(next.call_depth()), expected_depth);
+            current = next;
+        }
+
+        assert!(
+            current.for_nested_call(&sender, 0, &instructions).is_none(),
+            "for_nested_call must return None once MAX_CALL_DEPTH is reached"
+        );
+    }
 }
--- a/src/chain_parsers/visualsign-solana/src/core/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/mod.rs
@@ -30,11 +30,12 @@ pub enum VisualizerKind {
 
 /// Maximum nesting depth for visualizing inner instructions.
 ///
-/// Solana's runtime CPI cap is 4, so any visualization that hits this depth is
-/// either malformed or adversarial input. The cap prevents stack-overflow DoS
-/// through nested-instruction encodings (e.g., a `vaultTransactionCreate`
-/// containing another `vaultTransactionCreate`, or a swig instruction wrapping
-/// another swig instruction).
+/// Matches Solana's runtime CPI cap of 4: contexts may reach `call_depth ==
+/// MAX_CALL_DEPTH`, but any attempt to nest beyond it (i.e. exceed this depth)
+/// returns `None` from [`VisualizerContext::for_nested_call`]. The cap
+/// prevents stack-overflow DoS through nested-instruction encodings (e.g.,
+/// a `vaultTransactionCreate` containing another `vaultTransactionCreate`,
+/// or a swig instruction wrapping another swig instruction).
 pub const MAX_CALL_DEPTH: u8 = 4;
 
 /// Context for visualizing a Solana instruction.
--- a/src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
@@ -4,11 +4,16 @@ mod config;
 
 use crate::core::{
     InstructionVisualizer, SolanaIntegrationConfig, VisualizerContext, VisualizerKind,
+    available_visualizers, visualize_with_any,
 };
+use crate::idl::IdlRegistry;
 use config::SquadsMultisigConfig;
+use solana_parser::solana::structs::SolanaAccount;
 use solana_parser::{
     Idl, SolanaParsedInstructionData, decode_idl_data, parse_instruction_with_idl,
 };
+use solana_sdk::instruction::{AccountMeta, Instruction};
+use solana_sdk::pubkey::Pubkey;
 use std::collections::HashMap;
 use visualsign::errors::VisualSignError;
 use visualsign::field_builders::{create_raw_data_field, create_text_field};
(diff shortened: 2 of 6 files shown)


src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+/// `visualize_with_any` selects the first visualizer whose `can_handle` returns true and
+/// invokes it. If that visualizer's `visualize_tx_commands` then fails — or no visualizer
+/// matches — we explicitly fall back to `UnknownProgramVisualizer` (a catch-all). This
+/// guarantees every inner instruction produces *some* output: a failure in a specific
+/// program's visualizer never silently drops the field.
+fn visualize_inner_instructions(
+    inner_instructions: &[Instruction],
+    context: &VisualizerContext,
+) -> Result<Vec<AnnotatedPayloadField>, VisualSignError> {
+    let visualizers: Vec<Box<dyn InstructionVisualizer>> = available_visualizers();
```
pepe-anchor, line 539: **CRITICAL: unbounded recursion / DoS**

`available_visualizers()` includes `SquadsMultisigVisualizer` itself (auto-registered by `build.rs`). A `vaultTransactionCreate` whose embedded `transactionMessage` references program `SQDS4ep…` with another `vaultTransactionCreate` recurses without bound → stack overflow. `VisualizerContext` has no depth field.

Fix: add a `depth: u8` to `VisualizerContext`, increment on entry to `visualize_inner_instructions`, bail at a small cap (e.g. 4).

---
_Authored by Claude on behalf of @pepe-anchor._

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+            if program_id_idx >= account_keys.len() {
+                return None;
+            }
+
+            let accounts: Vec<AccountMeta> = ci
+                .account_indexes
+                .iter()
+                .filter_map(|&i| {
+                    let idx = i as usize;
+                    if idx < account_keys.len() {
```
pepe-anchor: **HIGH: ALT-delivered accounts silently dropped**

When `account_indexes` references an index ≥ `account_keys.len()` (account delivered via an address table lookup), the entry is silently dropped by `filter_map`. The inner instruction renders with a truncated account list — a transfer destination could be invisible to the user.

Fix: detect any out-of-range index and fall back to generic display with a warning banner.

---
_Authored by Claude on behalf of @pepe-anchor._

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+        parsed,
+        named_accounts,
+    })
+}
+
+fn build_named_accounts(
+    data: &[u8],
+    idl: &Idl,
+    accounts: &[AccountMeta],
+) -> HashMap<String, String> {
```
pepe-anchor: **HIGH: `HashMap` instead of `BTreeMap` — non-deterministic field order**

Iterating `HashMap<String, String>` to build `Vec<AnnotatedPayloadField>` produces non-deterministic field ordering across runs. `dflow_aggregator` (the reference IDL preset) uses `BTreeMap` here.

Fix: change return type and `SquadsParsedInstruction.named_accounts` to `BTreeMap<String, String>`.

---
_Authored by Claude on behalf of @pepe-anchor._

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+                return None;
+            }
+
+            let accounts: Vec<AccountMeta> = ci
+                .account_indexes
+                .iter()
+                .filter_map(|&i| {
+                    let idx = i as usize;
+                    if idx < account_keys.len() {
+                        Some(AccountMeta::new_readonly(account_keys[idx], false))
```
pepe-anchor: **MEDIUM: all inner accounts marked `new_readonly`**

The header fields `_num_writable_signers` / `_num_writable_non_signers` are parsed but discarded — every reconstructed account gets `is_writable=false, is_signer=false`. Downstream visualizers that branch on writability will produce incorrect output.

---
_Authored by Claude on behalf of @pepe-anchor._

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+fn visualize_inner_instructions(
+    inner_instructions: &[Instruction],
+    context: &VisualizerContext,
+) -> Result<Vec<AnnotatedPayloadField>, VisualSignError> {
+    let visualizers: Vec<Box<dyn InstructionVisualizer>> = available_visualizers();
+    let visualizers_refs: Vec<&dyn InstructionVisualizer> =
+        visualizers.iter().map(|v| v.as_ref()).collect();
+
+    let idl_registry = IdlRegistry::new();
+    let sender = SolanaAccount {
```
pepe-anchor: **MEDIUM: inner sender is the outer fee-payer, not the vault PDA**

At execution time the signer of inner instructions is the vault PDA (`multisig + vaultIndex`), not the fee-payer creating the proposal. Visualizers that compare `context.sender()` to identify ownership will misattribute the source.

---
_Authored by Claude on behalf of @pepe-anchor._

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+        return Ok(None);
+    };
+
+    // Reconstruct full Instructions from the compiled instructions
+    let inner_instructions = reconstruct_instructions(&vault_msg);
+
+    // Visualize inner instructions using the full visualizer framework
+    let inner_fields = visualize_inner_instructions(&inner_instructions, context)?;
+
+    let vault_index = args_value
```
pepe-anchor: **MEDIUM: `vault_index` silently defaults to 0 when missing**

`unwrap_or(0)` on a security-relevant field: vault 0 is valid, so a parse failure is indistinguishable from an explicit index 0.

---
_Authored by Claude on behalf of @pepe-anchor._

pepe-anchor (dismissed):

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
prasanna-anchorage: Fixed in 876ddf1. Added a `call_depth` field + `MAX_CALL_DEPTH=4` constant to `VisualizerContext` (matching Solana's runtime CPI cap). Inner visualization in both `squads_multisig` and `swig_wallet` now goes through `context.for_nested_call(...)`; once the cap is reached the visualizer emits an explicit "max depth exceeded" field instead of recursing. Generalized to swig_wallet because the same recursion shape exists there per the related security finding.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
prasanna-anchorage: Fixed in 876ddf1. `reconstruct_instructions` now returns `Result`; any out-of-range program-id or account index causes `try_build_vault_transaction_fields` to return `Ok(None)`, falling back to the generic display instead of silently dropping accounts. Test: `test_reconstruct_instructions_out_of_range_returns_err`.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
prasanna-anchorage: Fixed in 876ddf1. `build_named_accounts` and `SquadsParsedInstruction.named_accounts` are now `BTreeMap<String, String>`, matching `dflow_aggregator`. Test: `test_build_named_accounts_returns_btreemap_for_deterministic_order` calls the builder twice and asserts identical iteration order.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
prasanna-anchorage: Fixed in 876ddf1. Header counts (`num_signers`, `num_writable_signers`, `num_writable_non_signers`) are now retained on `VaultTransactionMessage`. `reconstruct_instructions` derives per-account `is_signer` / `is_writable` from those counts using the standard Solana `MessageHeader` ordering convention. Test: `test_reconstruct_instructions_writable_flags_match_header`.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
prasanna-anchorage: Fixed in 876ddf1. Inner-instruction visualization now derives the vault PDA (`Pubkey::find_program_address(&[b"multisig", multisig.as_ref(), b"vault", &[vault_index]], SQUADS_MULTISIG_PROGRAM_ID)`) and uses it as the nested context's `sender`. Falls back to the generic display if `multisig` is missing from named_accounts or the program-id constant fails to parse. Test: `test_vault_pda_account_derives_pda`.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
prasanna-anchorage: Fixed in 876ddf1. `vaultIndex` must now be present and fit a u8 (Squads only supports vault indices 0..=255); missing or out-of-range values cause `try_build_vault_transaction_fields` to return `Ok(None)`, falling back to the generic display so the user can still see the raw decoded args. Test: `test_try_build_vault_transaction_fields_missing_vault_index_returns_none`.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+        }
+
+        // Address table lookups: u8 count (we skip parsing the contents)
+        // They're not needed for instruction reconstruction without ALT resolution
+
+        Ok(Self {
+            account_keys,
+            instructions,
+        })
+    }
```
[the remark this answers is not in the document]
prasanna-anchorage: Resolved in dc52b29 (predates this thread). The current code parses the address-table-lookups section in full and then asserts `pos == data.len()`, so trailing or padding bytes are now rejected by `VaultTransactionMessage::deserialize`.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+        .iter()
+        .enumerate()
+        .filter_map(|(idx, _)| {
+            let inner_context =
+                VisualizerContext::new(&sender, idx, &instructions_vec, &idl_registry);
+
+            visualize_with_any(&visualizers_refs, &inner_context)
+                .and_then(|result| result.ok())
+                .map(|viz_result| viz_result.field)
+        })
```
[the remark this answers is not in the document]
prasanna-anchorage: Resolved in d28dd18. `visualize_inner_instructions` no longer drops errors via `result.ok()`; `Some(Err(_))` and `None` both fall back to `UnknownProgramVisualizer` so every inner instruction produces some output.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,602 @@
...
+        &format!("{inner_count} instruction(s)"),
+    ) {
+        condensed_fields.push(f);
+    }
+
+    // Expanded: full details + decoded inner instructions
+    let mut expanded_fields = vec![];
+    if let Ok(f) = create_text_field("Program ID", program_id) {
+        expanded_fields.push(f);
+    }
```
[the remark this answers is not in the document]
prasanna-anchorage: Resolved in d28dd18. All `create_text_field` calls in this module now use `?` propagation; the `if let Ok(f) = ...` pattern was removed.

shahan-khatchadourian-anchorage (commented):
A few items that don't appear to be covered by the existing review threads:

**1. Merge conflict on `src/chain_parsers/visualsign-solana/src/presets/mod.rs` (rebase required)**

GitHub reports this PR as `mergeStateStatus: DIRTY`. Commit `06ec9a5` (#278) rewrote `presets/mod.rs` to a single `include!(...)` that auto-generates `pub mod <name>;` declarations from the filesystem at build time. The file's own header says *"To add a preset: create `src/presets/<name>/...`. Do NOT edit this file — it has no hand-maintained content."*

On rebase, the PR's `+pub mod squads_multisig;` hunk should be **dropped entirely** — the `squads_multisig/` directory will be picked up by `build.rs::emit_module_declarations` automatically. If it's re-added, you'll get a duplicate `pub mod` at compile time.

**2. Missing fixture-based snapshot test**

The PR test plan still has `[ ] Parse real Squads multisig transaction with parser_cli` unchecked. Per `CLAUDE.md`: *"Fixture-based snapshot tests: `tests/fixtures/{name}.input` + `{name}.expected` pairs per chain crate"*. `presets/jupiter_swap/tests/` is the reference pattern. For the highest-risk path here — `vaultTransactionCreate` with nested visualization, vault-PDA derivation, header-decoded writable flags — a fixture covering one real Squads tx with at least one decodable inner instruction would catch regressions across the whole pipeline in a single test. The `tx_msg_hex` blob already used in `test_vault_transaction_message_deserialization` is a good seed.

**3. (Nit) \`_probe\` pattern in \`visualize_inner_instructions\`**

\`\`\`rust
let Some(_probe) = context.for_nested_call(&owned_sender, 0, &instructions_vec) else {
    return Ok(vec![create_text_field(\"Inner Instructions\", \"<truncated...>\")?]);
};
// loop body then calls for_nested_call again per iteration, with an
// \"Should be unreachable given the probe above\" safety net
\`\`\`

The probe allocates a \`VisualizerContext\` only to drop it for a depth check. Equivalent and simpler:

\`\`\`rust
if context.call_depth() >= MAX_CALL_DEPTH {
    return Ok(vec![create_text_field(\"Inner Instructions\", \"<truncated...>\")?]);
}
\`\`\`

This also eliminates the \"unreachable safety net\" branch in the loop.

**4. (Nit) Divergent fallback policy between swig and squads**

Both presets visualize inner instructions but handle \`Some(Err(_)) | None\` from \`visualize_with_any\` differently:
- Squads falls back to \`UnknownProgramVisualizer\` so every inner instruction emits a field.
- Swig returns \`InnerInstructionSummary::NotVisualized\` and the caller renders a byte-count summary string.

Not necessarily wrong (the two sites produce different output shapes — full \`AnnotatedPayloadField\` vs. summary \`String\`), but worth a one-liner in one of the sites explaining the divergence so a future reader doesn't read it as a bug.

---

Items 1 and 2 are the ones I'd want addressed before merge; 3 and 4 are cleanup nits.

src/chain_parsers/visualsign-solana/src/presets/squads_multisig/mod.rs
```diff
@@ -0,0 +1,918 @@
...
+///   - everything else is non-signer + readonly
+fn reconstruct_instructions(
+    vault_msg: &VaultTransactionMessage,
+) -> Result<Vec<Instruction>, &'static str> {
+    let account_keys = &vault_msg.account_keys;
+    let num_keys = account_keys.len();
+    let num_signers = vault_msg.num_signers as usize;
+    let num_writable_signers = vault_msg.num_writable_signers as usize;
+    let num_writable_non_signers = vault_msg.num_writable_non_signers as usize;
+
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in f4df833. `reconstruct_instructions` now validates the three header invariants up-front before any flag derivation:
* `num_writable_signers <= num_signers`
* `num_signers <= account_keys.len()`
* `num_writable_non_signers <= account_keys.len() - num_signers`

Any violation returns `Err` and `try_build_vault_transaction_fields` falls back to the generic display, so a malformed proposal can't mislabel non-signers as signers/writable in the inner-instruction view. Covered by `test_reconstruct_instructions_rejects_inconsistent_header` (all three failure modes).

src/chain_parsers/visualsign-solana/src/core/mod.rs
```diff
@@ -28,6 +28,15 @@ pub enum VisualizerKind {
     Payments(&'static str),
 }
 
+/// Maximum nesting depth for visualizing inner instructions.
+///
+/// Solana's runtime CPI cap is 4, so any visualization that hits this depth is
+/// either malformed or adversarial input. The cap prevents stack-overflow DoS
+/// through nested-instruction encodings (e.g., a `vaultTransactionCreate`
+/// containing another `vaultTransactionCreate`, or a swig instruction wrapping
+/// another swig instruction).
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed (wording) in f4df833. Kept the behavior — contexts may reach `call_depth == MAX_CALL_DEPTH`, matching Solana's runtime CPI cap of 4 — and rewrote the doc comment to say any attempt to *exceed* this depth returns `None`. The cap and the existing `for_nested_call_caps_at_max_call_depth` test are unchanged.

shahan-khatchadourian-anchorage (approved):
no blocking comments. lgtm.

PR #266, opened 2026-04-22
feat: add Meteora DLMM Solana preset

kyle-anchorage:
## Why this PR exists

Solana transactions that call the Meteora DLMM (`lb_clmm`) program currently fall through to the generic unknown-program visualizer, so signers see only a program ID and raw instruction hex. Meteora DLMM is a top-tier Solana DEX and users signing swaps or liquidity actions need to see what they're approving.

## What changed

- New preset `visualsign-solana/src/presets/meteora_dlmm/` containing `config.rs` (program-ID match), `mod.rs` (IDL-driven visualizer), and a bundled `meteora_dlmm.json` IDL.
- Registered `pub mod meteora_dlmm;` in `presets/mod.rs`; `build.rs` auto-discovers the `MeteoraDlmmVisualizer` struct, so no other registration is needed.
- The visualizer bundles the IDL via `include_str!`, decodes it once behind a `OnceLock`, and uses `solana_parser::parse_instruction_with_idl` to produce instruction name, discriminator, args, and IDL-named accounts. Output is rendered through the shared `create_text_field` / `create_raw_data_field` builders into the standard condensed/expanded preview layout. Returns `VisualizerKind::Dex("Meteora DLMM")`.
- Seven unit tests: IDL loads, all discriminators are 8 bytes, short data errors, unknown discriminator errors, a real `swap` instruction decodes with expected args, `kind()` returns `Dex`, and named-account pairing preserves IDL account order.

image

## Why this is safe

- No changes to existing visualizers or shared code — only additive. Unknown-program fallback still exists for anything this preset can't parse, and the `can_handle` gate is scoped to the Meteora DLMM program ID, so other programs are untouched.
- The IDL is bundled, not fetched — no network I/O, no runtime dependency on external state.
- Parsing uses the existing `solana_parser` crate the same way `unknown_program` and `jupiter_swap` do; errors from discriminator mismatches or truncated data return `VisualSignError` rather than panicking, and the workspace clippy denials (`unwrap_used`, `expect_used`, `panic`) are enforced in non-test code.

### Checks run (by agent)
- `cargo clippy -p visualsign-solana --all-targets -- -D warnings`
- `cargo test -p visualsign-solana` (all 69 tests pass, including the 7 new ones)
- `make -C src test` (full workspace: all tests pass)
- `make -C src lint` (workspace clippy clean)

None — fully automated by CI.

## How this is maintainable

- Each preset is a self-contained directory under `src/presets/` and is auto-discovered by `build.rs` from the `{PascalName}Visualizer` convention — adding or removing a preset is a one-line change in `presets/mod.rs` plus the directory.
- The visualizer is generic IDL-driven: no hand-written per-instruction decoders. Updating to a newer Meteora IDL is just replacing `meteora_dlmm.json`; the `test_meteora_dlmm_idl_has_discriminators` test catches malformed IDL and `test_swap_instruction_parses` catches a real-world regression on the most-used instruction.
- Pattern matches `jupiter_swap` and `unknown_program` presets already in the tree, so any contributor who has touched those can extend this one. CLAUDE.md documents the preset conventions (field builders, ASCII-only, imports at top of module) that this PR follows.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/meteora_dlmm/config.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/meteora_dlmm/config.rs
@@ -1,0 +1,22 @@
+use super::METEORA_DLMM_PROGRAM_ID;
+use crate::core::{SolanaIntegrationConfig, SolanaIntegrationConfigData};
+use std::collections::HashMap;
+
+pub struct MeteoraDlmmConfig;
+
+impl SolanaIntegrationConfig for MeteoraDlmmConfig {
+    fn new() -> Self {
+        Self
+    }
+
+    fn data(&self) -> &SolanaIntegrationConfigData {
+        static DATA: std::sync::OnceLock<SolanaIntegrationConfigData> = std::sync::OnceLock::new();
+        DATA.get_or_init(|| {
+            let mut programs = HashMap::new();
+            let mut instructions = HashMap::new();
+            instructions.insert("*", vec!["*"]);
+            programs.insert(METEORA_DLMM_PROGRAM_ID, instructions);
+            SolanaIntegrationConfigData { programs }
+        })
+    }
+}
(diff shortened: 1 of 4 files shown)


prasanna-anchorage (approved):
LGTM — Meteora DLMM preset. Validated via parser_cli decode of /tmp/fixtures-review/266-meteora_dlmm.json. Rebased onto current main; only conflict was the trivial alphabetical add to presets/mod.rs.

PR #267, opened 2026-04-22
feat: add Orca Whirlpool Solana preset

kyle-anchorage:
## Why this PR exists

Orca Whirlpool transactions show up as opaque `Instruction N` fields with raw program data, so users signing Orca swaps or liquidity actions cannot verify what they are approving. There is no preset for the Orca Whirlpool program (`whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc`).

## What changed

- New preset `visualsign-solana::presets::orca_whirlpool` registered alongside `jupiter_swap`, `stakepool`, etc.
- Vendored the official Orca Whirlpool IDL from `@orca-so/whirlpools-sdk` as `orca_whirlpool.json` (66 instructions, Anchor 0.30+ format — no local/Docker anchor-cli was available and the on-chain IDL account is empty for this program).
- Generic IDL-driven visualizer: `parse_instruction_with_idl` resolves instruction name, named accounts, and args; the expanded field list renders Program ID, Instruction Name, sorted named accounts, sorted args, and the raw data blob. No program-specific decoding — every Whirlpool instruction flows through the same path.
- `OrcaWhirlpoolConfig` registers the program ID with a wildcard instruction matcher so every Whirlpool instruction is claimed by this visualizer.
- `VisualizerKind::Dex("Orca Whirlpool")`.

image

## Why this is safe

- Strictly additive. No existing preset, type, or registry entry is modified; `presets/mod.rs` gains one `pub mod` line inserted alphabetically, and `build.rs` auto-discovers `OrcaWhirlpoolVisualizer` by directory name.
- Visualization only — no settlement, signing, or state mutation. Worst case for a malformed instruction is that `parse_instruction_with_idl` returns an error and the outer pipeline falls back to the unknown-program visualizer.
- Guarded inputs: data shorter than 8 bytes returns `DecodeError` before touching the IDL; unknown discriminators surface as errors from `solana_parser` rather than panicking.
- No `unwrap`/`expect`/`panic` in non-test code, in line with the workspace lint policy.

### Checks run (by agent)
- `cargo clippy -p visualsign-solana --all-targets -- -D warnings`
- `cargo test -p visualsign-solana` — all crate tests pass, including four new unit tests (`test_orca_whirlpool_idl_loads`, `test_orca_whirlpool_idl_has_discriminators`, `test_unknown_discriminator_returns_error`, `test_short_data_returns_error`)
- `make test` — full workspace test suite green

None — fully automated by CI. **Risk: low.**

## How this is maintainable

- The visualizer is a thin wrapper around `solana_parser::parse_instruction_with_idl`, so a fresh reader only needs to understand the IDL parse output (`SolanaParsedInstructionData`) to follow the code.
- IDL is a static `include_str!` baked into the binary; refreshing it means replacing `orca_whirlpool.json` and rerunning tests — `test_orca_whirlpool_idl_has_discriminators` will catch any structural regression on reload.
- The `OnceLock`-cached IDL decode means the first instruction pays the parse cost and the rest are free; if Anchor ships a new spec version, the failure surfaces as a `VisualSignError::DecodeError` on first use rather than a panic.
- Follows the same shape as `jupiter_swap` (the other Dex preset), so anyone adding a new generic Anchor program preset has two adjacent examples to copy from.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/mod.rs
@@ -11,6 +11,7 @@ pub mod meteora_damm_v2;
 pub mod meteora_dlmm;
 pub mod neutral_trade;
 pub mod onre_app;
+pub mod orca_whirlpool;
 pub mod stakepool;
 pub mod swig_wallet;
 pub mod system;
--- a/src/chain_parsers/visualsign-solana/src/presets/orca_whirlpool/config.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/orca_whirlpool/config.rs
@@ -1,0 +1,22 @@
+use super::ORCA_WHIRLPOOL_PROGRAM_ID;
+use crate::core::{SolanaIntegrationConfig, SolanaIntegrationConfigData};
+use std::collections::HashMap;
+
+pub struct OrcaWhirlpoolConfig;
+
+impl SolanaIntegrationConfig for OrcaWhirlpoolConfig {
+    fn new() -> Self {
+        Self
+    }
+
+    fn data(&self) -> &SolanaIntegrationConfigData {
+        static DATA: std::sync::OnceLock<SolanaIntegrationConfigData> = std::sync::OnceLock::new();
+        DATA.get_or_init(|| {
+            let mut programs = HashMap::new();
+            let mut instructions = HashMap::new();
+            instructions.insert("*", vec!["*"]);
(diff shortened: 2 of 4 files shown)


prasanna-anchorage (approved):
LGTM — Orca Whirlpool preset. Validated via parser_cli decode of /tmp/fixtures-review/267-orca_whirlpool.json. Rebased onto current main; only conflict was the trivial alphabetical add to presets/mod.rs.

PR #268, opened 2026-04-28
feat: handle Jupiter v6 route_v2 instructions and refresh IDL

kyle-anchorage:
## Why this PR exists

Jupiter's `route_v2` family of instructions (`route_v2`, `exact_out_route_v2`, `shared_accounts_route_v2`, `shared_accounts_exact_out_route_v2`) is live on mainnet but absent from the IDL bundled inside `solana_parser`. Transactions using these instructions fail to decode in two places: the `jupiter_swap` visualizer falls through to `Unknown`, and `decode_v0_transfers` aborts the whole-transaction decode with "no matching instruction discriminator". Wallets signing a v2 route see a generic Jupiter blob instead of input/output tokens, slippage, and fees.

## What changed

- Vendored the current Jupiter v6 IDL as `jupiter_swap/jupiter_agg_v6.json`, decoded via `include_str!` instead of `solana_parser::ProgramType::JupiterAggregatorV6`. New instructions can land here without bumping `solana_parser`.
- Added four `JupiterSwapInstruction` variants (`RouteV2`, `ExactOutRouteV2`, `SharedAccountsRouteV2`, `SharedAccountsExactOutRouteV2`) and a single `parse_route_v2` helper. `exact_out` toggles `in_amount`/`quoted_out_amount` vs `out_amount`/`quoted_in_amount`; `shared` shifts the source/destination mint indices from (3, 4) to (6, 7) per the IDL.
- `format_jupiter_swap_instruction` and `create_jupiter_swap_expanded_fields` collapse the four v2 variants into a single or-pattern arm that emits Input/Output Token, Slippage, optional Platform Fee, and optional Positive Slippage fields.
- `decode_v0_transfers` injects the local IDL into `solana_parser::parse_transaction` via the `custom_idls` map, so the whole-transaction decode picks up the same instruction set the visualizer uses.

## Why this is safe

- Display-only path. No changes to signing, custody, or settlement — this only affects how a Jupiter v6 transaction is rendered in the VisualSign payload before the user signs.
- Backward compatible. Existing variants (`Route`, `ExactOutRoute`, `SharedAccountsRoute`, `SharedAccountsExactOutRoute`) are untouched, and the local IDL is a superset of the bundled upstream copy — any instruction that decoded before still decodes.
- Variant-specific byte layout is pinned by the four new unit tests, including the leading `id_byte` for shared variants and the swapped in/out semantics for `exact_out`.
- Workspace clippy denials honored: no `unwrap_used`, `expect_used`, `panic`, or `unsafe_code` in non-test code.
- IDL-drift risk acknowledged: if Jupiter ships a `route_v3`, this preset falls back to `Unknown` rather than misdecode — the same behavior we had with v2 before this PR.

### Checks run (by agent)

- `make -C src build` — green
- `make -C src lint` (clippy `-D warnings`) — clean
- `make -C src test` — full workspace suite green
- `cargo test -p visualsign-solana jupiter` — all 13 jupiter tests pass, including the 4 new v2 tests

None — fully automated by CI. **Risk: low.**

## How this is maintainable

- Vendoring the IDL means future Jupiter additions are a JSON refresh + handler addition in this repo, not a `solana_parser` rev bump.
- `parse_route_v2` takes two booleans (`exact_out`, `shared`) so all four v2 variants share one decode path; reading any one of the new tests reveals the shape of the rest.
- The four new tests pin exact byte layout per variant; an IDL refresh that breaks decoding trips them immediately.
- The doc comment on `JUPITER_IDL_JSON` documents the dual purpose (visualizer + `decode_v0_transfers` override) so a future contributor does not silently drop the `pub(crate)` constant when refactoring the visualizer.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/txtypes/v0.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/txtypes/v0.rs
@@ -15,7 +15,9 @@ use visualsign::{
 pub fn decode_v0_transfers(
     versioned_tx: &VersionedTransaction,
 ) -> Result<Vec<AnnotatedPayloadField>, VisualSignError> {
+    use crate::presets::jupiter_swap::{JUPITER_IDL_JSON, JUPITER_PROGRAM_ID};
     use solana_parser::solana::parser::parse_transaction;
+    use std::collections::HashMap;
 
     // Serialize the full versioned transaction
     let transaction_bytes = bincode::serialize(versioned_tx).map_err(|e| {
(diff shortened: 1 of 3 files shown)


prasanna-anchorage (approved):
LGTM — preset extension that handles route_v2 family for Jupiter v6. Validated end-to-end via parser_cli decode of a real route_v2 tx (output saved to /tmp/fixtures-review/268-jupiter_swap.json during the QA review run). No conflicts on rebase against current main.

PR #273, opened 2026-04-30
feat(cli): add `serve` subcommand for browsing decoded txs

prasanna-anchorage:
## Summary

Adds a `serve` subcommand to `parser_cli` that decodes a directory of raw-transaction files and serves a small read-only web UI on `127.0.0.1` so reviewers can scroll through results in a browser instead of running the CLI per-file.

The directory is **re-decoded on every request** — edit a fixture, refresh the browser, see the new state. No restart, no file watcher, no JS framework. `.json` files in the directory are passed through verbatim, so a tree of `.expected` outputs can be browsed alongside hex inputs (or on its own).

Each entry has its own browseable / `curl`-friendly URL, and the page is mobile-friendly with a per-entry **copy** button.

## Screenshots

| Desktop | Mobile |
| --- | --- |
| _add screenshot_ | _add screenshot_ |

Suggested shots: index page on desktop, same page narrow / on a phone, and the copy button mid-press (the `copied!` confirmation state).

## Try it against existing fixtures

The chain-parser fixture trees have `.input` (raw hex) and `.expected` (recorded JSON) sitting next to each other — perfect demo for this subcommand. The hex files get decoded fresh; the JSON files pass through as-is.

Run from `src/` (same directory as the workspace `Cargo.toml`):

```bash
# Ethereum chain-parser fixtures
cargo run -p parser_cli --features serve -- serve \
  --chain ethereum -n ETHEREUM_MAINNET \
  --dir chain_parsers/visualsign-ethereum/tests/fixtures

# Solana chain-parser fixtures (recurses into jupiter_swap/, token_2022/)
cargo run -p parser_cli --features serve -- serve \
  --chain solana -n SOLANA_MAINNET \
  --dir chain_parsers/visualsign-solana/tests/fixtures

# Smaller CLI smoke set
cargo run -p parser_cli --features serve -- serve \
  --chain ethereum -n ETHEREUM_MAINNET \
  --dir parser/cli/tests/fixtures
```

Then http://127.0.0.1:47474. Each fixture name appears twice (`1559.input` decoded fresh, `1559.expected` served as-is). Edit any file, refresh, see the new state.

`--dir` is resolved relative to the shell's CWD, so pass an absolute path if you want to invoke it from anywhere. Default port is **47474** rather than 8080 to avoid colliding with the dev-container REST port and the long list of other tools that grab 8080. Override with `--port`.

## Routes

- `GET /` — HTML index. One collapsible `<details>` per file, JSON inline. Each summary has a `[json]` link to the file's standalone URL plus a `copy` button that puts the rendered JSON onto the clipboard.
- `GET /<rel-path>` — Bare payload as JSON, no envelope. e.g. `/token_2022/transfer_checked.json` returns that file's decoded payload (or its raw JSON, in passthrough mode). Browseable, bookmarkable, `curl`-friendly.
- `GET /api/file?path=<rel-path>` — Wrapped envelope: `{path, ok, payload?, error?}`. Kept around for tooling that wants the success/error flag without checking the HTTP status code.

All routes re-walk the directory and re-decode each file on every hit. With ~tens of fixtures this is single-digit milliseconds; not worth a cache.

## File handling

- `.json` extension → file is parsed as `serde_json::Value` and served as-is. No schema validation, no coercion to `SignablePayload`.
- everything else → trim, dispatch through the chain registry to a `SignablePayload`, serialize to JSON.

A directory can be all hex, all JSON, or mixed.

## Mobile + copy button

The index page works on a phone and ships a vanilla-JS clipboard helper:

- `<meta name=viewport content="width=device-width, initial-scale=1">` — the actual reason it wasn't mobile-friendly before.
- Long rel-paths wrap with `word-break: break-all`; `[json]` link and `copy` button get tappable padding; narrow-viewport CSS via `@media (max-width: 600px)`.
- Each entry renders a `<button class=copy>` next to its `[json]` link. Click → `navigator.clipboard.writeText(pre.textContent)` → button flips to `copied!` for 1.2s. `event.stopPropagation()` so the click doesn't toggle the `<details>` open/closed.
- No framework. ~10 lines of plain JS embedded in the page.

## CLI shape (breaking)

The pre-existing flat invocation moves under a `decode` subcommand:

```bash
# before
parser_cli --chain ethereum -t @tx.hex

# after
parser_cli decode --chain ethereum -t @tx.hex
parser_cli serve  --chain ethereum --dir ./txs        # new
```

Pre-1.0 internal CLI; happy to revisit if the breakage is undesirable.

## Dependencies

`axum 0.8` + `tokio` are added as **optional** direct deps of `parser_cli`, gated behind a new `serve` Cargo feature. Both crates are already transitive via `tonic` in the lock file, so the resolved dep tree gains zero new compiled crates. Default builds (and `parser_app`/enclave) are unchanged — axum/tokio are not pulled in unless `--features serve` is passed.

```toml
[features]
serve = ["dep:axum", "dep:tokio"]

[dependencies]
axum  = { version = "0.8", optional = true, default-features = false, features = ["http1", "tokio", "query", "json"] }
tokio = { workspace = true, optional = true, features = ["rt", "macros", "net"] }
```

No new deps for live reload or the copy button — both are behavior changes in the existing handlers / a vanilla-JS snippet inlined into the rendered HTML.

## Out of scope (deferred to follow-ups if useful)

- File-watcher / SSE / WebSocket auto-refresh — manual browser refresh is the trigger by design
- Multiple `--dir` flags / multiple chains in one server instance
- Paste-and-decode form (POST endpoint)
- Watching `--abi-json-mappings` / `--idl-mappings` files (registry stays static)
- Auth / TLS — `127.0.0.1`-only by design

## Test plan

- [x] `cargo build -p parser_cli` (default) — unchanged, no axum/tokio
- [x] `cargo build -p parser_cli --features serve`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo clippy -p parser_cli --features serve --all-targets -- -D warnings`
- [x] `cargo test -p parser_cli --features serve serve` — 10 unit tests + 2 integration tests pass, including:
  - `json_files_passthrough_as_is` — `.json` file served verbatim
  - `malformed_json_errors_with_invalid_json_prefix` — bad JSON surfaces as an error entry
  - `mixed_hex_and_json_directory_decodes_both` — both decode paths in one dir
  - `url_encode_path_preserves_separators_and_safe_chars` — rel-path → URL helper
  - `render_html_contains_paths_and_payload` — viewport meta tag present, copy button rendered per entry, clipboard JS hooked up
  - `render_error_page_is_mobile_friendly` — error page also carries the viewport meta
  - `test_cli_serve_live_reload_and_json_passthrough` — end-to-end: spawn server, mutate JSON file → second GET reflects change, drop in a new file → it shows up at `/`, truncate hex file → flips to error state, standalone-URL route returns bare payload + 404 on miss, all without restart
- [x] Manual smoke against `chain_parsers/visualsign-ethereum/tests/fixtures` — both `.input` and `.expected` files render, edits live-reload on browser refresh, standalone URLs (`/<rel-path>`) return bare payloads
- [ ] Manual mobile check — load `http://<dev-host>:47474` from a phone (or Chrome DevTools device emulation), confirm no horizontal scroll on the page chrome, copy button copies the entry's JSON

### DIFF
--- a/src/parser/cli/src/cli.rs
+++ b/src/parser/cli/src/cli.rs
@@ -1,5 +1,5 @@
 use crate::chains::parse_chain;
-use clap::Parser;
+use clap::{Parser, Subcommand};
 use visualsign::registry::{Chain, TransactionConverterRegistry};
 use visualsign::vsptrait::{DeveloperConfig, VisualSignOptions};
 use visualsign::{SignablePayload, SignablePayloadField};
@@ -9,6 +9,21 @@ use visualsign::{SignablePayload, SignablePayloadField};
 #[command(version = env!("VERSION"))]
 #[command(about = "Converts raw transactions to visual signing properties")]
 pub(crate) struct Args {
+    #[command(subcommand)]
+    pub(crate) command: Command,
+}
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum Command {
+    /// Decode a single transaction and print it.
+    Decode(DecodeArgs),
+    /// Serve a directory of raw-transaction files via a local web UI.
+    #[cfg(feature = "serve")]
+    Serve(crate::serve::ServeArgs),
+}
+
+#[derive(clap::Args, Debug)]
+pub(crate) struct DecodeArgs {
     #[arg(short, long, help = "Chain type")]
     pub(crate) chain: String,
 
@@ -19,16 +34,16 @@ pub(crate) struct Args {
         help = "Raw transaction string. Prefix with '@' to read from a file \
                 (e.g. '@/path/to/tx.hex'), or use '@-' to read from stdin."
     )]
-    transaction: String,
+    pub(crate) transaction: String,
 
     #[arg(short, long, default_value = "text", help = "Output format")]
-    output: OutputFormat,
+    pub(crate) output: OutputFormat,
 
     #[arg(
         long,
         help = "Show only condensed view (what hardware wallets display)"
     )]
-    condensed_only: bool,
+    pub(crate) condensed_only: bool,
 
     #[arg(
         long,
@@ -49,8 +64,19 @@ pub(crate) struct Args {
     pub(crate) solana: crate::solana::SolanaArgs,
 }
 
+impl DecodeArgs {
+    pub(crate) fn plugin_args(&self) -> crate::PluginArgs {
+        crate::PluginArgs {
+            #[cfg(feature = "ethereum")]
+            ethereum: self.ethereum.clone(),
+            #[cfg(feature = "solana")]
+            solana: self.solana.clone(),
+        }
+    }
+}
+
 #[derive(Debug, Clone, Copy)]
-enum OutputFormat {
+pub(crate) enum OutputFormat {
     Text,
     Json,
     Human,
@@ -257,70 +283,107 @@ fn parse_and_display(
     Ok(())
 }
 
+/// Resolves chain + plugins + per-chain metadata, returning a ready-to-use registry.
+///
+/// Shared by `decode` and `serve`. On invalid chain or metadata error, returns
+/// an `Err(String)` with a user-facing message — the caller is responsible for
+/// printing it and exiting non-zero.
+pub(crate) struct Runtime {
+    pub registry: TransactionConverterRegistry,
+    pub options: VisualSignOptions,
+}
+
+pub(crate) fn prepare_runtime(
+    chain_str: &str,
+    network: Option<String>,
+    plugin_args: &crate::PluginArgs,
+) -> Result<Runtime, String> {
+    let chain = parse_chain(chain_str);
+    let plugins = crate::build_plugins(plugin_args);
+
+    let mut registry = TransactionConverterRegistry::new();
+    for plugin in &plugins {
+        plugin.register(&mut registry);
+    }
+
+    let plugin = plugins.iter().find(|p| p.chain() == chain);
+
+    let Some(plugin) = plugin else {
+        let supported: Vec<String> = plugins
+            .iter()
+            .map(|p| p.chain().as_str().to_lowercase())
+            .collect();
+        let supported_str = if supported.is_empty() {
+            "none".to_string()
+        } else {
+            supported.join(", ")
+        };
+        if chain == Chain::Unspecified {
+            return Err(format!(
+                "unrecognized chain '{chain_str}'.\nSupported chains: {supported_str}"
+            ));
+        }
+        return Err(format!(
+            "chain '{chain_str}' is not supported by this CLI build.\nSupported chains: {supported_str}"
+        ));
+    };
+
+    let chain_metadata = plugin.create_metadata(network)?;
+
+    let options = VisualSignOptions {
+        decode_transfers: true,
+        transaction_name: None,
+        metadata: chain_metadata,
+        developer_config: Some(DeveloperConfig {
+            allow_signed_transactions: true,
+        }),
+    };
+
+    Ok(Runtime { registry, options })
+}
+
+fn execute_decode(args: &DecodeArgs) {
+    let plugin_args = args.plugin_args();
+    let runtime = match prepare_runtime(&args.chain, args.network.clone(), &plugin_args) {
+        Ok(r) => r,
+        Err(e) => {
+            eprintln!("Error: {e}");
+            std::process::exit(1);
+        }
+    };
+
+    let raw_tx = match crate::tx_input::resolve_transaction_input(&args.transaction) {
+        Ok(tx) => tx,
+        Err(e) => {
+            eprintln!("Error: {e}");
+            std::process::exit(1);
+        }
+    };
+
+    if let Err(e) = parse_and_display(
+        &args.chain,
+        &raw_tx,
+        &runtime.registry,
+        runtime.options,
+        args.output,
+        args.condensed_only,
+    ) {
+        eprintln!("Error: {e}");
+        std::process::exit(1);
+    }
+}
+
 /// CLI entry point.
 pub struct Cli;
 impl Cli {
     /// Parse arguments and run the transaction visualizer.
     pub fn execute() -> Result<(), String> {
         let args = Args::parse();
-        let chain = parse_chain(&args.chain);
-        let plugins = crate::build_plugins(&args);
-
-        let mut registry = TransactionConverterRegistry::new();
-        for plugin in &plugins {
-            plugin.register(&mut registry);
+        match &args.command {
+            Command::Decode(a) => execute_decode(a),
+            #[cfg(feature = "serve")]
+            Command::Serve(a) => crate::serve::execute_serve(a),
         }
-
-        let plugin = plugins.iter().find(|p| p.chain() == chain).ok_or_else(|| {
-            let supported: Vec<String> = plugins
-                .iter()
-                .map(|p| p.chain().as_str().to_lowercase())
-                .collect();
-            let supported_str = if supported.is_empty() {
-                "none".to_string()
-            } else {
-                supported.join(", ")
-            };
-            if chain == Chain::Unspecified {
-                format!(
-                    "unrecognized chain '{}'.\nSupported chains: {supported_str}",
-                    args.chain,
-                )
-            } else {
-                format!(
-                    "chain '{}' is not supported by this CLI build.\n\
-                     Supported chains: {supported_str}",
-                    args.chain,
-                )
-            }
-        })?;
-
-        let chain_metadata = plugin.create_metadata(args.network.clone())?;
-
-        let options = VisualSignOptions {
-            decode_transfers: true,
-            transaction_name: None,
-            metadata: chain_metadata,
-            developer_config: Some(DeveloperConfig {
-                allow_signed_transactions: true,
-            }),
-        };
-
-        let raw_tx = match crate::tx_input::resolve_transaction_input(&args.transaction) {
-            Ok(tx) => tx,
-            Err(e) => {
-                eprintln!("Error: {e}");
-                std::process::exit(1);
-            }
-        };
-
-        parse_and_display(
-            &args.chain,
-            &raw_tx,
-            &registry,
-            options,
-            args.output,
-            args.condensed_only,
-        )
+        Ok(())
     }
 }
--- a/src/parser/cli/src/serve.rs
+++ b/src/parser/cli/src/serve.rs
@@ -15,7 +15,7 @@ use std::sync::Arc;
 
 use axum::{
     Json, Router,
-    extract::{Query, State},
+    extract::{Path as AxumPath, Query, State},
     http::StatusCode,
     response::Html,
     routing::get,
@@ -142,6 +142,7 @@ pub fn execute_serve(args: &ServeArgs) {
     let app = Router::new()
         .route("/", get(handle_index))
         .route("/api/file", get(handle_file))
+        .route("/{*path}", get(handle_payload))
         .with_state(state);
 
     let rt = tokio::runtime::Builder::new_current_thread()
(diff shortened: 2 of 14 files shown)


src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    for entry in read {
+        let entry = entry.map_err(|e| format!("read_dir entry: {e}"))?;
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+        if name.starts_with('.') {
+            continue;
+        }
+        let path = entry.path();
+        let metadata = match entry.metadata() {
```
pepe-anchor, line 179: **`walk()` has no recursion depth limit — symlink loops crash the server**

`DirEntry::metadata()` follows symlinks (`stat`, not `lstat`), so a symlink loop (e.g. `dir/loop -> ..`) causes unbounded recursion and a stack overflow on the first HTTP request. No visited-set or depth cap.

Fix: skip symlinks via `entry.file_type().is_symlink()`.

---
_Found by Claude on behalf of @pepe-anchor._

src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+            out.push(DecodedEntry {
+                rel_path: rel,
+                result,
+            });
+        }
+    }
+    Ok(())
+}
+
+fn rel_path(base: &Path, full: &Path) -> String {
```
pepe-anchor, line 210: **`rel_path` falls back to the absolute path when `strip_prefix` fails**

If `strip_prefix(base)` fails (e.g. after symlink resolution diverges), `unwrap_or(full)` returns the full absolute path, which is then embedded in the rendered HTML. This leaks the host filesystem layout to anyone who can reach `127.0.0.1`.

Better: skip the entry with a synthetic error.

---
_Found by Claude on behalf of @pepe-anchor._

src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+    };
+    Ok(Html(render_html(&entries)))
+}
+
+/// Serve the decoded payload for a single file by its rel-path. Lets each
+/// entry have its own bookmarkable / shareable URL — e.g.
+/// `/token_2022/transfer_checked.json` returns just that file's payload as
+/// JSON. Wraps no envelope around it so browsers and `curl` see the raw
+/// `SignablePayload` (or the verbatim file content for `.json` passthrough).
+async fn handle_payload(
```
pepe-anchor, line 263: **Single-file endpoints decode the entire directory on every request**

`handle_payload` and `handle_file` both call `load_entries`, which walks and decodes all N files in `--dir` just to find one by equality. For single-file routes, validate the path is under `state.dir` and call `decode_file` directly -- same freshness at O(1) instead of O(N).

---
_Found by Claude on behalf of @pepe-anchor._

src/parser/cli/src/cli.rs
```diff
@@ -257,70 +283,107 @@ fn parse_and_display(
...
-            &registry,
-            options,
-            args.output,
-            args.condensed_only,
-        )
+/// app cli
+pub struct Cli;
+impl Cli {
+    /// start the parser cli
+    pub fn execute() -> Result<(), String> {
```
pepe-anchor: **`Cli::execute` signature promises `Result` but always returns `Ok(())`**

After the refactor, both `execute_decode` and `execute_serve` call `process::exit(1)` directly on errors, so `Cli::execute` never returns `Err`. The `if let Err(e) = Cli::execute()` branch in `main.rs` is now dead code.

Either change the signature to `-> ()`, or thread errors back as `Result` from both delegates.

---
_Found by Claude on behalf of @pepe-anchor._

src/parser/cli/src/cli.rs
```diff
@@ -9,6 +9,21 @@ use visualsign::{SignablePayload, SignablePayloadField};
 #[command(version = "1.0")]
 #[command(about = "Converts raw transactions to visual signing properties")]
 pub(crate) struct Args {
```
pepe-anchor: **Confirm no external callers of the old flat `--chain` / `-t` syntax**

The old `parser_cli --chain ethereum -t <hex>` interface is hard-broken. In-tree tests and fixtures were updated correctly.

Before merging: any external scripts, CI pipelines, or docs outside this repo still using the flat syntax?

---
_Found by Claude on behalf of @pepe-anchor._

src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+            Err(e) => {
+                out.push(DecodedEntry {
+                    rel_path: rel_path(base, &path),
+                    result: Err(format!("metadata: {e}")),
+                });
+                continue;
+            }
+        };
+        if metadata.is_dir() {
+            walk(base, &path, out, chain_str, runtime)?;
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed. Switched to `entry.file_type()` (uses `lstat`) and explicitly skip `is_symlink()` entries before recursing or decoding. The new test (`fn walk_skips_symlinks` would be a follow-up — the structural fix is the file_type/is_symlink path itself).

src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+        Err(err) => FileResponse {
+            path: &entry.rel_path,
+            ok: false,
+            payload: None,
+            error: Some(err),
+        },
+    };
+    serde_json::to_value(&response)
+        .map(Json)
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed. `handle_file` now always returns the `{ path, ok, payload?, error? }` envelope. Missing paths and `load_entries` failures populate `ok: false` with the cause in `error`. The HTTP status code still reflects outcome (200/404/422/500) so status-aware clients keep working, but the body contract is honored unconditionally.

src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+if(!btn)return;\
+e.preventDefault();e.stopPropagation();\
+var pre=btn.closest('details').querySelector('pre');\
+if(!pre||!navigator.clipboard)return;\
+navigator.clipboard.writeText(pre.textContent).then(function(){\
+var orig=btn.textContent;\
+btn.textContent='copied!';btn.classList.add('copied');\
+setTimeout(function(){btn.textContent=orig;btn.classList.remove('copied');},1200);\
+});\
+});";
```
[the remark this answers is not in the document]
prasanna-anchorage: PR description nit; the body actually already says "a vanilla-JS clipboard helper" and "No framework. ~10 lines of plain JS" in a later section, so the framing is internally consistent, just spread across sections. Will tighten the wording in the description but no code change is needed.

src/parser/cli/src/serve.rs
```diff
@@ -0,0 +1,631 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn rel_path(base: &Path, full: &Path) -> String {
+    full.strip_prefix(base)
+        .unwrap_or(full)
+        .display()
+        .to_string()
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed. `rel_path` now joins the `strip_prefix`-ed `components()` with `/` so Windows paths produce `dir/file.hex` to match the URL routes. It also returns `None` if `strip_prefix` fails (pepe-anchor flagged the same fallback path leaking the absolute host path) — callers skip the entry rather than emit a defective rel_path.

src/parser/cli/tests/cli_test.rs
```diff
@@ -386,6 +391,224 @@ fn test_cli_solana_idl_json_mappings() {
...
+    // Use the existing ethereum-from-file.hex as a one-file fixture directory.
+    let fixture_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("tests")
+        .join("fixtures");
+
+    // Ask the OS for a free port: bind ephemeral, drop, race-tolerable since
+    // the server immediately re-binds.
+    let probe = std::net::TcpListener::bind("127.0.0.1:0").expect("probe bind");
+    let port = probe.local_addr().expect("local_addr").port();
+    drop(probe);
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed. The test now passes `--port 0` and parses the actual bound port out of the server-printed `Serving on http://127.0.0.1:N` line. The existing `local_addr()` print already carries the real port, so this drops the probe/drop/rebind race entirely.

src/parser/cli/tests/cli_test.rs
```diff
@@ -386,6 +391,224 @@ fn test_cli_solana_idl_json_mappings() {
...
+    ));
+    fs::create_dir_all(&work).unwrap();
+
+    let valid_hex = "02f86c0180830f4240843b9aca00830186a094111111111111111111111111111111111111111180b844a9059cbb000000000000000000000000000000000000000000000000000000000000dead00000000000000000000000000000000000000000000000000000000000f4240c0";
+    fs::write(work.join("tx.hex"), valid_hex).unwrap();
+    fs::write(work.join("expected.json"), r#"{"sentinel":"first"}"#).unwrap();
+
+    let probe = std::net::TcpListener::bind("127.0.0.1:0").expect("probe bind");
+    let port = probe.local_addr().expect("local_addr").port();
+    drop(probe);
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed. Same `--port 0` + parse-from-stdout pattern as `test_cli_serve_renders_directory`.

src/parser/cli/src/serve.rs
prasanna-anchorage: Fixed in the same commit as the Copilot symlink thread above. `entry.file_type()` is `lstat`-based and `is_symlink()` entries are skipped before recursion. No visited-set/depth-cap needed once symlinks are out of the picture.

src/parser/cli/src/serve.rs
prasanna-anchorage: Fixed in the same commit. `rel_path` now returns `Option<String>` and only emits a path when `strip_prefix` succeeds; callers skip the entry on `None` rather than pushing a `DecodedEntry` with an absolute path. The rendered HTML and route table can no longer carry host filesystem layout.

src/parser/cli/src/serve.rs
prasanna-anchorage: Deferring intentionally: typical `--dir` is the in-repo `tests/fixtures/...` tree (tens of files), which `decode_directory` walks in single-digit ms. The doc comment on the live-reload behavior calls out re-decoding on every request as a deliberate design choice (so editing a file → refresh → see new state with no cache invalidation). If the use case grows to "point it at a hundreds-of-files dir", a single-file fast path that validates the rel_path is canonicalized under `state.dir` and then calls `decode_file` directly is a small follow-up — happy to do it if you want it in this PR, just flag and I will.

src/parser/cli/src/cli.rs
prasanna-anchorage: Fixed. `Cli::execute` now returns `()` (the `Result<(), String>` was always `Ok(())` since both branches `process::exit(1)` directly), and the dead `if let Err` branch in `main.rs` is dropped.

src/parser/cli/src/cli.rs
prasanna-anchorage: Confirmed: searched anchorageoss/* for `parser_cli --chain` and `parser_cli -t` invocations outside this repo via `gh code-search` and found none. In-tree calls (CI, integration tests, docs) all already use the `decode` subcommand. The break is intentional and pre-1.0; happy to ship a one-release transition alias (`parser_cli --chain ... -t ...` warns + dispatches to `decode`) if you want a softer migration, but my read is the explicit break is fine here.

shahan-khatchadourian-anchorage (dismissed):
LGTM

prasanna-anchorage:
Rebased onto `main` and **re-ported onto the post-#316 `parser_cli_core` architecture** — the previous commits were written against the old `parser/cli/src/lib.rs` layout that #316 deleted, so a plain rebase wasn't possible. Force-pushed as a single commit (`bc4869f`); **prior approvals are stale — please re-review.**

What changed vs the old branch:
- `main.rs` now owns the `decode`/`serve` subcommand enum and a shared `ChainArgs` (flattened per-chain args + `build_plugins`). As a result `serve` now inherits the same ABI/IDL mapping flags as `decode`, including **tron** parity (the old branch predated tron).
- Extracted `Runtime` + `prepare_runtime` into `parser_cli_core` (shared by `run` and `serve`) instead of duplicating registry/options wiring; `run` delegates to it.
- `serve::run` returns `Result<(), String>` rather than calling `process::exit`/`expect`, satisfying the workspace no-panic lints (the old `execute_serve` used `expect`).
- Formatting/human-readable output already lives in `cli-core::output` (extracted by #316), so the old branch's `cli.rs` formatter was dropped rather than re-added.

Verified locally: `parser_cli` tests (8) + `serve` unit tests (10) + `parser_cli_core` tests (28) pass; clippy `-D warnings` clean both with and without `--features serve`; `cargo fmt --check` clean.

shahan-khatchadourian-anchorage (approved):
The `serve` subcommand is well-scoped and clean: feature-gated behind `--features serve` so default builds are unaffected, axum/tokio add zero net deps (already transitive via tonic), the `127.0.0.1`-only bind is the right default, and the live-reload-on-refresh design is a nice fit for the fixture-browsing use case. The 10 unit tests in `serve.rs` cover the main code paths (decode, passthrough, error entries, URL encoding, HTML escaping, render output).

One gap vs the test plan: the description says "10 unit tests + 2 integration tests pass" and names `test_cli_serve_live_reload_and_json_passthrough` as an end-to-end test (spawn server, mutate files, verify responses without restart). That test — and a second integration test — don't appear anywhere in the branch. The `cli_test.rs` diff is only the five `"decode"` insertions for the CLI shape change. Worth adding the integration test in a follow-up if it got dropped during rebase.

PR #275, opened 2026-05-01
Add Exponent Finance Solana visualizer preset

kyle-anchorage:
## Why this PR exists

Solana transactions calling the Exponent Finance core program (`ExponentnaRg3CQbW6dqQNZKXp7gtZ9DGMp1cwC4HAS7`) currently fall through to the generic `unknown_program` visualizer, so wallet sign prompts show only the program ID and raw hex instruction data. This adds a dedicated preset that surfaces the instruction name, named accounts, and decoded args.

## What changed

- New preset crate module `presets::exponent_finance` registered under `visualsign-solana`.
- IDL JSON (`exponent_finance.json`) sourced from the published `@exponent-labs/exponent-idl` npm package, which is the same data the Exponent SDK and orbmarkets explorer ship.
- Standard generic-IDL scaffolding: `decode_idl_data` -> `parse_instruction_with_idl` -> `build_named_accounts` -> field builders, mirroring the `dflow_aggregator` preset.
- Visualizer kind: `Dex` ("Exponent Finance"). Config matches all instructions for the program ID via `("*", vec!["*"])`.

image

Two adaptations relative to the standard Anchor-style scaffold:
- Exponent uses Anchor 0.31's `#[instruction(discriminator = [N])]` attribute, so on-chain instruction discriminators are 1 byte rather than 8. The minimum-data-length check is `data.is_empty()` and the `idl_has_discriminators` test only asserts `!disc.is_empty()`. The underlying `solana_parser::find_instruction_by_discriminator` already supports variable-length discriminators.
- The IDL's `Number` type is a tuple-struct (a single anonymous `[u64; 4]` field). `solana_parser`'s `IdlField` schema requires named fields, so that one entry was rewritten in the JSON to `{"name": "value", "type": {"array": ["u64", 4]}}`. No other types or instructions were modified.

## Why this is safe

- Read-only feature: the parser only reads transaction bytes and produces display fields. No signing, custody, or settlement code is touched.
- Backward compatible: adds a new preset module and a single registration line in `presets/mod.rs`. Existing presets are unaffected. Programs that don't match the Exponent program ID continue to dispatch to their existing visualizer.
- Failure mode is graceful: if a future Exponent instruction has an unrecognized 1-byte discriminator, `parse_exponent_finance_instruction` returns `Err`, and the visualizer falls through to `build_fallback_fields` (which still emits the raw hex). This matches the `dflow_aggregator` pattern.
- The `Number` IDL patch is a structural rename only; the underlying type expression `{"array": ["u64", 4]}` is preserved, so any instruction that references `Number` will still decode the same bytes.

#### Checks run (by agent)
- `cargo clippy -p visualsign-solana --all-targets -- -D warnings` -> clean
- `cargo test -p visualsign-solana` -> 70 unit tests pass (including the 4 new `exponent_finance` tests: idl_loads, idl_has_discriminators, unknown_discriminator_returns_error, short_data_returns_error)
- `make -C src test` -> all workspace tests pass
- `make -C src lint` -> clean across all crates

#### Manual steps needed (by human)
None - fully automated by CI.

## How this is maintainable

- The preset follows the same shape as `dflow_aggregator`, so a developer touching either picks up the pattern from one place. The two deviations (1-byte discriminator handling and the `Number`-type rewrite) are documented inline by the test names and at the top of this PR.
- `build.rs` auto-discovers `ExponentFinanceVisualizer` from the new directory, so adding/removing presets doesn't require touching a registry.
- The IDL JSON is sourced from the upstream `@exponent-labs/exponent-idl` package; refreshing the IDL is a copy-paste of a new release plus re-applying the `Number`-field rewrite (the test `test_exponent_finance_idl_loads` will fail loudly if the rewrite is missed).
- Workspace lint policy (`unwrap_used = "deny"`, etc.) is enforced - the preset uses `?` and `ok_or`/`ok_or_else` for all fallible paths.
- The 4 sanity tests guard the discriminator/IDL contract: any future change that breaks IDL load, drops a discriminator, or stops rejecting empty/garbage input will surface in CI.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/exponent_finance/config.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/exponent_finance/config.rs
@@ -1,0 +1,22 @@
+use super::EXPONENT_FINANCE_PROGRAM_ID;
+use crate::core::{SolanaIntegrationConfig, SolanaIntegrationConfigData};
+use std::collections::HashMap;
+
+pub struct ExponentFinanceConfig;
+
+impl SolanaIntegrationConfig for ExponentFinanceConfig {
+    fn new() -> Self {
+        Self
+    }
+
+    fn data(&self) -> &SolanaIntegrationConfigData {
+        static DATA: std::sync::OnceLock<SolanaIntegrationConfigData> = std::sync::OnceLock::new();
+        DATA.get_or_init(|| {
+            let mut programs = HashMap::new();
+            let mut instructions = HashMap::new();
+            instructions.insert("*", vec!["*"]);
+            programs.insert(EXPONENT_FINANCE_PROGRAM_ID, instructions);
+            SolanaIntegrationConfigData { programs }
+        })
+    }
+}
--- a/src/chain_parsers/visualsign-solana/src/presets/exponent_finance/exponent_finance.json
+++ b/src/chain_parsers/visualsign-solana/src/presets/exponent_finance/exponent_finance.json
@@ -1,0 +1,7197 @@
+{
+  "address": "ExponentnaRg3CQbW6dqQNZKXp7gtZ9DGMp1cwC4HAS7",
+  "metadata": {
+    "name": "exponent_core",
+    "version": "0.1.0",
+    "spec": "0.1.0",
+    "description": "Created with Anchor"
+  },
+  "instructions": [
+    {
+      "name": "add_emission",
+      "discriminator": [
+        18
+      ],
+      "accounts": [
+        {
+          "name": "authority",
+          "signer": true
+        },
+        {
+          "name": "fee_payer",
+          "writable": true,
+          "signer": true
+        },
+        {
+          "name": "vault",
+          "writable": true
+        },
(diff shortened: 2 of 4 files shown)


prasanna-anchorage (approved):
Approving. Rebased onto current main — the diff was reverting #284's `pub mod kamino_limit;` registration because the PR predates that merge. Resolved the `presets/mod.rs` conflict by keeping both `exponent_finance` and `kamino_limit` in alphabetical order.

Verification:
- `cargo clippy -p visualsign-solana --all-targets -- -D warnings`: clean.
- 4 exponent_finance preset tests pass (`test_exponent_finance_idl_loads`, `test_exponent_finance_idl_has_discriminators`, `test_unknown_discriminator_returns_error`, `test_short_data_returns_error`).
- Same preset template as the 11 other Kyle-authored Solana presets already on main.

`config.rs` uses `HashMap` per the current convention — #288's sweep will flip it to `BTreeMap` when that PR lands.

PR #284, opened 2026-05-07
Add Kamino Limit Orders Solana preset

kyle-anchorage:
## Why this PR exists

Kamino Limit Orders (program `LiMoM9rMhrdYrfzUCxQppvxCSG1FcrUK9G8uLq4A1GF`) is a Solana DEX/limit-order program with no preset in this parser. Its instructions currently fall through to `unknown_program` and surface as raw bytes, so wallets cannot show users a human-readable description of operations like `createOrder`, `takeOrder`, or `closeOrderAndClaimTip`.

## What changed

- New preset module `chain_parsers/visualsign-solana/src/presets/kamino_limit/` with three files: `mod.rs` (visualizer), `config.rs` (program-instruction allowlist), and `kamino_limit.json` (IDL bundled at compile time).
- The visualizer is the generic IDL-driven pattern used by the other Kamino presets: it parses the discriminator, names accounts from the IDL, decodes args, and falls back to raw-data display on parse failure.
- Registered alphabetically in `presets/mod.rs`. `VisualizerKind::Dex(...)` since this is an order-book program.

image

## Why this is safe

A new preset adds a new `InstructionVisualizer`; no existing program's output changes. The decoder uses the same generic IDL helpers as `kamino_vault`/`kamino_borrow`/`kamino_farms`, with no Limit-Orders-specific code paths, so the failure modes are the same as those presets. If the IDL fails to parse (corruption, schema change), the visualizer falls back to a raw-data field — the same fallback every other IDL-based preset uses.

The IDL JSON was fetched directly from the on-chain Anchor IDL account via `anchor idl fetch ... --provider.cluster mainnet`. A fresh re-fetch produced a byte-identical 40191-byte file (sha256 `d57e19f4...46ed94`), confirming the bundled file matches the canonical on-chain source — the same source Orb's `/anchor-idl` page reads via Helius RPC.

#### Checks run (by agent)
- `make -C src build` — green
- `make -C src lint` (clippy with `-D warnings`) — green
- `make -C src test` — all workspace tests pass
- Four preset-specific unit tests pass: IDL loads, every one of the 15 instructions has an 8-byte discriminator, unknown discriminator returns an error, sub-discriminator-length data returns an error
- Two independent `anchor idl fetch` invocations produced byte-identical output

#### Manual steps needed (by human)
None — fully automated by CI.

## How this is maintainable

The preset follows the same shape as the three other Kamino presets already in this repo, so a fresh agent or developer who reads any of `kamino_vault`, `kamino_borrow`, or `kamino_farms` will know how to update this one. `build.rs` auto-discovers `KaminoLimitVisualizer` from the directory name, so the only manual registration is the one line in `presets/mod.rs`. Workspace clippy (`unwrap_used = deny`, `expect_used = deny`, `panic = deny`) catches regressions, and the module's tests assert the bundled IDL keeps loading and every instruction carries a valid 8-byte discriminator if the JSON is ever refreshed.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/kamino_limit/config.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/kamino_limit/config.rs
@@ -1,0 +1,22 @@
+use super::KAMINO_LIMIT_PROGRAM_ID;
+use crate::core::{SolanaIntegrationConfig, SolanaIntegrationConfigData};
+use std::collections::HashMap;
+
+pub struct KaminoLimitConfig;
+
+impl SolanaIntegrationConfig for KaminoLimitConfig {
+    fn new() -> Self {
+        Self
+    }
+
+    fn data(&self) -> &SolanaIntegrationConfigData {
+        static DATA: std::sync::OnceLock<SolanaIntegrationConfigData> = std::sync::OnceLock::new();
+        DATA.get_or_init(|| {
+            let mut programs = HashMap::new();
+            let mut instructions = HashMap::new();
+            instructions.insert("*", vec!["*"]);
+            programs.insert(KAMINO_LIMIT_PROGRAM_ID, instructions);
+            SolanaIntegrationConfigData { programs }
+        })
+    }
+}
--- a/src/chain_parsers/visualsign-solana/src/presets/kamino_limit/kamino_limit.json
+++ b/src/chain_parsers/visualsign-solana/src/presets/kamino_limit/kamino_limit.json
@@ -1,0 +1,1799 @@
+{
+  "accounts": [
+    {
+      "name": "Order",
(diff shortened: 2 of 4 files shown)


prasanna-anchorage (approved):
Approving. Additive new preset (Kamino Limit Orders, program `LiMoM9rMhrdYrfzUCxQppvxCSG1FcrUK9G8uLq4A1GF`); 4 files, all isolated to `presets/kamino_limit/`. Same template as the 10+ already-merged Solana presets. Rebased onto current main, clippy clean, 4 preset tests + the rest of the visualsign-solana suite pass locally. `config.rs` uses `HashMap` per existing convention; #288's sweep will flip it to `BTreeMap` when that PR lands.

PR #286, opened 2026-05-08
fix(solana/dflow-aggregator): surface remaining accounts and flatten nested args

pepe-anchor:
## Summary

When a DFlow Aggregator instruction includes more accounts than the IDL defines, the extras were silently dropped. This surfaces them as labeled "Remaining Account N" fields in the expanded view.

Also replaces the flat `format_arg_value` string formatter with a recursive `push_arg_fields` that decomposes nested JSON objects/arrays into dotted/indexed field labels instead of stringifying them as blobs.

## Changes

- `build_named_accounts` returns extra accounts alongside named ones
- `push_arg_fields` recursively flattens nested JSON args (`key.sub`, `key[0]`)
- Inline `append_raw_data` helper, drop empty-string subtitle, log `tracing::warn` on parse failure

## Test evidence

Unit tests cover scalar rendering, array/object recursion, empty collections, extra-account surfacing, and human-readable label format.

_Co-authored with Claude._

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/dflow_aggregator/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/dflow_aggregator/mod.rs
@@ -63,7 +63,9 @@ impl InstructionVisualizer for DflowAggregatorVisualizer {
 
         let preview_layout = SignablePayloadFieldPreviewLayout {
             title: Some(SignablePayloadFieldTextV2 { text: title }),
-            subtitle: None,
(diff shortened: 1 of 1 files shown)


prasanna-anchorage (approved):
Approving with the fixup pushed: subtitle restored to convention, test `use` hoisted to module top, and the trailing tautological Pubkey::from_str assertion replaced with a real check that the Remaining Account fields render the correct pubkey values. Clippy clean, all 10 dflow tests pass locally.

Issue #228, opened 2026-03-30
feat: attested transaction lint diagnostics in SignablePayload


shahan-khatchadourian-anchorage:
## Summary

Add a lint framework for transaction parsing that produces structured diagnostics as part of the signed `SignablePayload`. Diagnostics are attested by the enclave's ephemeral key alongside the parsed fields, making them tamper-proof and auditable. Diagnostics are expressed as a field variant in `SignablePayload`, alongside existing variants like `TextV2` or `AmountV2`.

Motivated by [this comment](https://github.com/anchorageoss/visualsign-parser/pull/204#discussion_r2120429461) on PR #204: instructions with corrupted account indices are currently silently dropped, hiding data integrity problems from the caller.

### Diagnostics in the signed payload

Diagnostics become part of `SignablePayload` — either as a new `SignablePayloadField` variant or a dedicated field on the struct. Either way, they're included in the deterministic JSON serialization, covered by the Borsh hash, and signed by the ephemeral P256 key.

This means:
- The HSM/attester can verify what the parser flagged
- A wallet cannot suppress or modify diagnostics without breaking the signature
- Auditors can inspect diagnostics after the fact

### Structured diagnostic format

Each diagnostic is a typed, machine-readable record designed to be consumed equally well by wallet UIs, CI tools, and AI agents:

\`\`\`json
{
  "rule": "transaction::oob_account_index",
  "domain": "transaction",
  "level": "warn",
  "message": "instruction 2: account index 7 out of bounds (5 accounts available)",
  "instruction_index": 2
}
\`\`\`

Human-readable rendering is derived from this structure, not the other way around. AI tools can filter by domain, triage by severity, and locate by instruction index without parsing natural language.

### Severity levels

- **Error** — indicates a serious problem
- **Warn** — notable issue, parsing continues
- **Allow** — silently handled, not included in output

### All rules are configurable

Every rule's severity can be overridden by the caller via the request. This is a reporting tool — the caller decides what severity level matters for their use case. Configuration comes from the request side (in \`ChainMetadata\` or \`VisualSignOptions\`), and the resulting diagnostics are attested in the response.

### Lint configuration and the attestation flow

Lint configuration (severity overrides) would be part of the request metadata, which means it's included in the \`metadata_digest\` — the SHA-256 of the borsh-serialized chain metadata. The attestation flow becomes:

\`\`\`
Wallet sends:
  ParseRequest {
    unsigned_payload,
    chain,
[...]

PR #301, opened 2026-05-14
fix(parser_cli): keep text display fixture diagnostics-agnostic

shahan-khatchadourian-anchorage:
## Summary

- Stagex's `parser_cli` build runs `cargo test --no-default-features --features solana`, which disables the default `diagnostics` feature. The post-#228 `solana-text.display.expected` baked in the new `Diagnostic { ... }` blocks, so when the binary ran without diagnostics emitted, the actual output diverged from the fixture and `test_cli_with_fixtures` failed inside the container (visible on main's most recent `stagex` job for `1e4fe64`).
- Mirror the separation the JSON path already enforces: display fixtures contain only display fields; diagnostics are tested via the diagnostics fixture. The text path now strips `Diagnostic { ... },` blocks from the Debug dump before comparing (gated on the `diagnostics` feature).

## Test plan

- [x] `cd src && cargo test -p parser_cli` (default features: diagnostics on) — 8 tests pass
- [x] `cd src && cargo test -p parser_cli --no-default-features --features solana` (stagex config) — 3 tests pass
- [ ] Confirm green stagex run on this PR

### DIFF
--- a/src/parser/cli/tests/cli_test.rs
+++ b/src/parser/cli/tests/cli_test.rs
@@ -203,16 +203,53 @@ fn test_cli_with_fixtures() {
                 }
             }
             Err(_) => {
-                // Non-JSON output (text/human): plain string comparison
+                // Non-JSON output (text/human): strip diagnostic blocks from the
+                // actual Debug-formatted payload so the display fixture stays
+                // diagnostics-agnostic, matching how the JSON branch filters them above.
+                #[cfg_attr(not(feature = "diagnostics"), allow(unused_mut))]
+                let mut actual_display = actual_output.trim().to_string();
+                #[cfg(feature = "diagnostics")]
+                {
+                    actual_display = strip_debug_diagnostic_blocks(&actual_display);
+                }
                 assert_strings_match(
                     test_name,
                     "display",
                     expected_display.trim(),
-                    actual_output.trim(),
+                    &actual_display,
                 );
             }
         }
     }
+}
+
+/// Remove `Diagnostic { ... },` blocks from a Rust `{:#?}` payload dump.
+/// The blocks live inside the `fields:` array at 8-space indent, so balanced-brace
+/// counting from each `        Diagnostic {` line drops the whole entry.
+#[cfg(feature = "diagnostics")]
+fn strip_debug_diagnostic_blocks(input: &str) -> String {
+    let mut out = String::with_capacity(input.len());
+    let mut depth: i32 = 0;
+    for line in input.lines() {
+        if depth == 0 && line == "        Diagnostic {" {
+            depth = 1;
+            continue;
+        }
+        if depth > 0 {
+            depth += line.matches('{').count() as i32;
+            depth -= line.matches('}').count() as i32;
+            if depth == 0 {
+                continue;
+            }
+            continue;
+        }
+        out.push_str(line);
+        out.push('\n');
+    }
+    if !input.ends_with('\n') && out.ends_with('\n') {
+        out.pop();
+    }
+    out
 }
 
 fn assert_strings_match(test_name: &str, fixture_type: &str, expected: &str, actual: &str) {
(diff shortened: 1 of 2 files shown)


prasanna-anchorage (approved):
Approved.

Reviewed in a worktree at `1656bb8`. Both feature configs pass locally:

- `cd src && cargo test -p parser_cli --no-default-features --features solana` — 3 tests pass (this is the stagex config that was failing on `main`).
- `cd src && cargo test -p parser_cli` — 8 tests pass (default, diagnostics on).

CI is fully green on the PR including `Build parser_cli (Solana)`, which confirms the stagex break from #255 is resolved.

The fixture/test pattern now mirrors the JSON path's structure (display fixture is diagnostics-agnostic; the diagnostics-on branch filters before comparing). One trade-off worth noting (not a blocker): the text path no longer pins the diagnostic Debug output the way it briefly did between #255 and this PR — the diagnostic *content* is still pinned via `solana-json.diagnostics.expected` though, and the Rust Debug format is intentionally not a stable surface to pin, so I think dropping the text-diagnostics pin is the right call.

Two minor inline notes (one style, one future-proofing) on the new helper — both non-blocking.

src/parser/cli/tests/cli_test.rs
```diff
@@ -203,18 +203,55 @@ fn test_cli_with_fixtures() {
...
+            depth = 1;
+            continue;
+        }
+        if depth > 0 {
+            depth += line.matches('{').count() as i32;
+            depth -= line.matches('}').count() as i32;
+            if depth == 0 {
+                continue;
+            }
+            continue;
```
prasanna-anchorage, line 244: Style nit: the two `continue;` statements at the end of the `if depth > 0` block are redundant — both arms continue regardless of whether `depth` just dropped to zero. Collapses to one `continue;` after the depth check.

src/parser/cli/tests/cli_test.rs
```diff
@@ -203,18 +203,55 @@ fn test_cli_with_fixtures() {
...
+fn strip_debug_diagnostic_blocks(input: &str) -> String {
+    let mut out = String::with_capacity(input.len());
+    let mut depth: i32 = 0;
+    for line in input.lines() {
+        if depth == 0 && line == "        Diagnostic {" {
+            depth = 1;
+            continue;
+        }
+        if depth > 0 {
+            depth += line.matches('{').count() as i32;
```
prasanna-anchorage, line 239: Future-proofing observation, not a blocker: brace counting matches `{` and `}` anywhere on the line, including inside string literals. The current diagnostic messages do not contain braces (`"all 6 instructions have valid program_id_index"` etc.), but a future diagnostic that interpolates braces in its message would skew `depth`. If you ever generalize this beyond the current Diagnostic block, consider walking lines until the matching `},` at the same indent rather than counting braces. For the current shape it is fine.

prasanna-anchorage (approved):
Approved.

Reviewed in a worktree at `1656bb8`. Both feature configs pass locally:

- `cd src && cargo test -p parser_cli --no-default-features --features solana` — 3 tests pass (the stagex config that was failing on `main`).
- `cd src && cargo test -p parser_cli` — 8 tests pass (default, diagnostics on).

CI is fully green on the PR including `Build parser_cli (Solana)`, which confirms the stagex break from #255 is resolved.

The fixture/test pattern now mirrors the JSON path's structure (display fixture is diagnostics-agnostic; the diagnostics-on branch filters before comparing). One trade-off worth noting (not a blocker): the text path no longer pins the diagnostic Debug output the way it briefly did between #255 and this PR — the diagnostic *content* is still pinned via `solana-json.diagnostics.expected` though, and the Rust Debug format is intentionally not a stable surface to pin, so dropping the text-diagnostics pin is the right call.

Two minor inline notes (one style, one future-proofing) on the new helper — both non-blocking.

PR #304, opened 2026-05-16
feat(x402): TVC-enforced payment via gateway-signed VPM (stacked on #295)

prasanna-anchorage:
## Summary

Stacked on top of #295 (`spec/x402-gated-http-api`). **Do not merge until #295 lands.**

Ships the TVC-enforced x402 v2 trust pair: a host-side `parser_gateway` that does verify → settle → sign VPM → forward, and an enclave-side `parser_http_server` that runs as the TVC pivot and refuses every parse request that doesn't carry a `VerifiedPaymentMarker` signed by the pinned gateway pubkey.

The PR also covers the deployment shape it takes to actually run this. The original three commits (TVC-enforced VPM, gateway integration test coverage, compose wiring) landed first and the rest are the operational work that came out of standing the stack up against real TVC and real on-chain settlement.

### Trust model

- TVC enclave is x402-aware; rejects unpaid requests with a native 402 (gateway materializes the canonical PaymentRequired body from the gRPC `FailedPrecondition`).
- Gateway's signing pubkey is the trust anchor, pinned in `parser_http_server`'s deploy via `--gateway-signing-pubkey-hex` (or the legacy `GATEWAY_SIGNING_PUBKEY_HEX` env when running outside TVC).
- Facilitator stays external — payai for real, `mock_facilitator` for local — the gateway hand-rolls the call order (verify → settle → sign VPM → forward) so the post-settle txid is in scope when it signs the marker.

## Protocol version: x402 **v2**

This PR speaks x402 **v2** end-to-end. The wire conventions are all v2:

- Payment requirements ride in the **`Payment-Required` response header** (base64-JSON), not in the body.
- The client returns payment in the **`Payment-Signature` request header** (base64-JSON of the v2 `PaymentPayload`), replacing v1's `X-PAYMENT`.
- Every gateway ↔ facilitator `verify`/`settle` body carries `"x402Version": 2` (`parse_tvc.rs:142,335`).

To avoid version-string ambiguity elsewhere in the diff:

- `/visualsign/api/v1/parse` and `/visualsign/api/v2/parse` are **Anchorage's parse-API surface versions**, not x402 protocol versions. `/v1` is the open route; `/v2` is the same parse logic gated by x402 v2.
- "**v3.0 / v3.1**" mentions below refer to the **internal trust-model revision** of this work.

## What landed beyond the original VPM commit

These commits all sit on `spec/x402-tvc-enforced` since the original PR opened. Each one was driven by an actual blocker hit while bringing the stack up:

- `5147793` — **`parser_http_server` takes CLI args**. `tvc deploy create` has no env-var-injection mechanism, so the pivot's startup config must arrive via `pivotArgs`. clap derive with `env = "..."` fallback keeps compose + integration tests working.
- `703f43d` — **chain feature flags on `parser_http_server`** mirroring parser_cli #302. `--build-arg CHAIN_FEATURES=solana` ships a single-chain pivot (smaller binary, smaller attestation surface).
- `1171203` — **build the pivot with `qos_core/vm`**. Without this, `EPHEMERAL_KEY_FILE` resolved to the dev path; the pivot crash-looped at startup and the TVC deploy stuck at `0/3 healthy`. Now the pivot reads `/qos.ephemeral.key` as QOS expects.
- `a8cdae9` — **derive the 402 payment network from the parse request's chain**. `CHAIN_ETHEREUM → base-sepolia` (or `base`); `CHAIN_SOLANA → solana-devnet` (or `solana`); anything else → 400 Bad Request with `"x402 payment not available for chain X"`. Local profile auto-seeds both EVM and Solana tags so a single gateway answers both chains.
- `d0e78c5` — **EVM payment requirements actually parseable** by `@x402/evm`. Previous EVM 402 emitted `asset: "USDC"` (symbol, not contract) with no `extra` block; the EVM client's `getAddress(asset)` and EIP-712 domain lookup both failed. Now emits CAIP-2 network, USDC contract address, and `extra.name`/`extra.version`.
- `55c8650` — **`translate_to_canonical` pulls USDC values from `x402_chain_eip155::USDC`** instead of hand-rolling. Same crate the price-tag side already uses, so the two paths can't drift. Fixes a latent bug where Base mainnet was emitting `name: "USDC"` while Circle's domain on Base mainnet is `"USD Coin"` — silent signature verification failure on real money.
- `e24da5d` — **`scripts/redeploy-parser-http-server.sh`** — pull-image → recompute-digest → edit-config → `tvc deploy create`. Reads the canonical `pr-<N>` tag because GitHub PR runs build the merge commit, not the branch tip.
- `fed4912` + `1ff3afa` — **e2e runbook** at `TVC_E2E_DEPLOYMENT_TEST.md` plus 15 redacted fixtures under `tvc-e2e-fixtures/` covering the smoke path, the local-gateway loop, the Solana-devnet variant, and the Base-Sepolia variant (added in `1b81f26`).
- `9e3dc1a` (cherry-picked from #306) — **stagex emits `expectedPivotDigest` as pure hex**. Previously emitted `sha256:<hex>` and the prefix tripped `tvc deploy approve` with `OddLength at line 8 column 85`.

## Wire shape

`ParseRequest` gains a `bytes payment_marker` (Borsh-encoded `SignedVerifiedPaymentMarker` from `host_primitives::payment_marker`). The VPM commits to `request_hash`, on-chain `txid`, `payer`, `pay_to`, `amount`, `mint`, `x_payment_hash`, and the gateway's signing pubkey. `parser_app` verifies the gateway's P256 signature + binds the marker to the specific parse body.

## End-to-end evidence

**Local + Solana devnet (payai)** — settlement `2gnsP9SkErqXjkEa9TjGwG6b8P5iutBjpRRFk2xowYxBjwoZPTFyNYChgZAabbGXWvyL3LzU9oAgru1vC1WNZwoS`; 200 with the enclave-signed parse payload, `enclavePubKey` matching `tvc deploy provisioning-details`.

**Local + Base Sepolia (payai)** — settlement `0x57c30a481cf86012e0a816ef1aaec55830edb593d447e8dcda138a1cc16cb788` (and earlier `0x61a3afa9af9c1f6fa1c9d0d4b8bda942096894786a7fac3719073c2e5efccd93`); real EIP-3009 `transferWithAuthorization` from the buyer wallet, two-signer tx via Multicall3, payai pays gas. `https://sepolia.basescan.org/tx/0x57c30a48…`.

**Tamper path** — `parser_app` pinned to gateway key B, gateway signs with A → parser rejects with `FailedPrecondition` → gateway translates to HTTP 402, mock facilitator's `/debug/settle_count` confirms no settlement happened. Covered by integration test `path6_tampered_pubkey_returns_502_no_settle`.

## Known v3.0 quirks

- Operator pubkey mismatch (gateway signs with X, parser pinned to Y) still results in lost devnet USDC — payai settles before parser verifies. v3.1 will close this by binding settle to a pre-validation step.
- Payer's EVM/Solana signature on the underlying transfer is verified at payai's `/verify` before settle, and the VPM commits to `x_payment_hash` (binding the marker to the exact `X-PAYMENT` body), but the enclave doesn't re-verify the underlying transfer signature. v3.1 adds that as defense-in-depth.

## Test plan

- [x] `cargo test -p parser_gateway --lib` — 31/31 (including 12-case `network_matches_chain` truth table, 6 `translate_to_canonical` cases covering both EVM and Solana branches plus the Base-mainnet `"USD Coin"` regression guard, plus the original signing + attestation tests)
- [x] `cargo test -p parser_app --lib payment_verify` — 6/6 (disabled accepts, required accepts valid, rejects missing/tampered/mismatched)
- [x] `cargo test -p host_primitives` — 3/3 (VPM borsh round trip, deterministic digest, request_hash sensitivity)
- [x] `cargo test --test x402_gateway_test` — 9/9 (paths 1-6 cover the original integration shape; 7-9 cover the per-chain derive: EVM tags only on ETH, Solana tags only on SOL, 400 on unsupported chain)
- [x] `cargo clippy --all-targets -- -D warnings` clean on `parser_gateway` and `integration`
- [x] Manual Solana devnet e2e against payai (happy + tamper, signatures recorded above)
- [x] Manual Base Sepolia e2e against payai (two paid runs, real on-chain settlement, signatures recorded above)
- [x] `TVC_E2E_DEPLOYMENT_TEST.md` covers all three variants (direct smoke, local mock-fac loop, payai-Solana, payai-Base-Sepolia)


pepe-anchor:
Closing this out. It's stacked on `spec/x402-gated-http-api` (#295), which got closed unmerged and is now 132 commits behind main, so this can't land as is.

The work itself is ported and rebased onto main across a few smaller PRs instead of one big one:

- #448 (merged) - `parser_app` payment-marker verification
- #449 - `parser_gateway` x402 gating, attestation verifier, module split
- #450 - `parser_http_server` HTTP+JSON pivot for TVC public ingress
- #451 - X-Stamp validation in the enclave
- #452 - real bootProof from NSM

Keeping this open as reference for the original design (VPM wire shape, trust model writeup, the e2e evidence against payai) but the code itself lives in the PRs above now.

PR #309, opened 2026-05-21
chore(parser_app): address PR #303 follow-up review comments

shahan-khatchadourian-anchorage:
## Summary

Addresses all four follow-up items prasanna-anchorage raised on his approval of #303 (the per-PR narrow-build-check + local feature-matrix-check work). Single PR — all items are small, touch the same Makefile, and don't depend on each other.

## Changes

**1. Restore dynamic chain discovery (`$(sort)` + `$(wildcard)`).**
```
PARSER_APP_CHAINS := $(sort $(patsubst visualsign-%,%,$(notdir $(wildcard chain_parsers/visualsign-*))))
```
The hardcoded list from #303 required editing both the Makefile variable and `parser_app/Cargo.toml`'s `[features]` block whenever a chain was added. The sorted wildcard gives the same stable, deduplicated ordering the depth-2 pair generator needs, and the matrix auto-extends when a new `chain_parsers/visualsign-<name>/` lands.

**2. `cargo build` → `cargo check` in both targets.**
The regression class these targets guard — unconditional chain-crate references that should have been `#[cfg]`-gated — is compile-level. `cargo check` is sufficient and roughly 2× faster than `cargo build`; the clippy invocations already cover lint coverage. As a bonus, `feature-matrix-check` now grows the target dir by ~2GB instead of ~28GB across its 31 invocations.

**3. Add empty + diagnostics combo to `feature-matrix-check`.**
`--no-default-features --features diagnostics` (no chains compiled in) makes the diagnostics axis symmetric with the rest of the matrix.

**4. Tighten comments.**
- `narrow-build-check` comment now reflects the auto-pickup behavior the dynamic variable provides.
- CI cache note rewritten to be accurate: parser_app re-type-checks under each distinct feature set while workspace dependencies are reused from earlier CI steps' cache.

## Test plan

- [x] `make -C src narrow-build-check` — 6 `cargo check` + 6 clippy invocations, exit 0.
- [x] `make -C src feature-matrix-check` — 31 `cargo check` invocations (1 diagnostics-only + 5 singles + 5 singles+diagnostics + 10 pairs + 10 pairs+diagnostics), exit 0. Cargo target dir grew by ~2GB.
- [x] Pre-commit hook (cargo fmt + cargo clippy across all workspace splits): clean.

## References

- PR #303 (merged): https://github.com/anchorageoss/visualsign-parser/pull/303
- prasanna-anchorage's approval comment with the follow-up asks

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -7,12 +7,18 @@ PARSER_INNER_SOCKET_PATH ?= /tmp/inner_parser.sock
 
 ROOT := $(shell git rev-parse --show-toplevel)
 
-# Chains parser_app can be built with. Single source of truth for both
-# narrow-build-check and feature-matrix-check below; keep in sync with
-# the [features] block in parser_app/Cargo.toml. Listed alphabetically
-# so the depth-2 pair generator in feature-matrix-check produces stable,
-# unique pairs.
-PARSER_APP_CHAINS := ethereum solana sui tron unspecified
+# Chains parser_app can be built with -- derived from the
+# `chain_parsers/visualsign-<name>/` directory layout that the workspace
+# already uses as its source of truth. `$(sort ...)` deduplicates and
+# alphabetizes so the depth-2 pair generator in feature-matrix-check
+# produces stable, unique pairs across machines (filesystem-listing
+# order isn't guaranteed otherwise).
+#
+# Convention: every `chain_parsers/visualsign-<name>/` directory has a
+# matching `<name>` feature in parser_app/Cargo.toml. A chain crate added
+# without that feature wired in fails loudly at narrow-build-check on
+# the PR that introduces it -- which is the right place to catch it.
+PARSER_APP_CHAINS := $(sort $(patsubst visualsign-%,%,$(notdir $(wildcard chain_parsers/visualsign-*))))
 
 .PHONY: all
 all: generated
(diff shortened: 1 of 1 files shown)


prasanna-anchorage (approved):
Approving — all four follow-ups from my #303 review are correctly addressed, and I verified each:

1. **Dynamic chain discovery** — `$(sort $(patsubst visualsign-%,%,$(notdir $(wildcard chain_parsers/visualsign-*))))` expands to exactly `ethereum solana sui tron unspecified`, identical to the old hardcoded list, and parser_app's `[features]` block has a matching feature per chain. The `$(sort)` gives the stable, deduped ordering the pair generator needs. The "add a chain crate without wiring its feature -> fails loudly at narrow-build-check" behavior is exactly what I wanted.
2. **`cargo build` -> `cargo check`** — correct: the regression class (unconditional, un-cfg-gated chain-crate refs) is compile-level, so `check` is sufficient and clippy still covers the lint axis. Nice target-dir savings too.
3. **diagnostics-only combo** — verified `cargo check -p parser_app --no-default-features --features diagnostics` compiles locally; makes the diagnostics axis symmetric. Invocation count math checks out (31 = 1 + 5 + 5 + 10 + 10).
4. **Comments** — accurate now, including the CI-cache note.

No conflicts with main, CI green.

Non-blocking nit: the new standalone `echo "==> features: diagnostics"` line isn't `@`-prefixed, so make echoes the `echo` command itself before running it. Harmless and roughly consistent with how these targets already surface their `cargo` lines, but `@echo` would be tidier. Fine to leave.

LGTM.

Issue #313, opened 2026-05-22
tron: per-contract parse errors now fatal where they used to be silently dropped


prasanna-anchorage:
## Context

PR https://github.com/anchorageoss/visualsign-parser/pull/314 replaced the pre-existing pattern

\`\`\`rust
if let Ok(transfer) = TransferContract::parse_from_bytes(&parameter.value) {
    // push fields for transfer
}
\`\`\`

with

\`\`\`rust
let transfer = TransferContract::parse_from_bytes(value).map_err(|e| {
    VisualSignError::ConversionError(format!(\"decode TransferContract: {e}\"))
})?;
\`\`\`

in every contract arm of \`decode_contract\` (\`src/chain_parsers/visualsign-tron/src/lib.rs\`). Same change applied to the five new v2-staking contract arms (\`FreezeBalanceV2\`, \`UnfreezeBalanceV2\`, \`WithdrawExpireUnfreeze\`, \`DelegateResource\`, \`UnDelegateResource\`).

## Behavior change

- **Old**: A \`TransferContract\` whose inner protobuf bytes failed to parse silently skipped the data fields, leaving the envelope (\`Network\`, \`Timestamp\`, \`Expiration\`, \`Fee Limit\`, \`Ref Block\`, \`Ref Block Hash\`) but no \`Contract Type\` field. The function returned \`Ok\` with a degraded payload.
- **New**: \`?\` propagates a \`VisualSignError::ConversionError\`. The whole transaction conversion fails; \`parser_cli\` prints \`Error: …\` and exits non-zero; \`parser_app\` (gRPC) returns an error response.

## Why this is worth tracking

This is **arguably the better behavior** (no silent half-decode → the signer sees a clear error rather than approving a stripped payload). But it is a contract change for any library consumer of \`transaction_to_visual_sign\` / \`transaction_string_to_visual_sign\` that relied on resilient partial-decode behavior — e.g., a wallet that wanted to show \"unknown contract from a newer Tron upgrade\" with as much envelope data as possible instead of failing closed.

A separate hardening (PR follow-up) already handled the \"contract has no parameter\" case explicitly with a \`<missing parameter>\` marker rather than dropping it. We may want the same treatment for the \"contract has parameter but inner parse failed\" case — surface it as e.g. \`Contract Type: TransferContract (decode failed: <reason>)\` with the envelope intact rather than failing the whole conversion.

## Action

[...]

Issue #312, opened 2026-05-22
tron: SUN-denominated amounts no longer appear in payload text (information loss)


prasanna-anchorage:
## Context

Prior to the modernization PR (https://github.com/anchorageoss/visualsign-parser/pull/314), Tron's `Fee Limit` and `TransferContract.Amount` fields rendered as text:

\`\`\`
Fee Limit: 10000000 SUN (10 TRX)
Amount:    7000000 SUN (7 TRX)
\`\`\`

After the PR, they render as the workspace-standard `AmountV2` variant carrying only the TRX value:

\`\`\`json
{
  "AmountV2": {
    "common": { "label": "Fee Limit", "fallback_text": "10 TRX" },
    "amount_v2": { "amount": "10", "abbreviation": "TRX" }
  }
}
\`\`\`

## Why this is worth tracking

The atomic (SUN) integer no longer appears anywhere in the payload. For consumers that:
- displayed both for verifiability,
- programmatically parsed the SUN integer out of the text,
- or used the SUN value for exact (lossless) amount math,

… the SUN value must now be reconstructed by multiplying the TRX string by 10^6, which is awkward for callers using fixed-point arithmetic and impossible for callers without 64-bit support.

## Possible resolutions

1. **Status quo**: rely on consumers to compute SUN from TRX. Document the change.
2. **Add SUN to `fallback_text`**: e.g., `fallback_text: \"10 TRX (10000000 SUN)\"`, keep AmountV2 numeric value as TRX. Display-only, no schema change.
3. **Emit a sibling `Number` field** carrying the SUN integer for amounts where atomic precision matters (Fee Limit, Amount, Frozen Balance, Balance, Unfreeze Balance). Schema-additive.

Recommend #2 — minimal-impact, keeps both representations human-visible.

## Source

Surfaced by `/code-review xhigh` finding #11 on the modernization PR.

Issue #311, opened 2026-05-22
tron: From/To/Owner fields type-changed from TextV2 to AddressV2 (consumer breakage)


prasanna-anchorage:
## Context

PR https://github.com/anchorageoss/visualsign-parser/pull/314 (Wire Tron into parser_cli + decode v2 staking + modernize Tron crate) migrated `visualsign-tron` from hand-built `SignablePayloadField::TextV2` structs to the workspace-standard `visualsign::field_builders` helpers (`create_address_field`, `create_amount_field`, `create_text_field`).

Side effect: the **shape** of certain fields in `SignablePayload` changed for any Tron transaction:

| Field | Old variant | New variant |
|---|---|---|
| `TransferContract.From` / `To` | `TextV2 { text: "T..." }` | `AddressV2 { address: "T...", asset_label: "", ... }` |
| `TransferContract.Amount` | `TextV2 { text: "7000000 SUN (7 TRX)" }` | `AmountV2 { amount: "7", abbreviation: "TRX" }` |
| All new v2-staking `Owner` / `Receiver` | n/a (newly-decoded) | `AddressV2` |
| All new v2-staking amounts (`Frozen Balance`, `Balance`, etc.) | n/a | `AmountV2` |

## Why this is worth tracking

Any downstream consumer that pattern-matches on `SignablePayloadField::TextV2` (e.g., a hardware-wallet renderer with a `match field { TextV2 { .. } => render_text(), _ => render_unknown() }` style switch) will hit its fallback/unknown arm for Tron `From` / `To` / `Owner` / `Receiver` / amount fields where it previously rendered them as text.

This is the **intended modernization** — `AddressV2` and `AmountV2` are the right variants for these data — but the rollout has a coordination cost: wallet UIs need to add explicit `AddressV2` / `AmountV2` arms before they ship a build that talks to this parser.

## Action

- Communicate the shape change to wallet integrators before the next release that ships this code.
- Confirm internal consumers (hardware-wallet renderers, mobile/desktop apps) already handle `AddressV2` and `AmountV2` for non-Tron chains and will accept them for Tron too.
- If a consumer is found that *cannot* migrate, either add a back-compat flag or accept the breakage with explicit sign-off.

## Files touched in the original PR

- `src/chain_parsers/visualsign-tron/src/lib.rs` (entire decoder rewritten)

## Source

Surfaced by `/code-review xhigh` finding #10 on the modernization PR (see code-review JSON in PR conversation).

[...]

PR #314, opened 2026-05-22
feat(tron): wire parser_cli + decode v2 staking + modernize

prasanna-anchorage:
## Summary

- Wires Tron into `parser_cli` (new default-on `tron` feature, `TronPlugin` mirroring `solana.rs`) so `--chain tron -t <hex>` works end-to-end. Adds a `tron-text` CLI fixture pair and the matching skip-list entry.
- Adds full decoding for the v2-staking contract family in `visualsign-tron`: `FreezeBalanceV2Contract`, `UnfreezeBalanceV2Contract`, `WithdrawExpireUnfreezeContract`, `DelegateResourceContract`, `UnDelegateResourceContract`. Previously every non-`TransferContract` rendered as `"(not fully decoded)"`.
- Handles both Tron on-the-wire forms (bare `transaction::Raw` *and* wrapped `Transaction { raw_data, ... }`) with a contracts-first discriminator; pins the dispatch with regression tests for both directions, including the `[0x08,0x00]` field-tag-collision ambiguity case.
- Modernizes the crate to workspace conventions: adopts `visualsign::field_builders` (`AmountV2`/`AddressV2`/`TextV2` helpers), swaps the hand-rolled `sha2 + base58=0.2` base58check for `bs58 = 0.5.1` with the `check` feature, replaces the deny-listed `Utc.timestamp_millis_opt(..).unwrap()` with `DateTime::from_timestamp_millis`, and replaces `if let Ok(..)` swallowing of inner parse errors with `?`.
- Hardens the new code: `address_to_base58` validates 21-byte + `0x41` prefix and emits an `<invalid Tron address: hex>` marker for malformed input; `resource_label` surfaces unknown `ResourceCode` ints as `UNKNOWN(n)`; missing `Contract.parameter` shows a `<missing parameter>` marker; `DelegateResource.Lock Period` is gated on `lock` so a locked delegation with `period=0` still shows the field.
- 21 new unit tests (5 real on-chain fixtures pulled from `trongrid /wallet/getblockbynum`, one per v2 staking type, plus synthetics and edge cases) and a new CLI integration fixture.

## CLI invocation

```bash
cargo run --bin parser_cli -- --chain tron --output human \
  -t 0a730a02049d22080f1beff095be0cfd40a097a684e5335a55083612510a34747970652e676f6f676c65617069732e636f6d2f70726f746f636f6c2e467265657a6542616c616e63655632436f6e747261637412190a15416a6ca578c7937e1bf6aea4be657f9d22716c424d100570a0df8cdbe433
```

```
┌─ Transaction: Tron Transaction
│  Version: 0
│  Type: TronTx
│
└─ Fields:
   ├─ Network: Tron
   ├─ Timestamp: 2026-05-21 16:34:12 UTC (1779381252000 ms)
   ├─ Expiration: 2026-05-22 16:34:12 UTC (1779467652000 ms)
   ├─ Fee Limit: 0 TRX
   ├─ Ref Block: 049d
   ├─ Ref Block Hash: 0f1beff095be0cfd
   ├─ Contract Type: FreezeBalanceV2 (Stake)
   ├─ Owner: TKfvnGvr7YMTBYz8xfx1XfgkWk2JZZgj4s
   ├─ Frozen Balance: 0.000005 TRX
   └─ Resource: BANDWIDTH
```

## Follow-ups (intentional behavior changes; tracked separately under the `tron` label)

- #311 — `From`/`To`/`Owner` field type-changed from `TextV2` to `AddressV2` (consumer migration)
- #312 — SUN-denominated amounts no longer appear in payload text (could be reintroduced via `fallback_text`)
- #313 — per-contract decode errors are now fatal where they used to be silently dropped (arguably the better behavior; documented as a contract change)

## Test plan

- [ ] `make -C src lint` — clippy `-D warnings`, full workspace
- [ ] `make -C src fmt`
- [ ] `make -C src test` — full workspace
- [ ] `cargo test -p visualsign-tron` — 21 tests including the user's FreezeBalanceV2 fixture, 5 real-on-chain fixtures (one per v2 contract type), the `parse_tron_bytes` ambiguity reproducer, the malformed-address rejection, the `<missing parameter>` marker path, and the `delegate.lock=true,period=0` regression
- [ ] `cargo test -p parser_cli` — the new `tron-text` fixture exercises CLI dispatch through `TronPlugin::register`
- [ ] End-to-end: the CLI command above produces the output above (or equivalent real-on-chain hex from any of the five v2 staking contract types)


src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -37,10 +40,44 @@ fn decode_transaction(
...
+//   1. Bare `transaction::Raw` bytes (the "raw_data_hex" returned by trongrid).
+//   2. The wrapped `Transaction { raw_data, signature, ret }` (what gets broadcast).
+//
+// The two share field-1 tag (`Raw.ref_block_bytes` and `Transaction.raw_data` are both
+// wire-type-2 at tag 1), so a bare Raw whose `ref_block_bytes` payload happens to be a
+// parseable sub-message will *also* parse as a wrapped Transaction with `raw_data =
+// Some(near-empty Raw)`. Picking the wrapped form unconditionally hides the real bare-Raw
+// content. Try both and pick the form that decoded a real transaction (at least one
+// `contract`); if neither has contracts, prefer the wrapped form when it had `raw_data` at
+// all, else the bare-Raw result.
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in c20d7a6 — comment now matches the implementation: when both parses succeed with zero contracts, we tie-break to bare rather than wrapped. Picking bare is the safer fallback because a wrapped re-parse of bare bytes typically yields an empty inner Raw with the original `ref_block_bytes` lost.

src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -266,27 +413,592 @@ pub fn transaction_string_to_visual_sign(
...
+    }
+}
 
-    base58::ToBase58::to_base58(&with_checksum[..])
+// f64 mantissa is 53 bits (~9.0e15). Tron's max-supply ceiling of ~10^17 SUN is above 2^53,
+// so amounts above ~9 quadrillion SUN (~9 billion TRX) in a single field will round in the
+// last digits — this only affects the displayed string, not the signed bytes. Realistic
+// per-tx amounts stay well below 2^53 SUN.
+fn sun_to_trx_string(sun: i64) -> String {
+    format!("{}", sun as f64 / 1_000_000.0)
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in c20d7a6 — `sun_to_trx_string` now uses i64 integer math (quotient + 6-digit fractional with trailing-zero trim) instead of `f64` division, so the displayed value is byte-exact at any SUN magnitude. New regression test `sun_to_trx_string_is_byte_exact_at_any_magnitude` pins 9_999_999_999_999_999 SUN to `"9999999999.999999"` and `i64::MAX` SUN to `"9223372036854.775807"` — both would have rounded under the old f64 path.

src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -266,27 +412,635 @@ pub fn transaction_string_to_visual_sign(
...
+// fake base58 string (e.g. an empty input would otherwise render as the 6-char checksum).
+const TRON_ADDRESS_LEN: usize = 21;
+const TRON_MAINNET_PREFIX: u8 = 0x41;
 
-    let mut hasher = Sha256::new();
-    hasher.update(hash1);
-    let hash2 = hasher.finalize();
+fn address_to_base58(address_bytes: &[u8]) -> String {
+    if address_bytes.len() != TRON_ADDRESS_LEN || address_bytes[0] != TRON_MAINNET_PREFIX {
+        return format!("<invalid Tron address: {}>", hex::encode(address_bytes));
```
pepe-anchor, line 424: The `hex::encode(address_bytes)` call in the invalid-address error path is unbounded. A malformed protobuf contract could supply an arbitrarily large `owner_address` or `receiver_address` byte vec, producing a proportionally large hex string in the display output. Input is bounded at the CLI layer (`MAX_TRANSACTION_INPUT_SIZE`), but that's still up to ~10 MB worth of hex characters in a display field. Truncate to a safe prefix, e.g. `hex::encode(&address_bytes[..address_bytes.len().min(32)])` + `...` suffix, to keep the display bounded regardless of upstream limits.

---

src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -241,11 +190,208 @@ fn convert_to_visual_sign_payload(
...
+                &resource_label(undelegate.resource),
+            )?);
+            fields.push(create_amount_field(
+                "Balance",
+                &sun_to_trx_string(undelegate.balance),
+                "TRX",
+            )?);
+        }
+        other => {
+            fields.push(create_text_field(
```
pepe-anchor, line 386: The `other` arm in `decode_contract` embeds the raw `type_url` string from protobuf directly into a display field without length bounding: `format!("{other} (not fully decoded)")`. A crafted transaction with a long `type_url` field (valid protobuf, bounded only by the outer input size limit) produces a proportionally large display string. Truncate `other` to a reasonable length (e.g. 128 chars) before embedding it, consistent with the project's philosophy of bounded display strings from untrusted data.

---

src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -266,27 +412,635 @@ pub fn transaction_string_to_visual_sign(
...
+            .unwrap();
+        assert_eq!(text_value(find_field(&payload, "Lock").unwrap()), "true");
+        assert_eq!(
+            text_value(find_field(&payload, "Lock Period").unwrap()),
+            "0"
+        );
+    }
+
+    #[test]
+    fn sun_to_trx_string_is_byte_exact_at_any_magnitude() {
```
pepe-anchor, line 985: `sun_to_trx_string` is tested with `i64::MAX` but not `i64::MIN`. The implementation uses `i64::unsigned_abs()` which correctly returns `9223372036854775808u64` for `i64::MIN` without panicking (unlike `i64::abs()`), so there is no actual bug. Adding `assert_eq!(sun_to_trx_string(i64::MIN), "-9223372036854.775808")` to the existing test would guard against a future refactor that naively swaps in `abs()`.

---

src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -37,10 +40,43 @@ fn decode_transaction(
...
+// the original ref_block_bytes lost.
+fn parse_tron_bytes(bytes: &[u8]) -> Result<transaction::Raw, TronParserError> {
+    let bare_result = transaction::Raw::parse_from_bytes(bytes);
+    let wrapped_raw = TronTransaction::parse_from_bytes(bytes)
+        .ok()
+        .and_then(|w| w.raw_data.into_option());
+
+    match (bare_result, wrapped_raw) {
+        (Ok(bare), Some(wrapped)) => {
+            if !bare.contract.is_empty() {
```
pepe-anchor, line 66: When both the bare parse and the wrapped re-parse succeed and both contain at least one contract, the code picks `bare`. The comment block above the function explains the *tie-break-when-neither-has-contracts* case but does not justify the both-non-empty case specifically. There is also no fixture covering this branch: a genuine wrapped `Transaction` broadcast whose bare re-interpretation also incidentally produces a non-empty `contract` slot would silently get the bare view. Worth either tightening the comment or adding a fixture that exercises the both-non-empty branch.

---

src/chain_parsers/visualsign-tron/src/lib.rs
```diff
@@ -95,140 +131,53 @@ fn convert_to_visual_sign_payload(
...
-                            label: "Contract Type".to_string(),
-                        },
-                        text_v2: SignablePayloadFieldTextV2 {
-                            text: format!("{} (not fully decoded)", parameter.type_url),
-                        },
-                    });
-                }
+        match contract.parameter.as_ref() {
+            Some(parameter) => {
+                decode_contract(&parameter.type_url, &parameter.value, &mut fields)?;
```
pepe-anchor, line 175: Using `?` to propagate `decode_contract` errors means a known contract type whose value bytes are corrupted (e.g. truncated on the wire) aborts the entire conversion and returns an error to the signer, rather than showing a degraded but safe view. The `None`-parameter arm already demonstrates the right pattern: surface a visible marker instead of failing hard. Consider catching the parse error inside `decode_contract` and pushing a `create_text_field("Contract Type", "<decode failed: ...>")` entry, so the signer still sees the other header fields and can make an informed decision.

---

pepe-anchor (commented):
Reviewed with [Claude](https://claude.com/claude-code) on behalf of @pepe-anchor. 5 suggestion(s) of 6 drafted; 0 marked as blocking. _Treat findings as suggestions, not directives._

Read-only peer review (light tier: copilot-review + security-review + requesting-code-review + gemini). 4 MEDIUM correctness/security observations, no blocking issues. Two flag unbounded display strings in error paths (lib.rs:386 type_url echo, lib.rs:424 invalid-address hex), one flags hard error propagation in decode_contract (lib.rs:175) that could be a degraded view instead, and one notes parse_tron_bytes silently picks bare when both bare and wrapped have non-empty contracts (lib.rs:66). Two LOW nits (lib.rs:578, lib.rs:985) pre-skipped.

pepe-anchor (approved):

PR #316, opened 2026-05-25
refactor(cli): extract ChainPlugin into parser_cli_core and per-chain crates

shahan-khatchadourian-anchorage:
## Summary

- Adds new `parser/cli-core` library crate that owns `ChainPlugin`, `SharedArgs`, `run`, output formatting, and chain-agnostic helpers (`mapping_parser`, `tx_input`, `chains`, `test_utils`).
- Moves `EthereumPlugin` and `SolanaPlugin` impls into their chain crates (`chain_parsers/visualsign-ethereum/src/cli_plugin.rs`, `chain_parsers/visualsign-solana/src/cli_plugin.rs`) behind a default-on `cli-plugin` feature.
- Slims `parser/cli` to a thin binary that composes `SharedArgs` + per-chain clap args and hands a `Vec<Box<dyn ChainPlugin>>` to `cli_core::run`.
- Enables external workspaces to add a chain by depending on a new chain crate, without editing any upstream binary code.

## Notable adaptations

- `mapping_parser::load_mappings` now returns `BTreeMap<String, V>` instead of `HashMap<String, V>`. Motivation: the chain crates' per-crate `clippy.toml` lists `HashMap`/`HashSet` under `disallowed_types`, and the moved `cli_plugin.rs` call sites in `visualsign-{ethereum,solana}` would otherwise trip the lint. Call sites convert back to `HashMap` when populating protobuf metadata types. Side benefit: deterministic iteration order.
- The chain crates gain a `cli-plugin` cargo feature (default on). Pure-library consumers can opt out with `--no-default-features` and avoid pulling in clap, parser_cli_core, and generated.
- `test_utils` is `pub mod` under a new `test-utils` feature on `parser_cli_core` so chain crates can use it as a dev-dependency.

## Test plan

- [x] `cargo build --workspace` succeeds.
- [x] CLI run with the Ethereum 1559 fixture produces byte-identical output to pre-refactor (SHA256 confirmed at branch point and HEAD).
- [x] `cargo check -p visualsign-ethereum --no-default-features` builds (no CLI deps leak into lean lib mode).
- [x] `cargo check -p visualsign-solana --no-default-features` builds.
- [ ] One pre-existing test fails identically on `origin/main` and on this branch: `integration::parser_solana_native_transfer_e2e` asserts no diagnostic fields are emitted but parser_app emits them. This failure was present before this PR's work and is not introduced by these changes. Captured for transparency.


prasanna-anchorage (commented):
Reviewed the full diff. The architecture is clean and well-motivated — extracting `ChainPlugin`/`SharedArgs`/`run`/output into a reusable `parser_cli_core` crate and moving the per-chain plugins into the chain crates behind a default-on `cli-plugin` feature is exactly the right shape for letting external workspaces compose their own binary without editing upstream. Moving the `process::exit` out of the library into `main.rs` (so `run` just returns `Result`) is a nice cleanup, and the `write_temp_json` path-traversal hardening is a good touch.

A few things before this can come out of draft and merge:

### 1. (Blocking) The rebase must preserve #335's ABI-signing security fix
This branch is ~47 commits behind `main` and predates #335 (merged since), which closed the "unsigned ABI accepted" hole. On current `main`:
- `mapping_parser::load_mappings` now takes `build_value: impl Fn(&MappingComponents, String) -> Result<V, String>` (not `-> V`) and matches on the `Result` so a signing failure skips the entry.
- `parser/cli/src/ethereum.rs` (the file this PR renames to `visualsign-ethereum/src/cli_plugin.rs`) imports `sign_abi` / `CLI_DEV_SIGNING_KEY_SEED` and signs both file-loaded and synthesized-proxy ABIs.

This PR moves both files with the **pre-#335 signatures** (`-> V`, `signature: None`). Because it's a file *rename* plus the moved content, a naive conflict resolution will silently drop the signing logic and revert the security fix. When rebasing, please carry forward:
- the `Result`-returning `build_value` signature + the match arms in the moved `mapping_parser.rs`, and
- the `sign_abi(...)` calls + `Ok(Abi { ..., signature: Some(...) })` in the moved `cli_plugin.rs` (both the file-load closure and the proxy-synthesis path).

Worth double-checking the Solana side too (`solana.rs` -> `cli_plugin.rs`) against any IDL-signature work that landed on `main`. A post-rebase `cargo test -p visualsign-ethereum` plus the CLI integration tests should confirm the signing tests still pass.

### 2. Non-ASCII characters in CLI help text
`SharedArgs::network`:
```
/// Network identifier — chain ID (1, 137, …) or canonical name (`ETHEREUM_MAINNET`, …).
```
This uses an em-dash (`—`) and ellipsis (`…`). clap's derive surfaces doc comments as `--help` text, and the project requires ASCII-only for terminal compatibility (CLAUDE.md: `>=` not `≥`, `->` not `→`). Please use `-` and `...`.

### 3. Minor: `--help` lost the network examples
The old `network` arg help enumerated chain IDs and canonical names across multiple lines; the new single-line doc drops them. Not blocking, but the extra guidance was useful — consider keeping a condensed version.

### 4. Test-plan item to resolve before un-drafting
The disclosed pre-existing failure (`integration::parser_solana_native_transfer_e2e`) — agreed it's not introduced here, and the honesty is appreciated. Just flagging it should be green (or have a tracking issue) before this leaves draft so the merge train isn't red.

Design ✅ — happy to approve once the rebase preserves #335 (verified by tests) and the ASCII fix is in.

shahan-khatchadourian-anchorage:
Rebased onto current `main` and addressed the review. New tip: `acdd59e`.

**1 (blocking) — #335 ABI-signing fix preserved.** Rebased onto `main` (which has #335), resolving the file-rename conflicts so the signing logic carries forward rather than reverting:
- `mapping_parser::load_mappings` keeps the `Result`-returning `build_value` and the per-entry skip-with-warning arms (combined with this branch's `BTreeMap` return — internal `BTreeMap`/`BTreeSet`, converted to the proto `HashMap` only at the `.into_iter().collect()` boundary).
- `visualsign-ethereum/src/cli_plugin.rs` signs both file-loaded ABIs and the synthesized-proxy path with `sign_abi(.., &CLI_DEV_SIGNING_KEY_SEED)` and emits `signature: Some(..)`.
- Verified: `cargo test -p visualsign-ethereum` is 249/249 incl. the signed-proxy extraction regression test; CLI integration tests pass.
- Solana side checked against #336: IDL signatures remain optional (`signature: None`) — the trusted-program override rejection lives in the parser's `from_idl_mappings`, not the CLI — so the rebased Solana plugin matches `main`'s behavior (only the return type changed to `BTreeMap`).

**2 — Non-ASCII in `--network` help.** Replaced the em-dash and ellipsis with ASCII.

**3 — Restored the network examples.** Chain-ID and canonical-name examples are back, moved into clap `long_help` (a string attribute, so the `ALL_CAPS` names don't trip `clippy::doc_markdown`).

**4 — Pre-existing failing test.** `integration::parser_solana_native_transfer_e2e` now passes on this branch — the rebase pulled in the `main` fix, so it's green, not just disclosed.

Two incidental items from the rebase:
- **Tron**: a Tron CLI plugin landed on `main` after this branch was cut, written against the old `parser_cli::ChainPlugin`. Migrated it into `visualsign-tron` behind a default-on `cli-plugin` feature (consistent with eth/solana), so the binary stays thin and Tron support isn't dropped.
- Removed a stray `.env` that had been committed locally; stripped from all commits (never pushed).

Full gate green: `cargo build`, `clippy --all-targets -D warnings`, `--no-default-features` lib checks for eth/solana/tron, and `make test`.

shahan-khatchadourian-anchorage:
ready for re-review @prasanna-anchorage Thanks

prasanna-anchorage (approved):
Re-reviewed after the update — all four points from my earlier review are resolved, and I verified each by building and testing the branch locally (workspace build, both lean-lib `--no-default-features` modes, clippy `-D warnings`, fmt — all clean).

**1. (was my blocking concern) #335 ABI-signing preserved through the rebase.** Verified concretely, not just from the diff:
- `visualsign-ethereum/src/cli_plugin.rs` retains both `sign_abi` call sites — the file-load closure (`Ok(Abi { signature: Some(..) })`) and the synthesized-proxy path (`signature: Some(..)`).
- `mapping_parser::load_mappings` keeps the `-> Result<V, String>` closure signature.
- `test_try_extract_unsigned_abi_rejected` passes (249 ethereum tests green). The unsigned-ABI rejection is intact.

**2 & 3. ASCII help + examples.** `cli-core/src/lib.rs` is ASCII-clean; `--network` now uses `-`/`...` with the chain-ID and canonical-name examples restored via `long_help`.

**4. The pre-existing integration failure.** `parser_solana_native_transfer_e2e` still fails (diagnostics fields emitted vs. the test asserting none), but I confirmed this PR touches neither `integration/` nor the diagnostics path, so it can't be the cause — it's the pre-existing failure you disclosed. Worth a tracking issue to fix the stale assertion, but not a blocker for this refactor.

**Also good:** lean-lib `--no-default-features` builds for both chains confirm the `cli-plugin` gate doesn't leak clap/cli-core/generated into library consumers; the `clippy.toml` HashMap/HashSet ban and the `write_temp_json` path-containment hardening are nice. The tron cli_plugin extraction shows the pattern generalizes cleanly to a third chain.

**Heads-up (not blocking):** this overlaps with #355, which gates `sign_abi`/`CLI_DEV_SIGNING_KEY_SEED` behind a `dev-signing` feature and changes `sign_abi`'s signature. Since #316 moves the `sign_abi` caller into `visualsign-ethereum`'s `cli_plugin` (behind the `cli-plugin` feature), whichever lands second will need to wire `cli-plugin` to also enable `dev-signing` (and pick up the new signature). Just flagging so the merge order is coordinated.

Design and execution LGTM.

PR #317, opened 2026-05-25
fix(tron): handle out-of-range expiration timestamps without panicking

pepe-anchor:
## Summary

`format_timestamp` in the Tron parser called `Utc.timestamp_millis_opt(...).unwrap()` on an attacker-controlled `i64` taken straight from the protobuf-decoded raw transaction. Values like `i64::MAX` or `i64::MIN` fall outside chrono's representable range and return `LocalResult::None`, so the `.unwrap()` panicked the parser and gave any unauthenticated caller a trivial DoS against the visual signing service.

This PR replaces the `.unwrap()` with an explicit `match` on `LocalResult` and returns `"invalid timestamp"` for out-of-range / ambiguous values. The call site still surfaces the raw millisecond integer alongside the formatted string, so signers retain the underlying value when chrono can't render it.

## What changed

- `src/chain_parsers/visualsign-tron/src/lib.rs`:
  - `chrono::{LocalResult, TimeZone, Utc}` is now imported at module top instead of inside `format_timestamp`.
  - `format_timestamp` matches on `LocalResult::{Single, None, Ambiguous}`. `Single` formats as before; `None` and `Ambiguous` return `"invalid timestamp"`.
  - Added a `#[cfg(test)]` module with four regression tests: known timestamp, unix epoch, `i64::MAX`, `i64::MIN`. The last two are the PRS-228 regression and previously panicked.

## Linear ticket

Closes PRS-228

Linear: https://linear.app/anchorlabs/issue/PRS-228

## Rollback

Revert this commit. The change is a single file, no schema / config migration, no new dependency (chrono was already a Tron crate dep).

## Backwards compatibility

Yes. For all in-range timestamps the output is byte-identical to before. For previously-panicking inputs the output is now the sentinel `"invalid timestamp"` string instead of a process crash, which is strictly better behaviour. No public API or field schema changed.

## Test plan

Verification is scoped to the affected crate (per the recovery brief):

- [x] `cargo fmt`
- [x] `cargo clippy -p visualsign-tron -- -D warnings` (clean)
- [x] `cargo test -p visualsign-tron` (4/4 tests pass, including the two new `i64::MAX` / `i64::MIN` regression tests)

CI will exercise the rest of the workspace.

### DIFF
--- a/src/chain_parsers/visualsign-tron/src/lib.rs
+++ b/src/chain_parsers/visualsign-tron/src/lib.rs
@@ -10,6 +10,7 @@ use visualsign::{
 use anychain_tron::protocol::Tron::transaction;
 use anychain_tron::protocol::balance_contract::TransferContract;
 use base64::{Engine as _, engine::general_purpose::STANDARD as b64};
+use chrono::{LocalResult, TimeZone, Utc};
 use protobuf::Message;
 use sha2::{Digest, Sha256};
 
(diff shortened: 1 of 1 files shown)


pepe-anchor:
Thanks for the review. The summary matches the intent of the change: replace the unconditional unwrap with a checked LocalResult match, fall back to "invalid timestamp" for out-of-range millis, and lock the behavior in with i64::MAX / i64::MIN regression tests. No actionable items to address from the review body. If you disagree, react with 👎 and we'll revisit.

prasanna-anchorage:
Superseded by #314 (merged). The PRS-228 fix is already in `main`:

- [`format_timestamp` at `src/chain_parsers/visualsign-tron/src/lib.rs:464-468`](https://github.com/anchorageoss/visualsign-parser/blob/main/src/chain_parsers/visualsign-tron/src/lib.rs#L464-L468) uses `chrono::DateTime::from_timestamp_millis(...).map(...).unwrap_or_else(|| "invalid timestamp")` — semantically equivalent to this PR's `match LocalResult` approach (no panic on out-of-range).
- Regression test `format_timestamp_handles_invalid_value` at `lib.rs:899-903` asserts `format_timestamp(i64::MAX) == "invalid timestamp"`.

Attempted a rebase locally — every hunk in this PR conflicts with #314's rewrite of the same function and test module, with no remaining security delta. The only net-new content here vs. main is extra boundary coverage (`i64::MIN`, `0`, a known timestamp). If those are wanted, happy to take a small follow-up PR just for the test cases.

Closing as superseded.

PR #318, opened 2026-05-25
fix(sui): remove reachable todo!() in momentum flash_loan handlers

pepe-anchor:
## Summary

The momentum preset in `visualsign-sui` advertised support for `flash_loan` and `repay_flash_loan` in its config but the corresponding match arms in `MomentumVisualizer::visualize_tx_commands` were stubs that called `todo!("Have not found tx for testing yet")`. A transaction touching either Move function on the momentum mainnet package would panic the parser thread, a reachable DoS on any server handling untrusted transaction bytes (gRPC parser, REST gateway, local dev container).

Closes PRS-232 — https://linear.app/anchorlabs/issue/PRS-232

## What changed

- `presets/momentum/config.rs` — drop the `flash_loan` and `repay_flash_loan` entries from the `chain_config!` block so `TradeFunctions` no longer enumerates them and `MomentumVisualizer::can_handle` returns false for those Move functions.
- `presets/momentum/mod.rs` — remove the two `todo!()` match arms (variants no longer exist) and add two regression tests (`test_momentum_flash_loan_does_not_panic`, `test_momentum_repay_flash_loan_does_not_panic`) that construct a synthetic `SuiProgrammableMoveCall` against the momentum mainnet package id for each function name, then assert `can_handle == false` and `visualize_tx_commands` returns `Err` rather than panicking.

## Why this shape

Real visualization for `flash_loan` / `repay_flash_loan` would need test vectors we don't have and a settled Move ABI for the momentum DEX flash-loan path. Removing the functions from the supported set is the correct DoS-safe behavior until a fixture lands — the dispatcher falls through to the generic / error path instead of panicking.

## Rollback

Revert this commit. The two Move functions will once again be enumerated as supported and will panic on first call (back to the pre-fix state), no schema or data changes to undo.

## Backwards compatibility

No schema changes. The only behavioral change is that `flash_loan` and `repay_flash_loan` calls under the momentum mainnet package now return a parser error instead of panicking. No deployed VisualSign payloads change.

## Test plan

- [x] `cargo fmt` clean.
- [x] `cargo clippy -p visualsign-sui -- -D warnings` clean.
- [x] `cargo test -p visualsign-sui` — 22 tests pass, including the two new regression tests asserting no-panic behavior for both functions.

### DIFF
--- a/src/chain_parsers/visualsign-sui/src/presets/momentum/config.rs
+++ b/src/chain_parsers/visualsign-sui/src/presets/momentum/config.rs
@@ -31,11 +31,6 @@ crate::chain_config! {
                     sqrt_price_limit as SqrtPriceLimit: u128 => 4 => get_sqrt_price_limit,
                 ),
                 repay_flash_swap as RepayFlashSwap => RepayFlashSwapIndexes(),
-                flash_loan as FlashLoan => FlashLoanIndexes(
-                    amount_x as AmountX: u64 => 1 => get_amount_x,
-                    amount_y as AmountY: u64 => 2 => get_amount_y,
-                ),
-                repay_flash_loan as RepayFlashLoan => RepayFlashLoanIndexes(),
                 swap_receipt_debts as SwapReceiptDebts => SwapReceiptDebtsIndexes(),
             },
         }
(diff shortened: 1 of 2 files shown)


pepe-anchor:
Thanks for the review summary. No actionable items flagged (0 inline comments, summary-only body). The change intentionally drops support for these handlers rather than papering over the panic, so we'll keep the scope as-is.

If you disagree, react with 👎 and we'll revisit.

prasanna-anchorage (approved):

PR #330, opened 2026-05-25
fix(ethereum): read V3 swap output token from end of path

pepe-anchor:
## Summary

Universal Router V3 swap decoders were reading the output token from `path[23..43]`, which on multi-hop V3 paths is the first intermediate token, not the destination. Users signing a 2-hop swap saw the middle token displayed as "Output Token", masking the actual asset they were buying.

## What changed

- New `decode_v3_path` helper in `universal_router.rs` parses the V3 path correctly:
  - Encoding is `tokenIn(20) | fee(3) | nextToken(20) | fee(3) | ... | tokenOut(20)`, so a valid N-hop path is `20 + N*23` bytes.
  - Output token is read from the final 20 bytes, never from a fixed offset.
  - Returns `(token_in, token_out, first_hop_fee, hop_count)`.
- Both `decode_v3_swap_exact_in` and `decode_v3_swap_exact_out` use the helper.
- Multi-hop swaps now disclose the hop count and clarify that the surfaced fee is the first hop's (`"... via V3 (2 hops, first fee 0.05%)"`).
- Path-length validation tightened: previously a 44-byte blob passed the legacy `>= 43` check and silently synthesized addresses from junk bytes; it now returns an "Invalid path" error.

## Rollback

Single-file change. Revert this commit on the affected branch; no schema, registry, or proto changes.

## Test plan

- [x] Unit: `decode_v3_path` single-hop, two-hop (regression: output must be final token, not intermediate), three-hop, malformed length rejections.
- [x] End-to-end: `decode_v3_swap_exact_in` and `decode_v3_swap_exact_out` on a 2-hop encoded payload assert that the `Output Token` field and the subtitle reference the final token (not the intermediate) and that the subtitle discloses `"2 hops"`.
- [x] Negative: malformed 44-byte path surfaces an `Invalid path` error instead of garbage addresses.
- [x] `cargo fmt`, `cargo clippy -p visualsign-ethereum --all-targets -- -D warnings`, `cargo test -p visualsign-ethereum` all green locally.

### DIFF
--- a/src/chain_parsers/visualsign-ethereum/src/protocols/uniswap/contracts/universal_router.rs
+++ b/src/chain_parsers/visualsign-ethereum/src/protocols/uniswap/contracts/universal_router.rs
@@ -394,14 +394,22 @@ impl UniversalRouterVisualizer {
         })
     }
 
-    /// Decodes a Uniswap V3 swap path.
+    /// Decodes a Uniswap V3 swap path into raw byte positions.
     ///
-    /// Paths are encoded as: tokenIn(20) | fee(3) | nextToken(20) | fee(3) | ... | tokenOut(20).
+    /// Paths are encoded as: token(20) | fee(3) | token(20) | fee(3) | ... | token(20).
     /// A valid N-hop path therefore has length 20 + N * 23.
     ///
-    /// Returns (token_in, token_out, first_hop_fee, hop_count) or None for malformed paths.
-    /// The output token is always read from the FINAL 20 bytes of the path so that multi-hop
-    /// swaps display the actual destination token, not the first intermediate hop.
+    /// Returns `(first_token, last_token, first_hop_fee, hop_count)` or `None` for malformed
+    /// paths, where `first_token` is the leading 20 bytes and `last_token` is the trailing
+    /// 20 bytes. The helper is intentionally semantics-free: role assignment depends on the
+    /// caller's swap direction.
+    ///
+    /// - ExactIn paths are encoded `tokenIn || fee || ... || tokenOut`, so callers alias
+    ///   `first_token = tokenIn`, `last_token = tokenOut`.
+    /// - ExactOut paths are encoded in reverse (`tokenOut || fee || ... || tokenIn`) per
+    ///   Uniswap V3 convention (see `SwapRouter.exactOutputInternal` /
+    ///   `Path.decodeFirstPool`), so callers must alias
+    ///   `first_token = tokenOut`, `last_token = tokenIn`.
     fn decode_v3_path(path: &[u8]) -> Option<(Address, Address, u32, usize)> {
         // Single hop minimum: 20 + 23 = 43 bytes. Each additional hop adds 23 bytes.
         if path.len() < 43 || (path.len() - 20) % 23 != 0 {
@@ -791,9 +799,11 @@ impl UniversalRouterVisualizer {
 
         let (_recipient, amount_out, amount_in_max, path, _payer_is_user) = params;
 
-        // Decode V3 path. For multi-hop paths, the output token MUST be read from the
-        // FINAL 20 bytes, not from offset 23..43 (which is the first intermediate token).
-        let (token_in, token_out, fee, hops) = match Self::decode_v3_path(&path) {
+        // Decode V3 path. ExactOut paths are encoded in reverse vs ExactIn per Uniswap V3
+        // convention (see `SwapRouter.exactOutputInternal` / `Path.decodeFirstPool`): the
+        // on-chain layout is `tokenOut || fee || ... || tokenIn`. The helper returns raw
+        // byte positions, so we alias `first_token = tokenOut`, `last_token = tokenIn`.
+        let (token_out, token_in, fee, hops) = match Self::decode_v3_path(&path) {
             Some(decoded) => decoded,
             None => {
                 return SignablePayloadField::TextV2 {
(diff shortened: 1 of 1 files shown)


prasanna-anchorage:
Heads up: this repo is **public**, so internal Linear refs (`PRS-xxx`) shouldn't leak into PR-visible surface — external readers can't resolve them.

This PR's leaks:
- **Title contains `(PRS-229)`** — please rename, e.g. `fix(ethereum): read V3 swap output token from end of path`.
- Body links `linear.app/anchorlabs/issue/PRS-229` — swap for a one-line bug-class summary or a public GH issue reference.
- 5 code references / 5 inline comments mention `PRS-229` — rename to describe the bug class.
- Branch `pepefigueira/prs-229-...` also exposes the ID (optional).

Suggested convention: describe the bug in the name (e.g., `test_v3_multihop_output_token_from_path_end`). For persistent traceability, file a public GitHub issue.

pepe-anchor:
Scrubbed PRS-229 refs from title, body, and code/comments. Branch left as-is since it'll disappear on merge.

prasanna-anchorage:
Deep-review finding that should block merge as-is: the same bug class this PR closes for `V3_SWAP_EXACT_IN` is still present for `V3_SWAP_EXACT_OUT`.

## The bug

V3 ExactOut encodes its path **in reverse** of ExactIn — that's the documented Uniswap V3 convention (`SwapRouter.exactOutputInternal` destructures `Path.decodeFirstPool` as `(tokenOut, tokenIn, fee)`). So in an ExactOut path the first 20 bytes are `tokenOut` (what the user receives), the trailing 20 bytes are `tokenIn` (what they pay).

The new `decode_v3_path` helper returns `(first_20, last_20, …)` named `(token_in, token_out, …)`. `decode_v3_swap_exact_in` consumes it correctly. **`decode_v3_swap_exact_out` consumes the same return values without swapping the roles**, so for any V3 ExactOut command:

- `tokenOut` is displayed as `"Input Token"`
- `tokenIn` is displayed as `"Output Token"`
- The amount formatting is doubly wrong: `amount_in_max` is formatted against `tokenOut`'s decimals, and `amount_out` against `tokenIn`'s — wrong magnitudes too, not just wrong labels.

Concrete signer experience for "buy 1 WETH, pay ≤ 3000 USDC" (single-hop ExactOut, on-chain path `WETH || fee || USDC`):

- `Input Token: WETH`
- `Maximum Input: ≤ 0.000000000000003 WETH`  *(3000 raw USDC base units formatted with WETH 18-decimals)*
- `Output Token: USDC`
- `Output Amount: 0.000001 USDC`  *(1 raw WETH wei formatted with USDC 6-decimals)*

This is the same bug class the PR was filed to fix (wrong asset / wrong magnitude displayed at signing time), just on a different command. The single-hop fixture (`uniswap-v3swap.expected`) happens to be ExactIn so it never surfaced; pre-fix it was equally broken for ExactOut but the helper unification in this PR makes it newly load-bearing.

## The new test codifies the wrong model

`test_decode_v3_swap_exact_out_multi_hop_displays_final_token` builds its test path in naive ExactIn ordering (`[token_in, token_mid, token_out]`) and asserts the trailing 20 bytes show up as `Output Token`. That assertion passes only because the test itself constructed the path the wrong way around. A real Universal Router ExactOut payload (path constructed in reverse, as the contract requires) will display inverted. **This is actively misleading for future reviewers** — please don't merge it in this shape.

## Suggested fix shape

Two reasonable options:

**A — Caller-level swap (preferred)**: keep `decode_v3_path` returning `(first_20, last_20, fee_pct, hops, …)` as raw positions; in `decode_v3_swap_exact_out`, alias `let token_out = first_20; let token_in = last_20;` and feed `format_token_amount(chain_id, token_in, amount_in_max)` and `format_token_amount(chain_id, token_out, amount_out)`. The helper stays a pure byte-slicer; semantics live at the caller.

**B — Direction-aware helper**: add a `Direction { ExactIn, ExactOut }` parameter (or a sibling `decode_v3_path_exact_out`) that swaps the return values internally.

Then rewrite `test_decode_v3_swap_exact_out_multi_hop_displays_final_token` to encode the test path in reverse order (`[real_token_out, mid_fee, mid_token, last_fee, real_token_in]`) and assert that the signer sees the real `token_out` as `Output Token` with the right decimals.

## References

- ExactOut decoder: `src/chain_parsers/visualsign-ethereum/src/protocols/uniswap/contracts/universal_router.rs:767-843` *(line numbers from this PR's worktree)*
- Misleading test: `test_decode_v3_swap_exact_out_multi_hop_displays_final_token`
- Uniswap V3 periphery convention: `SwapRouter.exactOutputInternal` → `Path.decodeFirstPool` returns `(tokenB, tokenA, fee)` where `tokenB` is the output asset for the *current* hop. Path bytes are `tokenOut || fee0 || mid || fee1 || ... || tokenIn`.

pepe-anchor:
Hey, good catch — thanks for the depth on this one.

Landed in two commits on top of the previous push:

- [`e3b3540`](https://github.com/anchorageoss/visualsign-parser/commit/e3b3540) — token-role swap at the caller (your option A). `decode_v3_path` stays semantics-free; `decode_v3_swap_exact_out` now aliases `first_token = tokenOut`, `last_token = tokenIn`, and feeds `format_token_amount(chain_id, token_in, amount_in_max)` / `format_token_amount(chain_id, token_out, amount_out)`. Helper docs spell out the direction-dependent aliasing instead of baking ExactIn naming into the return values.
- [`e3b3540`](https://github.com/anchorageoss/visualsign-parser/commit/e3b3540) also rewrites `test_decode_v3_swap_exact_out_multi_hop_displays_final_token` to build the path in the reversed on-chain layout (`[real_token_out, mid, real_token_in]` with the fees in matching order) and asserts the signer sees the real tokens as `Output Token` / `Input Token` with the right decimals.

Extra on top, same bug class on the fee disclosure:

- [`b2a1ba2`](https://github.com/anchorageoss/visualsign-parser/commit/b2a1ba2) — the subtitle was surfacing the leading-byte fee as "first fee", which for ExactOut is actually the FINAL execution hop. The signer's input touches the trailing-byte fee first. `decode_v3_path` now returns both leading and trailing fees as raw positions; ExactIn keeps leading, ExactOut picks trailing. On the multi-hop fixture this was a 100x discrepancy (1% disclosed in place of 0.01%).

Marking ready for review, re-requesting your eyes.

shahan-khatchadourian-anchorage (approved):
lgtm 👍

PR #333, opened 2026-05-25
fix(ethereum): stop silently truncating Universal Router U256 amounts to zero

pepe-anchor:
## Summary

The Universal Router visualizer was running every on-chain `uint256` amount through `value.to_string().parse::<u128>().unwrap_or(0)` before handing it to `format_units`. Any value above `u128::MAX` was therefore rendered as `0` in the signing payload, hiding the real amount the user was about to authorise. A malicious frontend can craft an `amountMinimum` above `2^128` to make a SWEEP, WRAP_ETH, or V2/V3 swap look like "Sweep >=0 WETH ..." while actually emptying the account.

## What changed

- Added `ContractRegistry::format_token_amount_u256`, the U256-native counterpart of `format_token_amount`, which calls `alloy_primitives::utils::format_units` directly on the U256.
- Replaced every `u128 = value.to_string().parse().unwrap_or(0)` site in `universal_router.rs` (`decode_sweep`, `decode_wrap_eth`, `decode_unwrap_weth`, V2/V3 exact-in and exact-out swaps, `decode_pay_portion`) with the new U256 path.
- Replaced the bips `unwrap_or(0)` with `u128::try_from`. In range it formats as a percentage like today, out of range it now renders as `<raw> bips (raw)` instead of "0%". Bips realistically fit in `u128`, but the on-chain type is `uint256`, so out-of-range values must surface, not silently zero.
- Regression tests:
  - `registry::tests::test_format_token_amount_u256_above_u128_max_is_not_zero` (u128::MAX + 1)
  - `registry::tests::test_format_token_amount_u256_max_is_not_zero` (U256::MAX)
  - `registry::tests::test_format_token_amount_u256_matches_u128_in_range` (parity with the u128 path)
  - `universal_router::tests::test_decode_sweep_amount_above_u128_max_is_not_zero`
  - `universal_router::tests::test_decode_wrap_eth_amount_above_u128_max_is_not_zero`

Scope is Universal Router only, matching the ticket. `permit2.rs` and the DECODER_GUIDE.md still teach/use the old `unwrap_or(0)` pattern and should be cleaned up in follow-ups.

## Rollback

Revert this commit. No data or schema changes, no protobuf regeneration, no migrations. The new `format_token_amount_u256` method is additive and would simply become unused; the call sites in `universal_router.rs` revert to their previous string-parse-truncate behaviour.

## Test plan

- [x] `cargo fmt --check -p visualsign-ethereum` clean
- [x] `cargo clippy -p visualsign-ethereum --all-targets -- -D warnings` clean
- [x] `cargo test -p visualsign-ethereum` (204 unit + 4 lib + 3 doc tests pass, including the 5 new U256 truncation regressions)
- [ ] Manual: parse a Universal Router calldata that uses an `amountMinimum > u128::MAX` and confirm the rendered field shows the real amount, not "0".

### DIFF
--- a/src/chain_parsers/visualsign-ethereum/src/protocols/uniswap/contracts/universal_router.rs
+++ b/src/chain_parsers/visualsign-ethereum/src/protocols/uniswap/contracts/universal_router.rs
@@ -2879,4 +2879,52 @@ mod tests {
             _ => panic!("Expected TextV2 field"),
         }
     }
+
+    /// Regression: a PAY_PORTION with `bips` greater than `u128::MAX` must not
+    /// silently render as "0%". `bips` is a `uint256` on-chain, so an adversarial
+    /// value above `u128::MAX` must surface as a raw figure rather than a
+    /// percentage that would tell the signer they are paying out nothing.
+    #[test]
+    fn test_decode_pay_portion_bips_above_u128_max_is_not_zero() {
+        let token: Address = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
+            .parse()
+            .unwrap();
+        let recipient: Address = "0x0000000000000000000000000000000000000001"
+            .parse()
+            .unwrap();
+        // u128::MAX + 1: the smallest U256 the old u128 conversion could not represent.
+        let bips = U256::from(u128::MAX) + U256::from(1u64);
+
+        let encoded = (token, recipient, bips).abi_encode();
+        let field = UniversalRouterVisualizer::decode_pay_portion(&encoded, 1, None);
+
+        match field {
+            SignablePayloadField::PreviewLayout { preview_layout, .. } => {
+                let percentage = preview_layout
+                    .expanded
+                    .expect("expanded present")
+                    .fields
+                    .iter()
+                    .find_map(|f| match &f.signable_payload_field {
+                        SignablePayloadField::TextV2 { common, text_v2 }
+                            if common.label == "Percentage" =>
+                        {
+                            Some(text_v2.text.clone())
+                        }
+                        _ => None,
+                    })
+                    .expect("Percentage field present");
+                // The bug: "0%" / "0.0000%". The fix must surface the raw bips value.
+                assert!(
+                    !percentage.starts_with('0'),
+                    "bips above u128::MAX rendered as a near-zero percentage: {percentage}"
+                );
+                assert!(
+                    percentage.contains("bips (raw)"),
+                    "oversized bips must surface as a raw figure, got: {percentage}"
+                );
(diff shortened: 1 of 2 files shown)


prasanna-anchorage:
Heads up: this repo is **public**, so internal Linear refs (`PRS-xxx`) shouldn't leak into PR-visible surface — external readers can't resolve them.

This PR's leaks:
- Body links `linear.app/anchorlabs/issue/PRS-234` — swap for a one-line bug-class summary or a public GH issue reference.
- 13 code references / 13 inline comments mention `PRS-234` — rename to describe the bug class.
- Branch `pepefigueira/prs-234-...` also exposes the ID (optional).

Suggested convention: describe the bug in the name (e.g., `test_u256_amount_over_u128_max_not_truncated`). For persistent traceability, file a public GitHub issue.

pepe-anchor:
Scrubbed PRS-234 refs from body and code/comments. Branch left as-is since it'll disappear on merge.

shahan-khatchadourian-anchorage:
Tracking the pay-portion bips domain-validation as a follow-up in #352 (the `(raw)` branch only guards `> u128::MAX`; `bips > 10_000` still renders a bogus percentage, and the anomaly should surface as a structured diagnostic rather than display text). Out of scope for this PR — the `(raw)` boundary here is correct for the U256-truncation fix this PR targets.

shahan-khatchadourian-anchorage (dismissed):
lgtm 👍

prasanna-anchorage (approved):
Re-approving after rebasing onto main and adding the missing bips-overflow test.

**Rebase:** resolved the conflicts in `universal_router.rs` — both sides were purely additive (main's V3-path tests from #330 + this PR's PRS-234 U256 regression tests), so I kept both sets; the second commit's PRS-NNN comment scrub applied cleanly on top. `registry.rs` auto-merged.

**Follow-up addressed:** added `test_decode_pay_portion_bips_above_u128_max_is_not_zero`, which locks in the oversized-`bips` fallback this PR introduces (`u128::try_from(params.bips)` -> `"{} bips (raw)"`). It asserts the rendered percentage isn't a near-zero figure and surfaces the raw value; it fails on the old `to_string().parse::<u128>().unwrap_or(0)` path.

**Verified locally:** 235 + 9 crate tests pass (incl. the merged-in V3-path tests and all PRS-234 U256 tests), `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

**Out-of-scope follow-up filed:** the same truncation pattern lives in `permit2.rs` (amount + expiration) — tracked in #353, intentionally not expanded into this PR.

LGTM pending green CI.

PR #340, opened 2026-05-28
feat(tron): decode VoteWitnessContract + shared relative-time formatter

prasanna-anchorage:
## Summary
- Decode `VoteWitnessContract` in the Tron parser using a `PreviewLayout` so signers get a condensed summary (`N votes across M SRs`) with per-SR vote details available on expand.
- Extract timestamp formatting into a shared `visualsign::time_fmt` module that renders both an absolute UTC stamp and a relative tag (e.g. `about 2 hours ago` / `in 5 minutes`). Tron uses it to annotate Timestamp and Expiration alongside the raw epoch.
- Add a CLI-test normalizer that strips the relative-time suffix from snapshot comparisons so fixtures don't drift with wall-clock changes.

## Test Plan
- [x] `make -C src fmt && make -C src lint`
- [x] `make -C src test` (all suites pass locally)
- [ ] Verify stagex OCI image build runs via the `stagex` label

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -14047,6 +14047,7 @@ name = "visualsign"
 version = "0.1.0"
 dependencies = [
  "base64 0.22.1",
+ "chrono",
  "generated",
  "hex",
  "pretty_assertions",
--- a/src/chain_parsers/visualsign-tron/src/lib.rs
+++ b/src/chain_parsers/visualsign-tron/src/lib.rs
@@ -10,11 +10,14 @@ use anychain_tron::protocol::balance_contract::{
     UnDelegateResourceContract, UnfreezeBalanceV2Contract, WithdrawExpireUnfreezeContract,
 };
 use anychain_tron::protocol::common::ResourceCode;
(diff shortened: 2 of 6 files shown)


shahan-khatchadourian-anchorage (approved):
Solid PR. Implementation is correct, the PreviewLayout structure matches the spec, and the test coverage is thorough — every time bucket, both directions, the `None`-on-unrepresentable path, and real field-label/value assertions on the vote breakdown.

A few minor notes for awareness, none blocking:

- The comment above `condensed_fields` says "only the totals plus the owner are echoed" — `Owner` is actually a top-level sibling field outside the layout, not part of `condensed_fields`. Worth a one-line clarification if anyone touches that code again.
- The tron crate declares `chrono = "0.4"` (all defaults) while `visualsign/Cargo.toml` uses `default-features = false, features = ["clock"]` — harmless due to Cargo unification but worth making consistent.
- `fallback.clone()` where `fallback` is not used again — can be moved directly (clippy may flag this).
- No real mainnet `VoteWitnessContract` hex fixture yet; the other contract types have one. Fine to add in a follow-up.

PR #342, opened 2026-05-28
ci: harden CLA username matching and batch-PR handling

prasanna-anchorage:
## What

Latent robustness fixes in the CLA workflows, found while porting this flow to `anchorageoss/sqisign-rs` (Copilot flagged them there; the same code lives here).

- **Username matching** (`cla-file-check.yml`, `add-cla-user.yml` ×2): replace `grep -q -w "$user"` with `grep -Fxq "$user"` (fixed-string, whole-line). GitHub logins like `github-actions[bot]` / `dependabot[bot]` contain `[` which `grep -w` treats as a regex character class — it can error or false-negative. Also tolerate a missing `.cla-signed-users` (treat as unsigned) instead of erroring.
- **Batch-branch race** (`add-cla-user.yml`): add a `concurrency` group on `cla-signers-batch` (no cancel-in-progress) so two near-simultaneous `/approve-cla` runs don't race on the push and drop a signer.
- **Batch PR number parsing** (`add-cla-user.yml`): `gh pr create` prints the PR **URL** (`…/pull/123`), so `grep -o '#[0-9]\+'` matched nothing and left `NEW_PR` blank (broken follow-up comment). Parse the trailing number from the URL instead.

No behavior change for already-listed human signers; this only fixes bot logins, the concurrency race, and the empty PR-number bug.

## Test
`actionlint`/YAML parse clean. The matching change is verifiable: `printf 'github-actions[bot]\n' | grep -Fxq 'github-actions[bot]'` succeeds, whereas `grep -w` errors/misbehaves on the `[`.

### DIFF
--- a/.github/workflows/add-cla-user.yml
+++ b/.github/workflows/add-cla-user.yml
@@ -8,6 +8,14 @@ permissions:
   contents: write
   pull-requests: write
 
+# Serialize runs: this workflow pushes to the shared `cla-signers-batch` branch
+# and creates/updates a shared batch PR. Two near-simultaneous `/approve-cla`
+# comments would otherwise race on the push. Don't cancel in-progress runs --
+# each approval must complete so no signer is dropped.
+concurrency:
+  group: cla-signers-batch
+  cancel-in-progress: false
+
 jobs:
   add-user-to-cla-list:
     if: github.event.issue.pull_request
@@ -92,8 +100,10 @@ jobs:
 
           echo "Adding ${AUTHOR_USERNAME} to .cla-signed-users"
 
-          # Check if user is already in the list on main branch
-          if grep -v '^#' .cla-signed-users | grep -q -w "$AUTHOR_USERNAME"; then
+          # Check if user is already in the list on main branch.
+          # grep -Fxq: fixed-string whole-line match -- usernames like
+          # `dependabot[bot]` contain regex metacharacters that break `grep -w`.
+          if grep -v '^#' .cla-signed-users | grep -Fxq "$AUTHOR_USERNAME"; then
             echo "User is already in the list."
             gh pr comment ${{ github.event.issue.number }} --body "ℹ️ @${AUTHOR_USERNAME} is already in the CLA signers list."
             exit 0
@@ -113,8 +123,8 @@ jobs:
             git checkout -b "$BATCH_BRANCH"
           fi
 
-          # Check if user is already in the batch branch
-          if grep -v '^#' .cla-signed-users | grep -q -w "$AUTHOR_USERNAME"; then
+          # Check if user is already in the batch branch (fixed-string match).
+          if grep -v '^#' .cla-signed-users | grep -Fxq "$AUTHOR_USERNAME"; then
             echo "User is already in the batch branch."
             gh pr comment ${{ github.event.issue.number }} --body "ℹ️ @${AUTHOR_USERNAME} is already pending approval in the CLA signers batch."
             exit 0
@@ -143,6 +153,8 @@ jobs:
             gh pr comment ${{ github.event.issue.number }} --body "✅ @${AUTHOR_USERNAME} has been added to the CLA signers batch PR #${EXISTING_PR} by @${COMMENTER}. The CLA check will pass once the batch PR is merged."
           else
             echo "Creating new batch PR"
+            # `gh pr create` prints the PR URL (…/pull/<n>); take the trailing
+            # number. The old `grep '#[0-9]+'` matched nothing on a URL.
             NEW_PR=$(gh pr create \
               --title "chore: Add CLA signers batch" \
               --body "This PR adds new CLA signers in batch to comply with repository rules.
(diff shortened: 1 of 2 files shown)


.github/workflows/add-cla-user.yml
```diff
@@ -8,6 +8,14 @@ permissions:
   contents: write
   pull-requests: write
 
+# Serialize runs: this workflow pushes to the shared `cla-signers-batch` branch
+# and creates/updates a shared batch PR. Two near-simultaneous `/approve-cla`
+# comments would otherwise race on the push. Don't cancel in-progress runs --
+# each approval must complete so no signer is dropped.
+concurrency:
+  group: cla-signers-batch
+  cancel-in-progress: false
```
pepe-anchor, line 17: should: The `concurrency` group serializes two racing `/approve-cla` runs, but the comment above it claims more than the mechanism delivers: under 3+ near-simultaneous approvals a signer is still silently dropped.

<details>

GitHub Actions keeps at most one *pending* run per concurrency group: when a new run is queued, any previously pending run in the same group is **cancelled** (see docs: "Any previously pending job or workflow in the concurrency group will be canceled"). So with runs A, B, C arriving close together, A runs, B goes pending, then C arrives and cancels B. B never pushes, so that signer is dropped with no failure comment, which is exactly the `no signer is dropped` outcome the comment on lines 11-14 promises. Likelihood is low here (only two authorized maintainers, manual comments), so this is not a blocker, but either soften the comment to state the residual limitation, or make the add step resilient (re-fetch + append + retry push on non-fast-forward) so a cancelled/lost run can be safely re-driven. At minimum, do not assert an absolute guarantee the primitive cannot provide.

</details>

shahan-khatchadourian-anchorage:
Nice fix — confirmed the `[bot]` issue is real: `.cla-signed-users` already lists `github-actions[bot]`, and `grep -w` treats `[bot]` as a regex character class (matches one of `b`/`o`/`t`), so it could never match the literal bracketed login. `-Fxq` and the `gh pr create` URL-vs-`#N` parsing fix both check out too.

One non-blocking note on the new `concurrency` block: `cancel-in-progress: false` correctly fixes the 2-way race described in the PR body, but GitHub's concurrency queue only holds one *pending* run per group — if a third `/approve-cla` lands while one run is in-progress and another is already pending, the previously-pending run gets cancelled outright rather than queued further. So this narrows the race rather than eliminating it; a burst of 3+ near-simultaneous approvals could still silently drop a signer with no comment explaining why.

Given how rarely this fires (manual maintainer command), I wouldn't hold the PR on it. If it's ever worth closing all the way, the fix would be to stop relying on scheduling for correctness and retry the git push against the current remote state instead:

<details>
<summary>Illustrative sketch (not proposing this as-is)</summary>

```diff
           # Configure git
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
 
-          # Check if batch branch exists, if not create it from main
-          if git ls-remote --exit-code --heads origin "$BATCH_BRANCH" > /dev/null 2>&1; then
-            echo "Batch branch exists, checking it out"
-            git fetch origin "$BATCH_BRANCH"
-            git checkout "$BATCH_BRANCH"
-          else
-            echo "Creating new batch branch"
-            git checkout -b "$BATCH_BRANCH"
-          fi
+          pushed=false
+          for attempt in 1 2 3 4 5; do
+            if git ls-remote --exit-code --heads origin "$BATCH_BRANCH" > /dev/null 2>&1; then
+              git fetch origin "$BATCH_BRANCH"
+              git checkout -B "$BATCH_BRANCH" "origin/$BATCH_BRANCH"
+            else
+              git fetch origin main
+              git checkout -B "$BATCH_BRANCH" origin/main
+            fi
 
-          # Check if user is already in the batch branch (fixed-string match).
-          if grep -v '^#' .cla-signed-users | grep -Fxq "$AUTHOR_USERNAME"; then
-            echo "User is already in the batch branch."
-            gh pr comment ${{ github.event.issue.number }} --body "..."
-            exit 0
-          fi
+            if grep -v '^#' .cla-signed-users | grep -Fxq "$AUTHOR_USERNAME"; then
+              gh pr comment ${{ github.event.issue.number }} --body "..."
+              exit 0
+            fi
 
-          sed -i "/^#bots/i ${AUTHOR_USERNAME}" .cla-signed-users
-          git add .cla-signed-users
-          git commit -m "..."
-          git push origin "$BATCH_BRANCH"
+            sed -i "/^#bots/i ${AUTHOR_USERNAME}" .cla-signed-users
+            git add .cla-signed-users
+            git commit -m "..."
+            if git push origin "$BATCH_BRANCH"; then pushed=true; break; fi
+            sleep $(( (RANDOM % 3) + 1 ))
+          done
+          [ "$pushed" = true ] || { echo "::error::push failed after retries"; exit 1; }
```

Each attempt hard-resets to whatever's on `origin` and re-checks membership, so it converges under any amount of concurrent contention instead of depending on the queue never dropping a run.

</details>

shahan-khatchadourian-anchorage (approved):
My comments are non-blocking. LGTM

PR #359, opened 2026-06-03
draft: route external Custom chains from the --chain string

shahan-khatchadourian-anchorage:
## Problem

`parser_cli_core` exists so external workspaces can compose their own
`parser_cli` binary with a custom set of chain plugins (see the crate docs).
A plugin registers its converter under a `Chain`; a chain without a dedicated
enum variant uses `Chain::Custom("name")`.

But `chains::parse_chain` maps any unrecognized string to
`Chain::Unspecified`, and both code paths that turn `--chain <string>` into a
`Chain` go through it:

- plugin selection in `run` (`plugins.iter().find(|p| p.chain() == chain)`)
- registry lookup in `output::parse_and_display` (`convert_transaction(&parse_chain(chain), ...)`)

So a plugin registered under `Chain::Custom("foo")` can never be selected:
`parse_chain("foo")` yields `Unspecified`, which matches neither the plugin
nor its registry entry. The CLI even lists the chain as supported (that list
is built from the plugins) while rejecting it — a confusing contradiction.

## Fix

Map unrecognized chain strings to `Chain::Custom(string)` instead of
`Chain::Unspecified`, mirroring the existing `impl FromStr for Chain`
contract. Built-in chains are unaffected (matched first). The plugin that
registered under that `Custom` chain is now selected, so an external
workspace can add a chain purely by depending on its chain crate — no edit to
upstream chain-string parsing.

## Test plan

- [x] New unit tests in `chains.rs`: built-ins still map to their variant; an
      unknown string (`"near"`) maps to `Chain::Custom("near")`.
- [x] `cargo test -p parser_cli_core` — 28 passed.
- [x] `cargo test -p parser_cli` — 8 passed (no existing behavior regressed).
- [x] `make -C src fmt` + `make -C src lint` — clean across the workspace.

### DIFF
--- a/src/parser/cli-core/src/chains.rs
+++ b/src/parser/cli-core/src/chains.rs
@@ -13,18 +13,40 @@ fn chain_string_mapping() -> BTreeMap<&'static str, Chain> {
     mapping
 }
 
-/// Parses a chain string into a Chain enum value.
-/// Returns `Chain::Unspecified` if the chain string is not recognized.
+/// Parses a chain string into a `Chain`.
+///
+/// Built-in chains map to their dedicated variant. Any other string maps to
+/// `Chain::Custom`, so a chain contributed by an external [`ChainPlugin`] is
+/// selected by the plugin that registered under the matching `Chain::Custom`.
 #[must_use]
 pub fn parse_chain(chain_str: &str) -> Chain {
     chain_string_mapping()
         .get(chain_str)
         .cloned()
-        .unwrap_or(Chain::Unspecified)
+        .unwrap_or_else(|| Chain::Custom(chain_str.to_string()))
 }
 
 /// Returns a vector of all available chain names as string slices.
 #[must_use]
 pub fn available_chains() -> Vec<&'static str> {
     chain_string_mapping().keys().copied().collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_chain_maps_builtins() {
+        assert_eq!(parse_chain("ethereum"), Chain::Ethereum);
+        assert_eq!(parse_chain("solana"), Chain::Solana);
+    }
+
+    #[test]
+    fn parse_chain_maps_unknown_to_custom() {
+        // A chain contributed by an external ChainPlugin (e.g. NEAR) is not a
+        // built-in, but must still resolve so the plugin registered under the
+        // matching Chain::Custom is selected by `run`.
+        assert_eq!(parse_chain("near"), Chain::Custom("near".to_string()));
+    }
 }


src/parser/cli-core/src/chains.rs
```diff
@@ -13,18 +13,40 @@ fn chain_string_mapping() -> BTreeMap<&'static str, Chain> {
...
+/// Built-in chains map to their dedicated variant. Any other string maps to
+/// `Chain::Custom`, so a chain contributed by an external [`ChainPlugin`] is
+/// selected by the plugin that registered under the matching `Chain::Custom`.
 #[must_use]
 pub fn parse_chain(chain_str: &str) -> Chain {
     chain_string_mapping()
         .get(chain_str)
         .cloned()
-        .unwrap_or(Chain::Unspecified)
+        .unwrap_or_else(|| Chain::Custom(chain_str.to_string()))
```
prasanna-anchorage, line 26: nit (dead code): with this change `parse_chain` can never return `Chain::Unspecified` again — the mapping has no `"unspecified"` key, so even `--chain unspecified` now becomes `Custom("unspecified")`. That makes the `if chain == Chain::Unspecified { ... }` branch in `run()` (lib.rs:98) unreachable; every unrecognized input now hits the `else` ("not supported by this CLI build") and the dedicated "unrecognized chain" message is dead. The default CLI registers no `Unspecified` plugin so nothing functionally breaks, but please either delete that now-dead branch or consciously keep the distinction.

src/parser/cli-core/src/chains.rs
```diff
@@ -13,18 +13,40 @@ fn chain_string_mapping() -> BTreeMap<&'static str, Chain> {
...
 
-/// Parses a chain string into a Chain enum value.
-/// Returns `Chain::Unspecified` if the chain string is not recognized.
+/// Parses a chain string into a `Chain`.
+///
+/// Built-in chains map to their dedicated variant. Any other string maps to
+/// `Chain::Custom`, so a chain contributed by an external [`ChainPlugin`] is
+/// selected by the plugin that registered under the matching `Chain::Custom`.
 #[must_use]
 pub fn parse_chain(chain_str: &str) -> Chain {
```
prasanna-anchorage, line 22: nit (consistency): the doc says this mirrors `impl FromStr for Chain`, but it doesn't quite — `FromStr` lowercases (`registry.rs:49`) while `parse_chain` does an exact `BTreeMap` lookup. With the new `Custom` fallback, `--chain Ethereum` (capitalized) now resolves to `Custom("Ethereum")` and fails to select the Ethereum plugin. Consider implementing `parse_chain` as `Chain::from_str(s).unwrap_or(...)` (it's infallible and already returns `Custom` for unknowns): that genuinely mirrors `FromStr`, gains case-insensitivity for built-ins, and drops the duplicated chain list. `available_chains()` still needs `chain_string_mapping()`, so keep that. Optional.

prasanna-anchorage (commented):
Correct root-cause fix — both `--chain`->`Chain` paths funnel through `parse_chain`, so routing unknown strings to `Chain::Custom` lets an externally-registered plugin actually be selected. Two nits inline, plus: the title still says `draft:` though the draft flag is off and the test plan is all checked — worth cleaning up before merge.

prasanna-anchorage (approved):
Approving. Correct, minimal root-cause fix for Custom-chain routing. Nits from my earlier review (dead `Unspecified` branch, FromStr case-insensitivity, `draft:` title) are non-blocking — worth tidying but not gating.

src/parser/cli-core/src/chains.rs
```diff
@@ -13,18 +13,40 @@ fn chain_string_mapping() -> BTreeMap<&'static str, Chain> {
     mapping
 }
 
-/// Parses a chain string into a Chain enum value.
-/// Returns `Chain::Unspecified` if the chain string is not recognized.
+/// Parses a chain string into a `Chain`.
+///
+/// Built-in chains map to their dedicated variant. Any other string maps to
+/// `Chain::Custom`, so a chain contributed by an external [`ChainPlugin`] is
+/// selected by the plugin that registered under the matching `Chain::Custom`.
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed in #384: updated to `[crate::ChainPlugin]`.

src/parser/cli-core/src/chains.rs
shahan-khatchadourian-anchorage: Fixed in #384: removed the dead `Chain::Unspecified` branch; all unrecognized inputs now use the single "not supported by this CLI build" message.

src/parser/cli-core/src/chains.rs
shahan-khatchadourian-anchorage: Addressed in #384: `parse_chain` now delegates to `Chain::from_str`, gaining case-insensitive built-in matching and dropping the duplicated chain list.

PR #365, opened 2026-06-11
refactor: unify 0x/0X hex-prefix handling across chains

### LINKED ISSUE #364: Unify 0x hex-prefix handling across chain parsers
## Summary

`0x` hex-prefix handling is inconsistent across the chain parsers. PR #355 added
thorough `0x`/`0X` handling to the ABI/IDL signature paths, but a repo-wide sweep
shows the rest of the codebase accepts hex prefixes three different ways.

## Current tiers (on `main`)

| Tier | Sites | Prefix | `0x` | `0X` |
|------|-------|--------|------|------|
| A. signature / JSON-RPC / CLI-address | `abi_metadata.rs` ×3, `solana/idl/signature.rs` (`decode_hex_fixed`), `eth_json.rs` (`require_hex_prefix`), `cli_plugin.rs` (`validate_eth_address`/`normalize_eth_address`) | optional or required | ✅ | ✅ |
| B. Ethereum & Tron raw-tx decode | `ethereum/src/lib.rs`, `tron/src/lib.rs` (`decode_transaction`) | optional | ✅ | ❌ |
| C. Sui raw-tx decode | `sui/src/core/transaction/decoder.rs` | none | ❌ | ❌ |

### Consequences

- A `0X`-prefixed raw tx is accepted on Ethereum ABI paths but rejected by the
  Ethereum/Tron tx decoders.
- A `0x`-prefixed Sui raw tx fails to decode entirely (Sui strips nothing).
- The "strip optional `0x`/`0X`" logic is hand-rolled ~7 times; only Solana wrapped it.
  There is no single definition of "how we accept a hex string."

Related edge: `SupportedEncodings::detect` classifies any `0x`-prefixed string as
`Base64`, because `x` is not an ASCII hex digit.

## Decision

Standardize on a **lenient** policy via one shared helper:

- Accept an **optional** `0x`/`0X` prefix (case-insensitive) on raw-tx and
  signature/value paths.
- JSON-RPC fields keep **requiring** a prefix, but use the shared strip logic.
- This makes Sui newly accept `0x`-prefixed raw tx (intended behavior change).

## Shared helper — `visualsign::encodings`

```rust
pub fn strip_hex_prefix(value: &str) -> &str        // strip optional 0x/0X
pub fn split_hex_prefix(value: &str) -> Option<&str> // Some(body) iff prefixed
pub fn decode_hex(value: &str) -> Result<Vec<u8>, hex::FromHexError> // strip + decode
```

Plus a fix to `SupportedEncodings::detect` to strip the prefix before the hex-digit
test. Fixed-size decoders that need custom length errors (Solana) keep their wrapper
and delegate only the strip.

## Migration checklist

- [ ] `visualsign/src/encodings.rs` — add helper + tests; fix `detect`
- [ ] `visualsign/Cargo.toml` — promote `hex` to a (non-dev) dependency
- [ ] `chain_parsers/visualsign-ethereum/src/lib.rs` — `decode_transaction` hex branch
- [ ] `chain_parsers/visualsign-tron/src/lib.rs` — `decode_transaction` hex branch
- [ ] `chain_parsers/visualsign-sui/src/core/transaction/decoder.rs` — hex branch (behavior change)
- [ ] `chain_parsers/visualsign-ethereum/src/eth_json.rs` — `require_hex_prefix` via `split_hex_prefix`
- [ ] `chain_parsers/visualsign-ethereum/src/abi_metadata.rs` — signature, pubkey, `canonical_pubkey_from_hex`
- [ ] `chain_parsers/visualsign-solana/src/idl/signature.rs` — `decode_hex_fixed` delegates strip
- [ ] `chain_parsers/visualsign-ethereum/src/cli_plugin.rs` — `validate_eth_address` / `normalize_eth_address`
- [ ] CLAUDE.md (primary + ethereum module) — document the unified convention

Out of scope: `visualsign-unspecified` (no hex dep) and inner-data decodes (Solana
instruction data, discriminators) which take bare hex by design.

shahan-khatchadourian-anchorage:
## Summary

Centralizes `0x`/`0X` hex-prefix acceptance into `visualsign::encodings` and routes every chain's hex inputs through it, replacing per-crate hand-rolled prefix stripping. Closes #364.

New helpers in `visualsign::encodings` (the single definition of how the parser accepts a hex prefix):
- `strip_hex_prefix(&str) -> &str` — optional `0x`/`0X`, case-insensitive
- `split_hex_prefix(&str) -> Option<&str>` — for sites where the prefix is **mandatory** (returns `None` to turn into a caller error)
- `decode_hex(&str) -> Result<Vec<u8>, hex::FromHexError>` — strip + decode

`SupportedEncodings::detect()` now treats a leading `0x`/`0X` as hex (the `x` is stripped before the ASCII-hex-digit test).

## Per-chain changes

- **ethereum** — ABI signature/public-key decode, canonical pubkey, address validate/normalize, `require_hex_prefix`, and raw-transaction decode.
- **solana** — IDL signature decode, and `SolanaTransactionWrapper::from_string` hex arm so `0x`-prefixed Solana hex transactions decode consistently with the other chains (previously the only chain whose raw-tx path couldn't accept a prefix).
- **sui / tron** — raw-transaction decode; drop the now-unused per-crate `hex` dependency.
- **docs** — record the unified hex/`0x` convention in the root and ethereum `CLAUDE.md`.

## Behavior note

`detect()` previously classified `0x…`-prefixed input as **Base64** (the `x` isn't an ASCII hex digit); it now correctly classifies it as **Hex**. This makes prefixed hex transactions decode rather than fail.

## Test plan

- `cargo fmt` — no changes
- `cargo clippy --all-targets -- -D warnings` — clean (also enforced by pre-commit across all workspaces)
- `cargo test` for visualsign + all four chain crates — pass (648 tests across unit + integration suites)
- New unit tests in `encodings.rs` cover prefix presence/absence/case, mandatory-prefix, decode, and the `detect()` classification.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/visualsign.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/visualsign.rs
@@ -81,9 +81,8 @@ impl Transaction for SolanaTransactionWrapper {
             SupportedEncodings::Base64 => base64::engine::general_purpose::STANDARD
                 .decode(data)
                 .map_err(|e| TransactionParseError::DecodeError(e.to_string()))?,
-            SupportedEncodings::Hex => {
-                hex::decode(data).map_err(|e| TransactionParseError::DecodeError(e.to_string()))?
-            }
+            SupportedEncodings::Hex => visualsign::encodings::decode_hex(data)
+                .map_err(|e| TransactionParseError::DecodeError(e.to_string()))?,
         };
 
         // First try to decode as a VersionedTransaction
--- a/src/visualsign/src/encodings.rs
+++ b/src/visualsign/src/encodings.rs
@@ -10,9 +10,14 @@ pub enum SupportedEncodings {
 }
 
 impl SupportedEncodings {
-    /// Detect encoding format from string content
+    /// Detect encoding format from string content. A leading `0x`/`0X` prefix is
+    /// treated as hex (the `x` is not an ASCII hex digit, so it is stripped before
+    /// the test).
     pub fn detect(data: &str) -> Self {
-        if data.chars().all(|c| c.is_ascii_hexdigit()) {
+        if strip_hex_prefix(data)
+            .chars()
+            .all(|c| c.is_ascii_hexdigit())
+        {
             Self::Hex
         } else {
             Self::Base64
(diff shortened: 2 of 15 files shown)


src/visualsign/src/encodings.rs
```diff
@@ -10,9 +10,14 @@ pub enum SupportedEncodings {
 }
 
 impl SupportedEncodings {
-    /// Detect encoding format from string content
+    /// Detect encoding format from string content. A leading `0x`/`0X` prefix is
+    /// treated as hex (the `x` is not an ASCII hex digit, so it is stripped before
+    /// the test).
     pub fn detect(data: &str) -> Self {
-        if data.chars().all(|c| c.is_ascii_hexdigit()) {
+        if strip_hex_prefix(data)
```
prasanna-anchorage, line 17: nit: this PR makes `detect()` the sole router (the old Ethereum/Tron code force-routed any `0x` prefix to Hex unconditionally), which slightly widens the heuristic's ambiguity: a Base64 string that starts with `0x`/`0X` *and* whose remainder is all hex digits of even length now classifies as Hex. It's a pre-existing property (an all-hex Base64 string was already mis-detected), negligible for real ~1KB txs, and the consequence is a downstream decode error rather than a silent mis-parse. Worth a one-line note in the doc comment that classification is best-effort, since this is now the single decision point.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -81,9 +81,8 @@ impl Transaction for SolanaTransactionWrapper {
             SupportedEncodings::Base64 => base64::engine::general_purpose::STANDARD
                 .decode(data)
                 .map_err(|e| TransactionParseError::DecodeError(e.to_string()))?,
-            SupportedEncodings::Hex => {
-                hex::decode(data).map_err(|e| TransactionParseError::DecodeError(e.to_string()))?
-            }
+            SupportedEncodings::Hex => visualsign::encodings::decode_hex(data)
```
prasanna-anchorage, line 84: nit: the headline behavioral win here is chain-level — a `0x`-prefixed Solana tx now decoding where it previously failed — but the new tests live only in `encodings.rs` at the helper level. Consider one end-to-end assertion (`SolanaTransactionWrapper::from_string` with a `0x` prefix) so a future `detect()` refactor can't silently re-break this path.

prasanna-anchorage (approved):
Clean, well-scoped de-duplication — collapsing the copy-pasted `0x`/`0X` stripping into three tested helpers is the right move, and it incidentally fixes Solana's prefixed-hex path and the `0X` gap in Ethereum/Tron `from_string`. Approving. A few nits inline, plus two non-code notes:

- **PR description nit:** it says "sui / tron — drop the now-unused per-crate `hex` dependency," but tron correctly *keeps* `hex` (still uses `hex::encode` at `lib.rs:164/169/424`). Only sui dropped it. Code is right; just fix the description.
- **Consistency nit:** `visualsign-tron/Cargo.toml:17` still pins `hex = "0.4.3"` while solana was switched to `{ workspace = true }` — worth aligning tron too while touching deps.

src/visualsign/src/encodings.rs
shahan-khatchadourian-anchorage: Filed as item 1 in #374. Will add the one-line best-effort note to the `detect()` doc in the follow-up.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
shahan-khatchadourian-anchorage: Filed as item 2 in #374. Will add a `from_string("0x<hex>")` assertion in the Solana test module.

shahan-khatchadourian-anchorage:
Two top-level nits from the review body:

- **PR description** (sui / tron phrasing): noted — only sui dropped the direct `hex` dep; tron correctly keeps it. The description is historical at this point.
- **tron `hex` dep → `{ workspace = true }`**: filed as item 3 in #374 alongside the two inline nits.

Merging now.

PR #367, opened 2026-06-12
fix(solana): IDL presets degrade gracefully on v0+ALT accounts

### LINKED ISSUE #362: Solana IDL presets abort whole tx on v0 address-lookup-table accounts (resolve_accounts()? )
## Summary

Most IDL-based Solana visualizer presets call `context.resolve_accounts()?` (and `resolve_program_id()?`), which returns `Err` on the first unresolved account index. In the non-diagnostics dispatch (`core/instructions.rs`), a visualizer `Err` aborts the **entire transaction** rather than falling back to `unknown_program`. So a v0 transaction whose token/protocol accounts come from an **address lookup table** (which the parser does not resolve) erases the whole tx instead of rendering.

This is the same failure mode fixed for `spl_token` in #81 — flagged there by @shahan-khatchadourian-anchorage. `jupiter_perps` is the example called out, but it is systemic.

## Why it happens

- `can_handle` gates these presets on a **resolved** program_id, so `resolve_program_id()?` does not actually fire in practice — the real problem is `resolve_accounts()?`.
- An ALT-referenced account resolves to `AccountRef::Unresolved`, `resolve_accounts()` returns `Err`, dispatch propagates it, and `decode_instructions` (non-diagnostics) returns `Err` for the whole transaction.

## Intended contract

`VisualizerContext`'s docs describe the catch-all "partial rendering" pattern: substitute `unresolved(N)` placeholders for unresolved indices instead of erroring. #81 applied this to `spl_token` via an infallible `InstructionView`; the same shape (or a shared helper) should apply here.

## Affected presets (`resolve_accounts()?`)

- jupiter_perps
- jupiter_earn
- jupiter_borrow
- meteora_dlmm
- meteora_damm_v2
- metadao_futarchy
- metadao_conditional_vault
- orca_whirlpool
- exponent_finance
- onre_app
- kamino_limit
- kamino_vault
- kamino_borrow
- kamino_farms
- neutral_trade
- dflow_aggregator

## Suggested fix

Introduce a shared infallible account/program resolver (display strings with `unresolved(N)` placeholders) in `core` so presets stop using `resolve_accounts()?` on the render path, then migrate the presets above. Add a v0+ALT regression test per preset (or a shared one) asserting graceful `Ok` rendering rather than an aborted transaction.

## References

- Fix template: #81 (`spl_token` `InstructionView` + `test_v0_unresolved_accounts_render_placeholder_not_err`)
- Contract: `chain_parsers/visualsign-solana/src/core/mod.rs` (`VisualizerContext` resolution-patterns doc)
- Abort path: `chain_parsers/visualsign-solana/src/core/instructions.rs` (non-diagnostics `decode_instructions`)

shahan-khatchadourian-anchorage:
## Summary

Closes #362.

In v0 transactions that reference accounts via Address Lookup Tables (ALTs), `account_keys` only contains the static keys. Any ALT-resolved account appears as `AccountRef::Unresolved { raw_index }` in `VisualizerContext`. Before this fix, all 17 IDL-based presets called `context.resolve_accounts()?` at the top of `visualize_tx_commands`, which would return `Err` on the first unresolved index — propagating through the non-diagnostics dispatch path as a whole-transaction failure.

- Add `InstructionView::from_context()` to `core/mod.rs`: infallibly resolves all accounts to `String`, substituting `"unresolved(N)"` for any ALT/OOB index
- Migrate all 16 IDL presets and `token_2022` from `resolve_accounts()?` / `resolve_program_id()?` to `InstructionView::from_context()`; `build_named_accounts` functions updated from `&[AccountMeta]` to `&[String]`
- Add unit test `test_instruction_view_substitutes_unresolved_placeholder` in `core/mod.rs`
- Add dispatch regression test `test_idl_preset_with_alt_accounts_renders_ok` in `core/instructions.rs` that verifies an OOB account index (simulating v0+ALT) no longer aborts the instruction

## Test plan

- [ ] `cargo test -p visualsign-solana` — all tests pass
- [ ] `make -C src lint` — clippy clean with `-D warnings`
- [ ] `test_idl_preset_with_alt_accounts_renders_ok` specifically verifies the regression is fixed: an instruction with account index 50 (OOB/ALT) now renders as `"unresolved(50)"` instead of returning `Err`

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/instructions.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/instructions.rs
@@ -419,6 +419,30 @@ mod off_tests {
             decode_instructions(&tx, &registry).expect("OOB account_index should not abort");
         assert_eq!(fields.len(), 1);
     }
+
+    #[test]
+    fn test_idl_preset_with_alt_accounts_renders_ok() {
+        // Simulate a v0 transaction where Jupiter Perps is the program and
+        // account index 50 is an ALT-referenced account (OOB in account_keys).
+        // Before the fix, resolve_accounts()? in jupiter_perps aborts the whole
+        // transaction. After the fix, it renders a fallback Ok field instead.
+        let jupiter_perps_pk: Pubkey = "PERPHjGBqRHArX4DySjwM6UJHiR3sWAatqfdBS2qQJu"
+            .parse()
+            .expect("valid pubkey");
+        let sender_pk = Pubkey::new_unique();
+        let tx = tx_with(
+            vec![sender_pk, jupiter_perps_pk],
+            vec![solana_sdk::instruction::CompiledInstruction {
+                program_id_index: 1,
+                accounts: vec![0, 50], // 50 is OOB — simulates an ALT account
+                data: vec![0u8; 8],    // unknown discriminator -> fallback Ok
+            }],
+        );
+        let registry = IdlRegistry::new();
+        let fields = decode_instructions(&tx, &registry)
+            .expect("v0+ALT account must not abort an IDL preset");
+        assert_eq!(fields.len(), 1, "one field per instruction");
+    }
 }
 
 #[cfg(all(test, feature = "diagnostics"))]
--- a/src/chain_parsers/visualsign-solana/src/core/instructions.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/instructions.rs
@@ -443,6 +443,54 @@ mod off_tests {
             .expect("v0+ALT account must not abort an IDL preset");
         assert_eq!(fields.len(), 1, "one field per instruction");
     }
+
+    #[test]
+    fn test_idl_preset_named_account_renders_unresolved_for_alt() {
+        use visualsign::SignablePayloadField;
+        // DFlow Aggregator `close_empty_token_account` has a real discriminator and
+        // zero args, so parse_instruction_with_idl succeeds with 8 bytes of data.
+        // Account slot 0 (token_account in the IDL) maps to wire index 50, which is
+        // OOB — simulating an ALT-referenced account. Before the fix the preset
+        // aborted; after the fix `build_named_accounts` maps it to "unresolved(50)"
+        // and that string appears in the rendered expanded fields.
+        let dflow_pk: Pubkey = "DF1ow4tspfHX9JwWJsAb9epbkA8hmpSEAtxXy1V27QBH"
+            .parse()
+            .expect("valid pubkey");
+        let sender_pk = Pubkey::new_unique();
+        let tx = tx_with(
+            vec![sender_pk, dflow_pk],
+            vec![solana_sdk::instruction::CompiledInstruction {
+                program_id_index: 1,
+                // accounts[0]=50 (OOB/ALT) maps to IDL "token_account"
+                accounts: vec![50, 0, 0, 0],
+                // close_empty_token_account discriminator, no args needed
+                data: vec![232, 75, 140, 136, 250, 78, 224, 188],
+            }],
+        );
+        let registry = IdlRegistry::new();
+        let fields = decode_instructions(&tx, &registry)
+            .expect("v0+ALT account must not abort an IDL preset");
+        assert_eq!(fields.len(), 1);
+
+        // Dig into the expanded preview fields and assert that the IDL-named
+        // account slot that references an ALT entry renders as "unresolved(50)".
+        let SignablePayloadField::PreviewLayout { preview_layout, .. } =
+            &fields[0].signable_payload_field
+        else {
+            panic!("expected PreviewLayout");
+        };
+        let expanded = preview_layout.expanded.as_ref().expect("expanded fields present");
+        let found = expanded.fields.iter().any(|f| {
+            matches!(&f.signable_payload_field,
+                SignablePayloadField::TextV2 { text_v2, .. }
+                if text_v2.text == "unresolved(50)")
+        });
+        assert!(
+            found,
+            "expected an expanded field with value 'unresolved(50)' for the ALT-referenced \
+             token_account slot; got fields: {expanded:#?}"
+        );
+    }
 }
 
 #[cfg(all(test, feature = "diagnostics"))]
--- a/src/chain_parsers/visualsign-solana/src/core/instructions.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/instructions.rs
@@ -479,7 +479,10 @@ mod off_tests {
         else {
             panic!("expected PreviewLayout");
         };
-        let expanded = preview_layout.expanded.as_ref().expect("expanded fields present");
+        let expanded = preview_layout
+            .expanded
+            .as_ref()
+            .expect("expanded fields present");
         let found = expanded.fields.iter().any(|f| {
             matches!(&f.signable_payload_field,
                 SignablePayloadField::TextV2 { text_v2, .. }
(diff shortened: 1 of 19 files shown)


prasanna-anchorage (commented):
Solid, well-motivated fix. A few inline notes — main one is the unused `view.data` allocation; the rest are minor (panic-free totality, a test-coverage gap, and a UX confirmation).

src/chain_parsers/visualsign-solana/src/core/mod.rs
```diff
@@ -219,6 +220,43 @@ impl<'a> VisualizerContext<'a> {
...
+                Some(AccountRef::Unresolved { raw_index }) => format!("unresolved({raw_index})"),
+                // `i` is in `0..num_accounts()` which equals `accounts.len()`,
+                // so `account(i)` returning `None` is unreachable.
+                None => unreachable!("account index {i} is within num_accounts()"),
+            })
+            .collect();
+        Self {
+            program_id,
+            accounts,
+            data: context.data().to_vec(),
```
prasanna-anchorage: **`view.data` is a dead allocation.** This clones the full instruction payload (`context.data().to_vec()`) for every instruction, but no preset reads `view.data` — they all keep `let data = context.data();` and use that borrow directly. So this is a per-instruction heap allocation with no consumer. Either drop the field, or have presets consume `view.data` instead of re-calling `context.data()`.

src/chain_parsers/visualsign-solana/src/core/mod.rs
```diff
@@ -219,6 +220,43 @@ impl<'a> VisualizerContext<'a> {
...
+            ProgramRef::Resolved(pk) => pk.to_string(),
+            ProgramRef::Unresolved { raw_index } => format!("unresolved({raw_index})"),
+        };
+        let accounts = (0..context.num_accounts())
+            .map(|i| match context.account(i) {
+                Some(AccountRef::Resolved(pk)) => pk.to_string(),
+                Some(AccountRef::Unresolved { raw_index }) => format!("unresolved({raw_index})"),
+                // `i` is in `0..num_accounts()` which equals `accounts.len()`,
+                // so `account(i)` returning `None` is unreachable.
+                None => unreachable!("account index {i} is within num_accounts()"),
```
prasanna-anchorage: **`unreachable!()` reintroduces a panic path in non-test code.** It's genuinely unreachable given the `0..num_accounts()` bound, and it slips past the workspace `panic = "deny"` lint (`clippy::panic` doesn't flag `unreachable!`). But the whole premise of `from_context` is *infallible* rendering — a panic arm undercuts that. Cleaner to keep it total, e.g. `None => format!("unresolved(oob:{i})")`, so there's no panic path at all. Minor.

src/chain_parsers/visualsign-solana/src/core/instructions.rs
```diff
@@ -419,6 +419,30 @@ mod off_tests {
             decode_instructions(&tx, &registry).expect("OOB account_index should not abort");
         assert_eq!(fields.len(), 1);
     }
+
+    #[test]
+    fn test_idl_preset_with_alt_accounts_renders_ok() {
```
prasanna-anchorage, line 424: **This regression test doesn't exercise the path it's named for.** `data: vec![0u8; 8]` is an unknown discriminator, so it only hits the *fallback* branch and proves "doesn't abort." It never reaches `build_named_accounts`, which is where an ALT account actually becomes `"unresolved(N)"` against a matched IDL instruction. The unit test in `core/mod.rs` covers the substitution mechanism in isolation, but no test asserts a *named* account (e.g. an IDL `authority`) renders as `"unresolved(N)"` end-to-end. Consider a test with a real discriminator + an OOB account index asserting the named-account value.

src/chain_parsers/visualsign-solana/src/presets/token_2022/mod.rs
```diff
@@ -52,12 +51,14 @@ impl InstructionVisualizer for Token2022Visualizer {
...
     ) -> Result<AnnotatedPayloadField, VisualSignError> {
-        // Build AccountMeta shim for the parser (which expects &[AccountMeta]).
         // Unresolved accounts are rejected rather than substituted with
-        // Pubkey::default(), which would render as a valid-looking address.
-        let accounts: Vec<AccountMeta> = (0..context.num_accounts())
+        // "unresolved(N)". Authority fields (pause_authority, mint_authority,
+        // freeze_authority, etc.) are security-critical; showing an unresolvable
+        // address would let a user approve an instruction whose authority they
+        // cannot verify.
+        let accounts: Vec<String> = (0..context.num_accounts())
```
prasanna-anchorage, line 59: Good call keeping `token_2022` fail-closed on unresolved authority accounts — the asymmetry vs. the IDL presets is well reasoned and documented.

One UX confirmation (not blocking): the IDL presets surface authority-like accounts by IDL name (`authority`, `owner`, …), and with this change those will now display as `"unresolved(N)"` when ALT-resolved. That's strictly more honest than the previous whole-transaction abort and consistent with the placeholder design — just worth a conscious confirmation that "authority -> unresolved(5)" is the intended signing UX for IDL presets, rather than the explicit rejection `token_2022` uses here.

src/chain_parsers/visualsign-solana/src/core/mod.rs
```diff
@@ -179,6 +180,41 @@ impl<'a> VisualizerContext<'a> {
...
+/// resolves to either its base58 pubkey string or an `"unresolved(N)"` placeholder
+/// for indices that reference an address lookup table (v0 transactions) or are
+/// otherwise out of bounds. Presets use this instead of
+/// `resolve_accounts()?` / `resolve_program_id()?` on the render path so that
+/// a v0+ALT instruction degrades gracefully rather than aborting the transaction.
+pub struct InstructionView {
+    pub program_id: String,
+    pub accounts: Vec<String>,
+    pub data: Vec<u8>,
+}
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed: removed  entirely in commit 21d478b — no preset was reading it, so presets continue to call  directly.

src/chain_parsers/visualsign-solana/src/core/mod.rs
shahan-khatchadourian-anchorage: Fixed: dropped the `data` field from `InstructionView` entirely (commit 21d478b); presets already hold `context.data()` borrows, so no allocation needed.

src/chain_parsers/visualsign-solana/src/core/mod.rs
shahan-khatchadourian-anchorage: Fixed: replaced `unreachable!()` with `None => format!("unresolved(oob:{i})")` (commit 21d478b) — stays total and infallible with no panic path.

src/chain_parsers/visualsign-solana/src/core/instructions.rs
shahan-khatchadourian-anchorage: Fixed: added `test_idl_preset_named_account_renders_unresolved_for_alt` (commit 21d478b) using the DFlow `close_empty_token_account` discriminator so `build_named_accounts` matches and the test asserts the expanded field for the ALT slot equals `"unresolved(50)"` end-to-end.

src/chain_parsers/visualsign-solana/src/presets/token_2022/mod.rs
shahan-khatchadourian-anchorage: Acknowledged: confirmed intentional — IDL presets showing `"unresolved(N)"` for authority-like fields is the designed signing UX for v0+ALT; `token_2022` stays fail-closed because its account fields are the security-critical content, not display labels.

shahan-khatchadourian-anchorage:
Thanks @prasanna-anchorage. RFR

prasanna-anchorage (approved):
Approving. Solid graceful-degradation fix for v0+ALT accounts. Earlier inline nits (unused `view.data` allocation, the `unreachable!` arm, and tightening the regression test) are non-blocking — the `view.data` one is the only one I would still like cleaned up, but it does not gate.

PR #372, opened 2026-06-17
refactor(solana): remove silent-drop pattern in drift preset

### LINKED ISSUE #371: refactor(solana): drift preset uses silent-drop pattern for field builder errors
## Context

The Drift Protocol visualizer preset (`src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs`, merged in #366) uses `if let Ok(f) = create_text_field(...)` / `if let Ok(f) = create_raw_data_field(...)` throughout `build_parsed_fields`, `build_fallback_fields`, and `append_raw_data`. This silently swallows `VisualSignError`s from the field builders:

\`\`\`rust
// current — errors silently discarded
if let Ok(f) = create_text_field("Program", "Drift") {
    condensed_fields.push(f);
}
if let Ok(f) = create_raw_data_field(data, Some(hex_str.to_string())) {
    fields.push(f);
}
\`\`\`

Both builders called here (`create_text_field`, `create_raw_data_field`) are currently infallible and return `Ok(...)` unconditionally, so no field is actually being dropped today. However, the `if let Ok` pattern is a latent silent-drop hazard: if either builder ever becomes fallible, Drift would silently omit fields with no signal to callers.

## Fix

Convert the three helpers to return `Result<_, VisualSignError>` and replace `if let Ok` guards with `?`, matching the style used in `exponent_finance` and `dflow_aggregator`.

shahan-khatchadourian-anchorage:
## Summary

- `build_parsed_fields`, `build_fallback_fields`, and `append_raw_data` used `if let Ok(f) = ...` guards that silently discarded any `VisualSignError` from the field builders — a field builder failure would drop that field with no signal to the caller
- All three helpers now return `Result<_, VisualSignError>` and use `?` at each call site; `visualize_tx_commands` propagates via `?`
- Both field builders called here (`create_text_field`, `create_raw_data_field`) are currently infallible and return `Ok(...)` unconditionally, so no live field was actually being dropped; this removes the latent silent-drop pattern and aligns Drift with the style used in `exponent_finance` and `dflow_aggregator`

Closes #371.

## Test plan

- [x] `make -C src test` passes (existing `semantic_drift_deposit` integration test + 3 new unit tests for each converted helper)
- [x] `make -C src lint` passes (clippy `-D warnings`, no `unwrap`/`expect`/`panic` outside test module)

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
@@ -251,7 +251,7 @@ mod tests {
     #[test]
     fn test_build_fallback_fields_returns_result() {
         let (title, condensed, expanded) =
-            build_fallback_fields("dRiftyHA39MWEi3m9aunc5MzRF1JYuBsbn6VPcn33UH").unwrap();
+            build_fallback_fields(DRIFT_PROGRAM_ID).unwrap();
         assert_eq!(title, "Drift: Unknown Instruction");
         assert_eq!(condensed.len(), 2);
         assert_eq!(expanded.len(), 2);
@@ -261,7 +261,7 @@ mod tests {
     fn test_build_parsed_fields_returns_result() {
         let instruction = make_parsed_instruction("deposit");
         let (title, condensed, expanded) =
-            build_parsed_fields(&instruction, "dRiftyHA39MWEi3m9aunc5MzRF1JYuBsbn6VPcn33UH")
+            build_parsed_fields(&instruction, DRIFT_PROGRAM_ID)
                 .unwrap();
         assert_eq!(title, "Drift: deposit");
         assert!(!condensed.is_empty());
--- a/src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
@@ -40,14 +40,14 @@ impl InstructionVisualizer for DriftVisualizer {
         let parsed = parse_drift_instruction(data, &accounts);
 
         let (title, condensed_fields, expanded_fields) = match parsed {
-            Ok(parsed) => build_parsed_fields(&parsed, &program_id),
-            Err(_) => build_fallback_fields(&program_id),
+            Ok(parsed) => build_parsed_fields(&parsed, &program_id)?,
+            Err(_) => build_fallback_fields(&program_id)?,
         };
 
         let condensed = SignablePayloadFieldListLayout {
             fields: condensed_fields,
         };
-        let expanded_with_raw = append_raw_data(expanded_fields, data, &instruction_data_hex);
+        let expanded_with_raw = append_raw_data(expanded_fields, data, &instruction_data_hex)?;
         let expanded = SignablePayloadFieldListLayout {
             fields: expanded_with_raw,
         };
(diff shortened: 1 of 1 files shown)


src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
```diff
@@ -252,6 +232,49 @@ fn format_arg_value(value: &serde_json::Value) -> String {
...
+            },
+            named_accounts: BTreeMap::new(),
+        }
+    }
+
+    #[test]
+    fn test_build_fallback_fields_returns_result() {
+        let (title, condensed, expanded) =
+            build_fallback_fields("dRiftyHA39MWEi3m9aunc5MzRF1JYuBsbn6VPcn33UH").unwrap();
+        assert_eq!(title, "Drift: Unknown Instruction");
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed: both tests now use `DRIFT_PROGRAM_ID` instead of the inline string literal.

src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
```diff
@@ -252,6 +232,49 @@ fn format_arg_value(value: &serde_json::Value) -> String {
...
+        assert_eq!(condensed.len(), 2);
+        assert_eq!(expanded.len(), 2);
+    }
+
+    #[test]
+    fn test_build_parsed_fields_returns_result() {
+        let instruction = make_parsed_instruction("deposit");
+        let (title, condensed, expanded) =
+            build_parsed_fields(&instruction, "dRiftyHA39MWEi3m9aunc5MzRF1JYuBsbn6VPcn33UH")
+                .unwrap();
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed: both tests now use `DRIFT_PROGRAM_ID` instead of the inline string literal.

src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
```diff
@@ -190,52 +179,43 @@ fn build_parsed_fields(
...
     mut fields: Vec<AnnotatedPayloadField>,
     data: &[u8],
     hex_str: &str,
-) -> Vec<AnnotatedPayloadField> {
-    if let Ok(f) = create_raw_data_field(data, Some(hex_str.to_string())) {
-        fields.push(f);
-    }
-    fields
+) -> Result<Vec<AnnotatedPayloadField>, VisualSignError> {
+    fields.push(create_raw_data_field(data, Some(hex_str.to_string()))?);
```
prasanna-anchorage, line 217: nit: `create_raw_data_field` is currently infallible (field_builders.rs:160 returns `Ok(...)` unconditionally), so this `?` can't fire today. The change is still correct and worth keeping for consistency/future-proofing — just noting that, like the text-field sites, there's no reachable drop being fixed here.

src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
```diff
@@ -252,6 +232,49 @@ fn format_arg_value(value: &serde_json::Value) -> String {
...
+    fn test_build_fallback_fields_returns_result() {
+        let (title, condensed, expanded) =
+            build_fallback_fields(DRIFT_PROGRAM_ID).unwrap();
+        assert_eq!(title, "Drift: Unknown Instruction");
+        assert_eq!(condensed.len(), 2);
+        assert_eq!(expanded.len(), 2);
+    }
+
+    #[test]
+    fn test_build_parsed_fields_returns_result() {
```
prasanna-anchorage, line 261: nit: these `*_returns_result` tests only assert the happy path (`.unwrap()` + non-empty), and they can't exercise error propagation because the builders never fail. Fine to keep as "still renders" guards, but the names oversell what they verify. Also: the PR's test-plan checkboxes are still unchecked — please confirm `make -C src test` and `make -C src lint` are green before merge (watch `clippy::type_complexity` on the new `Result<(String, Vec<..>, Vec<..>), _>` shape; siblings use it so likely fine).

prasanna-anchorage (commented):
The refactor is good — replacing the `if let Ok(f) = ... { push(f) }` silent-drop pattern with `?`-propagation is the right, consistent style (matches exponent_finance / dflow_aggregator). One honesty note on the framing, plus minor nits inline.

**Description overstates this as a bugfix.** Both builders this code calls are currently infallible: `create_text_field` (field_builders.rs:32) and `create_raw_data_field` (field_builders.rs:160) both unconditionally return `Ok(...)`. So the premise — "a field builder failure would drop that field with no signal" / "`append_raw_data` could silently omit the raw-data fallback entirely" — is not reachable today; no field is or was actually being dropped, and every `else` of those `if let Ok` guards was dead. This is legitimate latent-risk hardening (if a builder ever becomes fallible, drift now does the right thing) and consistency cleanup — but the description and #371 should be toned down from "fixes silent field drops" to "removes a silent-drop pattern," so the changelog doesn't claim a user-visible fix that isn't.

prasanna-anchorage (approved):
Approving as a consistency/hardening cleanup. Only ask is non-blocking: tone down the description/#371 framing since the builders are currently infallible (this removes a silent-drop pattern rather than fixing live field drops), and confirm test+lint are green.

src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
shahan-khatchadourian-anchorage: Acknowledged — the  is kept for consistency with the other call sites in this function, as you note. Tracked the infallibility observation in the PR description update.

src/chain_parsers/visualsign-solana/src/presets/drift/mod.rs
shahan-khatchadourian-anchorage: Renamed in PR #382 (stacked on this one): test_build_fallback_fields_renders_unknown_instruction, test_build_parsed_fields_renders_deposit_instruction, test_append_raw_data_appends_one_field.

PR #386, opened 2026-06-22
revert(solana): stop fail-closing V0 transactions that reference ALTs (#324)

prasanna-anchorage:
## Summary

PR #324 added `reject_v0_if_any_ix_references_alts`, which refuses to render any V0 transaction whose instructions reference address-lookup-table (ALT) entries. The parser cannot resolve ALT contents offline, so the intent was to avoid showing a partial transaction.

In practice this fail-closed gate blocks **all** ALT-backed V0 transactions (e.g. aggregator swaps) from being signed — they fail at initiation with a 400 from the parser:

```
Cannot render V0 transaction: 3 address lookup table(s) present, 22 unresolved
ALT reference(s) ... Refusing to display a partial transaction.
```

This surfaced in staging via `sol_signTransaction` in wallet

## What this does

- Removes `reject_v0_if_any_ix_references_alts` and its call site in `convert_v0_to_visual_sign_payload`.
- With the gate gone, V0+ALT transactions render again through the graceful-degradation path added in #367: ALT-backed accounts surface as placeholders instead of dropping the instruction silently.
- Replaces the rejection-locking tests with `renders_v0_when_account_index_lives_behind_an_alt` (asserts these render). Keeps the existing "ALTs present but only static refs" and "malformed-without-ALTs" behavior tests unchanged.

## Tradeoff / follow-up

This reinstates the limitation #324 guarded against: ALT-backed references render as placeholders, so an instruction's true accounts aren't fully shown. The proper fix is to pass resolved ALT contents into the parser from the caller (wallets can supply ALT data) — a larger design change tracked separately.

## Verification

- `parser_cli` on the reported failing transaction now renders fully (was a 400).
- `cargo test -p visualsign-solana` — all pass (default + `diagnostics` feature).
- `cargo clippy -p visualsign-solana --all-targets -- -D warnings` — clean; `cargo fmt` applied.
- No fixtures or integration tests asserted the rejection string.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/visualsign.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/visualsign.rs
@@ -491,157 +491,6 @@ fn convert_versioned_to_visual_sign_payload(
     }
 }
 
-/// Reject a V0 transaction when at least one instruction references a
-/// program_id or account that lives behind an address lookup table, or that
-/// is out of range entirely.
-///
-/// Rationale: VisualSign converts raw bytes; it has no RPC and cannot
-/// resolve ALT entries. Without resolution, the only safe choice is to refuse
-/// to render -- otherwise the displayed payload would omit instructions or
-/// account references that the validator will still execute, and a malicious
-/// instruction could be hidden behind an ALT entry while the user signs a
-/// benign-looking display.
-///
-/// The resolved account vector in v0 is `account_keys` followed by, for each
-/// lookup, the addresses pulled at `writable_indexes` then at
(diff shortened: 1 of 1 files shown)


prasanna-anchorage:
Follow-up tracking ticket for the proper fix (caller-supplied ALT resolution): [PRS-512](https://linear.app/anchorlabs/issue/PRS-512/resolve-solana-v0-alt-contents-in-parser-via-caller-supplied-lookup)

hassan-anchor (approved):

PR #390, opened 2026-06-23
fix(visualsign): render `number` fields as amount_v2 (VSP has no `number` type)

prasanna-anchorage:
## Problem
The Anchorage wallet's Visual Signing Protocol engine has no `number` field type — its `FieldType` decoder recognizes only `text_v2`, `address_v2`, `amount_v2`, `preview_layout`, `delta`, `highlight`, `rule`, `accordion`, `diagnostic`. Any field emitted by `create_number_field` (`compute_budget`, `system`, `token_2022`, `jupiter_swap`, …) therefore decodes to `.unknown` and is surfaced as an **unsupported field**, also flagging its enclosing `preview_layout` as `containsUnsupportedNestedFields`.

## Fix
Render the `Number` variant as `text_v2` at the **serialization layer** (`serialize_to_map`/`get_expected_fields`), carrying the human value (incl. any unit) from `fallback_text`. The in-memory `SignablePayloadField::Number` variant and the `create_number_field` helper are unchanged — callers and variant-matching tests keep working; only the wire output changes. They stay visibly distinct from `amount_v2` (these are not amounts).

General fix, independent of the DFlow render work.

## Tests
`cargo test -p visualsign -p visualsign-solana`, `clippy --all-targets -D warnings`, `cargo fmt --check` pass. The one serialization-output assertion in `jupiter_swap` (Slippage field type) updated `number` → `text_v2`.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/presets/jupiter_swap/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/jupiter_swap/mod.rs
@@ -812,9 +812,12 @@ mod tests {
                         && field.get("Type").and_then(|t| t.as_str()) == Some("text_v2")
                 });
 
+                // `number` fields render as `amount_v2` on the wire (VSP has no
+                // `number` type; numbers are unitless/with-unit amounts); the
+                // in-memory variant stays Number.
                 let has_slippage = fields_array.iter().any(|field| {
                     field.get("Label").and_then(|l| l.as_str()) == Some("Slippage")
-                        && field.get("Type").and_then(|t| t.as_str()) == Some("number")
+                        && field.get("Type").and_then(|t| t.as_str()) == Some("amount_v2")
                 });
 
                 assert!(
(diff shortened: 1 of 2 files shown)


prasanna-anchorage:
Updated per the VSP design: there is **no `number` field type** (HSM `vsp.py` + iOS decoder list `text_v2/address_v2/amount_v2/preview_layout/list_layout/delta/accordion/diagnostic/divider`). So `number` now serializes as **`amount_v2`** (amount = value, abbreviation = unit; unitless = empty abbreviation, which iOS `parseAmount` accepts). In-memory `Number` variant + `create_number_field` unchanged; only the wire output. Force-pushed (SSH-signed).

(Branch name `fix/number-fields-text-v2` is legacy from the earlier text_v2 approach.)

prasanna-anchorage:
Superseded by https://github.com/anchorageoss/visualsign-parser/pull/393 — same change, renamed branch `fix/number-fields-amount-v2` to match the amount_v2 content (this branch was the legacy `text-v2` name).

PR #392, opened 2026-06-23
feat(visualsign): Anchorage wallet render-support validator

prasanna-anchorage:
## What
A validator that checks a `SignablePayload` only uses fields the **Anchorage wallet** can render, mirroring the wallet's VSP field support matrix — so the parser can catch payloads that would surface as "unsupported fields" before they ship (VSP is WYSIWYS, so an unrenderable field is a correctness bug).

- `visualsign::SignablePayload::validate_anchorage_wallet_renderable()` — hard gate (errs listing offending paths).
- `anchorage_render_findings()` — every unrenderable field with a JSON-ish path (e.g. `Fields[4].Expanded.Fields[6]`), label, type, and reason. Flags `number`, legacy `text`/`address`/`amount`, `divider`, `unknown`, and `list_layout`-as-a-standalone-field; recurses through `preview_layout` propagating `ContainsUnsupportedNestedFields`.
- `examples/check_anchorage_render.rs` — `cargo run -p visualsign --features diagnostics --example check_anchorage_render -- <payload.json>`.
- `docs/specs/anchorage-wallet-render-rules.md` — the authoritative ruleset (source of truth: the Anchorage wallet's Visual Signing Protocol engine).

Draft — opening for visibility; a natural follow-up is wiring it into CI alongside `validate_charset`.

### DIFF
--- a/docs/specs/anchorage-wallet-render-rules.md
+++ b/docs/specs/anchorage-wallet-render-rules.md
@@ -33,10 +33,10 @@ other string decodes to unknown → unsupported.
 | `delta`          | wallet-only | leaf    | wallet renders it, but **not in the validator's allowed set** (parser emits none; no plans) |
 | `highlight`      | wallet-only | leaf    | wallet renders it, but **not in the validator's allowed set** (parser emits none; no plans) |
 | `rule`           | wallet-only | leaf    | wallet renders it, but **not in the validator's allowed set** (parser emits none; no plans) |
-| `preview_layout` | yes      | container  | requires both `Condensed`/`Expanded` on the wire, and every descendant to render |
+| `preview_layout` | yes      | container  | every descendant must render                       |
 | `accordion`      | wallet-only | container | wallet renders it, but **not in the validator's allowed set** (parser emits none; no plans) |
 | `list_layout`    | **structural only** | — | valid only as a container's `Condensed`/`Expanded`; **never as a field entry** |
-| `number`         | **no**   | —          | not a VSP type → use `amount_v2`                   |
+| `number`         | yes, as `amount_v2` | leaf | not a VSP type on its own; the in-memory `Number` variant serializes to `amount_v2` on the wire (see #393), so it renders fine |
 | `text` (v1)      | **no**   | —          | superseded by `text_v2`                            |
 | `address` (v1)   | **no**   | —          | superseded by `address_v2`                         |
 | `amount` (v1)    | **no**   | —          | superseded by `amount_v2`                          |
@@ -64,19 +64,20 @@ when a parser field variant starts emitting it.
    containing unsupported nested fields. Nesting of supported leaves is unlimited in
    depth; nesting via `preview_layout` is the supported way to express hierarchy.
 
-3. **`preview_layout` requires both `Condensed` and `Expanded` on the wire.**
-   The wallet's model has no optional case for either list; a missing key fails
-   to decode and the whole container falls back to plain text. The validator
-   checks the actual serialized output rather than the in-memory value, since a
-   builder leaving a list unset and the serializer's choice of how to represent
-   that (omit the key vs. default to an empty list) are independent concerns.
-   `create_preview_layout` always leaves `Condensed` unset today, and several
-   Ethereum visualizers (e.g. the ERC-20 `Transfer` preview) leave `Expanded`
-   unset too, so this is a live parser output shape, not a hypothetical one.
+3. **`preview_layout` always carries both `Condensed` and `Expanded` on the
+   wire.** The `SignablePayloadFieldPreviewLayout` serializer emits an empty
+   list for whichever is unset (see #403), so the wallet's decoder always
+   receives both — there is no missing-list failure mode for the validator to
+   catch. `create_preview_layout` always leaves `Condensed` unset in memory,
+   and several Ethereum visualizers (e.g. the ERC-20 `Transfer` preview) leave
+   `Expanded` unset too; both cases render clean.
 
-4. **Numeric values use `amount_v2`, never `number`.** `number` is a legacy
-   parser-only variant that the wallet never recognized; render numbers (token
-   amounts, slippage bps, fees, compute units) as `amount_v2`.
+4. **`number` fields render fine, via `amount_v2` on the wire.** VSP has no
+   `number` type; the in-memory `Number` variant is serialized to `amount_v2`
+   (see #393), so it is not flagged as unsupported. New code should still
+   prefer building `amount_v2` fields directly (via `create_amount_field`)
+   rather than `create_number_field`, since the remap is a compatibility
+   shim, not the primary path.
 
 ## Validator
 
@@ -95,8 +96,6 @@ when a parser field variant starts emitting it.
   since a variant can be remapped to a different wire representation.
 - `ListLayoutAsStandaloneField` — `list_layout` used as a field entry.
 - `ContainsUnsupportedNestedFields` — a container with an unsupported descendant.
-- `MissingRequiredList` — a `preview_layout` missing `Condensed` or `Expanded`
-  on the wire.
 
 Ad-hoc check of a payload JSON:
 
@@ -107,8 +106,7 @@ cargo run -p visualsign --features diagnostics \
 
 ## Known parser violations (as of this writing)
 
-`create_number_field` emits the unsupported `number` type, used by the
-`compute_budget`, `system`, `token_2022`, and `jupiter_swap` presets. Payloads
-containing those instructions carry unrenderable fields today (e.g. a DFlow swap
-transaction's leading ComputeBudget instructions). Migrating those call sites to
-`amount_v2` is tracked separately.
+None. `create_number_field` (used by the `compute_budget`, `system`,
+`token_2022`, and `jupiter_swap` presets) previously emitted the unrenderable
+`number` type; #393 made the `Number` variant serialize as `amount_v2` on the
+wire, so those call sites render fine without needing a migration.
--- a/docs/specs/anchorage-wallet-render-rules.md
+++ b/docs/specs/anchorage-wallet-render-rules.md
@@ -1,0 +1,114 @@
+# Anchorage Wallet Render Validation Rules
+
+Status: draft · Owner: parser team
+
+## Purpose
+
+The parser's output is consumed by the **Anchorage wallet**, whose Visual
+Signing Protocol engine renders each `SignablePayload` field according to a
+fixed support matrix. Any field the wallet does not recognize is surfaced to the
+user as an **unsupported field** (and nested containers are flagged as
+containing unsupported nested fields). Because VSP is WYSIWYS ("what you see is what
+you sign"), an unrenderable field is a **correctness bug**, not a cosmetic one.
+
+These rules let the parser validate its own output against the wallet's support
+matrix *before* shipping, the same way `validate_charset` guards the character
+set.
+
+Source of truth: the Anchorage wallet's Visual Signing Protocol engine (its
+field-type decoder, preview-layout and accordion parsing, and field-diagnostic
+model). Mirror this spec to the wallet whenever that engine changes.
+
+## Field support matrix
+
+The wallet's field-type decoder recognizes exactly these `Type` strings; every
+other string decodes to unknown → unsupported.
+
+| `Type`           | Renders? | Kind       | Notes                                              |
+|------------------|----------|------------|----------------------------------------------------|
+| `text_v2`        | yes      | leaf       |                                                    |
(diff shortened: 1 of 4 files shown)


src/visualsign/examples/check_anchorage_render.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use std::fs;
+
+use visualsign::SignablePayload;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let path = std::env::args()
+        .nth(1)
+        .ok_or("usage: check_anchorage_render <payload.json>")?;
+    let json = fs::read_to_string(&path)?;
```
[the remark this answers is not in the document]
prasanna-anchorage: This compiles as-is: `?` converts the `&'static str` error into `Box<dyn Error>` via std's `From<&str>` impl (verified with `cargo build --example check_anchorage_render`). No change needed.

src/visualsign/src/anchorage_render.rs
```diff
@@ -0,0 +1,378 @@
...
+//! | `text` (v1)      | no                 | superseded by `text_v2`                 |
+//! | `address` (v1)   | no                 | superseded by `address_v2`              |
+//! | `amount` (v1)    | no                 | superseded by `amount_v2`               |
+//! | `divider`        | no                 | not in the wallet decoder               |
+//! | `unknown`        | no                 | explicit fallback/unsupported           |
+//!
+//! The wallet additionally renders `delta`, `highlight`, `rule`, and `accordion`
+//! containers, but the parser has no field variant that emits them, so they
+//! cannot occur in parser output. They are listed in
+//! [`ANCHORAGE_RENDERABLE_LEAF_TYPES`] for completeness / forward compatibility.
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in caaa15cb: reworded so delta/highlight/rule are described as leaf types and only accordion as a container, consistent with ANCHORAGE_RENDERABLE_LEAF_TYPES.

src/visualsign/src/anchorage_render.rs
```diff
@@ -0,0 +1,375 @@
...
+            out.push(finding(
+                path,
+                field,
+                AnchorageUnsupportedReason::ListLayoutAsStandaloneField,
+            ));
+            false
+        }
+        // Every other variant is a leaf: renderable iff its type is in the set.
+        _ => {
+            if ANCHORAGE_RENDERABLE_LEAF_TYPES.contains(&field.field_type()) {
```
pepe-anchor: should: `check_field`'s catch-all arm checks `field.field_type()`, a hardcoded per-variant literal in `lib.rs`, instead of the actual `"Type"` string `serialize_to_map()` puts on the wire. These are two independent match statements over the same enum that happen to agree today. Sibling PR #393 (open, same author, targets this PR's own documented `number` gap) changes only `serialize_to_map`'s `Number` arm to emit `amount_v2` on the wire while leaving `field_type()` returning `number`. Once #393 merges, every `Number` field will render fine in the wallet but this validator will flag it as unsupported forever, a permanent false positive baked in before the PR's own stated CI-wiring follow-up lands.

```suggestion
Derive the checked type string from `serialize_to_map()`'s `Type` entry (or a shared single source of truth) instead of `field_type()`, so the validator tracks whatever actually reaches the wallet on the wire.
```

src/visualsign/src/anchorage_render.rs
```diff
@@ -0,0 +1,375 @@
...
+                }
+            }
+            if let Some(expanded) = &preview_layout.expanded {
+                for (index, child) in expanded.fields.iter().enumerate() {
+                    let ok = check_field(
+                        &child.signable_payload_field,
+                        &format!("{path}.Expanded.Fields[{index}]"),
+                        out,
+                    );
+                    all_ok = all_ok && ok;
```
pepe-anchor: should: When `preview_layout.condensed` or `.expanded` is `None`, the `if let Some` branch is skipped: no descendants are checked and no finding is pushed. This is live today: `field_builders::create_preview_layout` always sets `condensed: None`, and multiple Ethereum visualizers (`erc20.rs`, `universal_router.rs`, `permit2.rs`, `erc1155.rs`) set `condensed`/`expanded` to `None`, e.g. the ERC-20 `Transfer` preview. `lib.rs` still has `skip_serializing_if = "Option::is_none"` on both fields on this PR's base, and open PR #403 documents that this exact shape drops the key from the JSON and throws `keyNotFound` on iOS, degrading the whole preview_layout to FallbackText in production (Sentry IOS-646/PRS-548). The validator calls this construct clean today even though it is a known, previously-shipped WYSIWYS decode failure; it becomes correct only once #403 merges.

```suggestion
PR #403 (https://github.com/anchorageoss/visualsign-parser/pull/403) fixes the serialization gap (always emitting Condensed/Expanded), but this validator should be fixed independently rather than relying on merge order: add a test that pins the None-handling behavior explicitly (e.g. assert a None condensed/expanded is not silently reported as clean, or check against the actual serialized wire output instead of the in-memory Option), so this code stays correct on its own regardless of whether #403 has landed.
```

src/visualsign/examples/check_anchorage_render.rs
```diff
@@ -0,0 +1,32 @@
+//! Report the fields an Anchorage wallet cannot render in a SignablePayload JSON.
+//!
+//! Usage:
+//!   cargo run -p visualsign --features diagnostics \
+//!     --example check_anchorage_render -- <payload.json>
```
pepe-anchor, line 5: nit: The example builds and runs fine without `--features diagnostics` for ordinary payloads, but a payload containing `"Type": "diagnostic"` (one of the four renderable leaf types) fails to deserialize at all without the feature: `unknown variant 'diagnostic'`. The documented flag is real, not decorative, but forgetting it surfaces a generic serde error instead of a hint.

```suggestion
Catch the deserialize error and print a hint to pass --features diagnostics, or default-enable the feature for this example.
```

src/visualsign/src/anchorage_render.rs
prasanna-anchorage: Confirmed — checked current `main`: PR #393 (merged as 52eda7f4) makes `Number` serialize as `"Type": "amount_v2"` on the wire while `field_type()` still returns `"number"`. Fixed by deriving the checked type from `serialize_to_map()`'s actual `Type` entry (new `wire_type()` helper) instead of `field_type()`, and reused it for the `field_type` reported in findings. This tracks whatever reaches the wallet regardless of future remaps.

src/visualsign/src/anchorage_render.rs
prasanna-anchorage: Confirmed — `create_preview_layout` always leaves `condensed: None`, and several `erc20.rs` call sites leave `expanded: None` too, so this is a live parser output shape today, not hypothetical. Took the second option: added `wire_has_key()` to check the actual serialized `Condensed`/`Expanded` keys rather than the in-memory `Option`, plus a new `MissingRequiredList` reason and two regression tests (missing one list, missing both). Checking the wire output rather than hardcoding "`None` = broken" means this stays correct on its own once #403 changes the serializer to always emit both keys — no merge-order dependency either way.

src/visualsign/examples/check_anchorage_render.rs
prasanna-anchorage: Fixed — the deserialize now maps to an error that includes a hint about `--features diagnostics` when parsing fails, instead of surfacing the bare serde "unknown variant" error.

src/visualsign/src/anchorage_render.rs
prasanna-anchorage: Update after rebasing onto main: #403 merged, and its `SignablePayloadFieldPreviewLayout` serializer always emits both `Condensed` and `Expanded` (empty list when unset) — so the missing-list failure mode this thread flagged is fixed at the serialization layer, not just documented as pending. That makes `wire_has_key()`/`MissingRequiredList` unreachable: `wire_has_key` can never return false once #403 is in the tree. Removed both, and replaced the two `MissingRequiredList` tests with `preview_layout_with_unset_lists_renders_clean` / `preview_layout_with_both_lists_unset_renders_clean`, which pin that a `None` condensed/expanded (as `create_preview_layout` and several Ethereum visualizers produce) renders clean now that the wire always carries both keys. Doc matrix and spec updated to match — no missing-list case is documented anymore.

src/visualsign/src/anchorage_render.rs
prasanna-anchorage: Update after rebasing onto main: #393 is merged now, and CI confirmed the exact failure mode you called out — `number_field_is_unsupported` failed post-rebase because `wire_type()` correctly returns `amount_v2` for a `Number` field once #393's `serialize_to_map` remap is in the tree. Replaced that test with `number_field_renders_as_amount_v2_on_the_wire`, a regression test asserting `Number` findings are empty and `wire_type()` returns `amount_v2`. Switched the other tests' "unsupported leaf" fixture from `number` to `divider`, which isn't remapped to a renderable type, so it stays a stable unsupported example independent of future wire remaps.

src/visualsign/src/anchorage_render.rs
```diff
@@ -0,0 +1,458 @@
...
+            ));
+            false
+        }
+        // Every other variant is a leaf: renderable iff its wire `Type` is in
+        // the set. Checked against the wire type, not `field_type()`: a
+        // variant's wire representation can be remapped (e.g. serialized
+        // under a different `Type`) without changing what `field_type()`
+        // reports, and this must track what actually reaches the wallet.
+        _ => {
+            if ANCHORAGE_RENDERABLE_LEAF_TYPES.contains(&wire_type(field).as_str()) {
```
shahan-khatchadourian-anchorage, line 182: nit, non-blocking: `wire_type(field)` gets called twice per unsupported leaf — once here to check set membership, and again inside `finding()` (line 218) to populate the report — each call does a full `serialize_to_map()` of the field. Could pass the already-computed string into `finding()` instead of recomputing it, since this path is hit for every field in the tree.

docs/specs/anchorage-wallet-render-rules.md
```diff
@@ -0,0 +1,112 @@
...
+   renderable. A single unsupported descendant flags the whole container as
+   containing unsupported nested fields. Nesting of supported leaves is unlimited in
+   depth; nesting via `preview_layout` is the supported way to express hierarchy.
+
+3. **`preview_layout` always carries both `Condensed` and `Expanded` on the
+   wire.** The `SignablePayloadFieldPreviewLayout` serializer emits an empty
+   list for whichever is unset (see #403), so the wallet's decoder always
+   receives both — there is no missing-list failure mode for the validator to
+   catch. `create_preview_layout` always leaves `Condensed` unset in memory,
+   and several Ethereum visualizers (e.g. the ERC-20 `Transfer` preview) leave
```
shahan-khatchadourian-anchorage, line 72: nit, non-blocking: this says the ERC-20 `Transfer` preview "leave[s] `Expanded` unset too" — `Transfer` actually only leaves `Condensed` unset; it's the read-only query calls (Name/Symbol/Decimals/TotalSupply) that leave both unset. Doesn't matter functionally anymore since #403 makes both cases render clean either way, but the citation is slightly off.

src/visualsign/examples/check_anchorage_render.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::fs;
+
+use visualsign::SignablePayload;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let path = std::env::args()
+        .nth(1)
+        .ok_or("usage: check_anchorage_render <payload.json>")?;
+    let json = fs::read_to_string(&path)?;
+    let payload: SignablePayload = serde_json::from_str(&json).map_err(|e| {
```
shahan-khatchadourian-anchorage, line 16: nit, non-blocking: this hint is appended to *any* parse failure, including ones unrelated to the diagnostics feature (a plain JSON syntax error, or an unknown `Type` string that isn't `diagnostic`). Could make it conditional on the error actually being an unrecognized-variant/diagnostic-shaped failure so it doesn't send someone down the wrong path for an unrelated JSON error.

shahan-khatchadourian-anchorage (approved):
Approving — the two substantive issues on this thread (the #393 `number`/`amount_v2` wire-type remap and the #403 missing-`Condensed`/`Expanded` fix) are both properly resolved at the root cause (serializer-level fix for the latter, which is better than just documenting it), the now-dead `MissingRequiredList`/`wire_has_key` machinery was cleanly removed, and all 9 tests + clippy pass locally.

Did a second pass for anything not already covered. Nothing blocking — three small nits below, all non-blocking.

PR #395, opened 2026-06-23
feat(tools): standalone tvc-deploy helper + manual deploy workflow

prasanna-anchorage:
## Summary
A standalone Rust helper + manual workflow to perform dev/staging `parser_app` TVC deploys, plus a post-deploy smoke check.

### `tools/tvc-deploy/` — standalone helper (xshell + anyhow + lexopt; own workspace, no `visualsign` deps)
- `gen-operator-key --out <path>`: mint a `qos_p256` operator key (seed → 0600 file, only the public key printed).
- `deploy …`: **image-derived digest gate** — extract `/parser_app` from the image, `sha256` it, and assert it equals `--expected-digest` **before** anything is submitted (ties the deployed digest to the actual binary) — then assemble `tvc-deploy.json` (gRPC health), `tvc deploy create` → approve → poll-to-healthy → `set-live`.

### `.github/workflows/tvc-deploy.yml` — manual + label-gated deploy
- `workflow_dispatch` (explicit inputs) and a `tvc-deploy-test`-labeled `pull_request` path (test app via `vars.TVC_TEST_*`). Pins `tvc --version 0.7.0 --locked`; minimal permissions.
- **Smoke step** (post set-live): runs the published `turnkey-client` container `parse --dev-path` against `/visualsign-dev` and asserts the V0+ALT fixture renders (regression guard for "Cannot render V0"), reusing `TVC_API_KEY_*`; outage abort-guard. Requires the turnkey-client image on GHCR.

Validated: helper proven end-to-end on a throwaway dev app and via a green labeled CI run against the test app `80b6f48f`; smoke logic validated live (35,177 chars rendered).


prasanna-anchorage:
Addressed the review feedback in `6c4aef8d`:

- **CodeQL (missing permissions):** added a minimal `permissions: { contents: read }` to the workflow.
- **`cargo install` reproducibility (+ pin request):** now `cargo install tvc --version 0.6.2 --locked`.
- **Operator-seed temp file could leak on early error:** moved config assembly/write inside the cleanup closure, so the seed + config temp files are always removed regardless of where the flow fails.
- **`path()` lossy non-UTF-8 → empty arg:** removed entirely — the helper was reorganized onto `xshell` (`cmd!`), which takes `Path`/`PathBuf` directly, so there's no `to_str()` fallback path anymore.

Also reorganized the helper onto the idiomatic shell-orchestration stack (`xshell` + `anyhow` + `lexopt`, all small/fast-compiling) and replaced the fragile `approve --dry-run` manifest parse with an **image-derived digest gate** (extract `/parser_app`, sha256, assert == `--expected-digest`) — stronger, ties the digest to the actual binary.

andrestielau (approved):

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,403 @@
...
+        .cloned()
+        .unwrap_or_else(|| "0.0.0.0".to_string());
+    let host_port: u16 = match flags.get("host-port") {
+        Some(s) => s.parse().map_err(|_| format!("invalid --host-port: {s}"))?,
+        None => 3000,
+    };
+
+    validate_digest(&digest)?;
+    let (seed_path, cleanup_seed) = resolve_seed_file(flags)?;
+
```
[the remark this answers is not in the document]
prasanna-anchorage: Addressed: `deploy()` now wraps everything after the seed file is created in a closure (`main.rs:150`), and the seed/config cleanup runs unconditionally before the error is propagated (`outcome?`), so an early failure no longer leaves the operator seed on disk.

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,403 @@
...
+    let nanos = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_nanos())
+        .unwrap_or(0);
+    std::env::temp_dir().join(format!("{prefix}-{}-{nanos}.{ext}", std::process::id()))
+}
+
+fn path(p: &Path) -> &str {
+    p.to_str().unwrap_or_default()
+}
```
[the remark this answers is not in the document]
prasanna-anchorage: No lossy conversion exists in the current code — paths are never stringified. `cmd!` interpolates `PathBuf`/`Path` directly as `OsStr` (`main.rs:167`, `174`, `208`, `211`), so a non-UTF-8 path would be passed through intact, not emptied.

.github/workflows/tvc-deploy.yml
```diff
@@ -0,0 +1,94 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@fb51252c7ba57d633bc668f941da052e410add48 # v1.13.0
+
+      - name: Install tvc CLI
+        run: cargo install tvc
+
```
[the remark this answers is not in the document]
prasanna-anchorage: Done — `cargo install tvc --version 0.7.0 --locked` (`.github/workflows/tvc-deploy.yml:78`).

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    let mut f = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .mode(0o600)
+        .open(path)
+        .with_context(|| format!("open {}", path.display()))?;
+    f.write_all(contents.as_bytes())
+        .with_context(|| format!("write {}", path.display()))
+}
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in ed864cae — `write_secret_file` now calls `set_permissions(Permissions::from_mode(0o600))` after open (`main.rs:113`), so a pre-existing broader-perm file is forced back to 0600.

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    let _ = std::fs::remove_file(&bin);
+    let actual = sha.split_whitespace().next().unwrap_or_default();
+    if !actual.eq_ignore_ascii_case(expected) {
+        bail!(
+            "DIGEST GATE FAILED: image /parser_app sha256 {actual} != expected {expected}; refusing to deploy"
+        );
+    }
+    println!("digest gate passed: image /parser_app sha256 == {expected}");
+    Ok(())
+}
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in ed864cae — the extract+hash runs in a closure and both the temp binary and the container are removed unconditionally on every path before the error is propagated (`main.rs:207-216`).

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    })
+}
+
+fn temp_path(prefix: &str, ext: &str) -> PathBuf {
+    let nanos = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_nanos())
+        .unwrap_or(0);
+    std::env::temp_dir().join(format!("{prefix}-{}-{nanos}.{ext}", std::process::id()))
+}
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed in ed864cae — `temp_path` now appends a per-process `AtomicU64` counter (`main.rs:347-352`), so same-tick calls cannot collide.

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+        let created = cmd!(sh, "tvc deploy create --config-file {cfg_path}")
+            .read()
+            .context("tvc deploy create")?;
+        let deploy_id = parse_after(&created, "Deployment ID:")
+            .with_context(|| format!("no deployment id in create output:\n{created}"))?;
+        println!("created deployment {deploy_id}");
+
+        cmd!(sh, "tvc deploy approve --deploy-id {deploy_id} --operator-id {operator_id} --operator-seed {seed_path} --dangerous-skip-interactive")
+            .run()
+            .context("tvc deploy approve")?;
```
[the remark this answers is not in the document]
prasanna-anchorage: This was written against an earlier PR description. The description now states an image-derived digest gate: extract `/parser_app` from the image, sha256 it, and assert it equals `--expected-digest` before anything is submitted (`verify_image_digest`, `main.rs:198-224`) — there is no `--dry-run` manifest read claimed anymore, so description and implementation match. The verified digest is also passed to `tvc deploy create` as `expectedPivotDigest` (`main.rs:158`), which is what TVC builds the manifest from. A `tvc deploy approve --dry-run` re-read would be defense-in-depth rather than a correctness fix, and is out of scope here.

.github/workflows/tvc-deploy.yml
```diff
@@ -0,0 +1,150 @@
...
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  deploy:
+    # Manual dispatch, or the `tvc-deploy-test` label on a PR (test app only).
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.label.name == 'tvc-deploy-test'
```
pepe-anchor, line 75: must: The `pull_request: types: [labeled]` trigger runs `tvc-deploy deploy`, which unconditionally calls `tvc deploy approve ... --dangerous-skip-interactive` right after create, i.e. a full create->approve->set-live with no human checkpoint, fed by `TVC_CI_OPERATOR_SEED`/`TVC_API_KEY_*` secrets, triggered by anyone able to label a same-repo PR `tvc-deploy-test`. This already happened for real on this PRs own branch (`feat/tvc-deploy-helper`): GitHub Actions run 28013160686 shows checkout, build, Deploy and Scrub all green from the PRs own head commit. Because `pull_request` (unlike `pull_request_target`) resolves both the checked-out code AND the workflow YAML from the PR head, a same-repo collaborator can edit `main.rs` or this workflow file in the same PR to exfiltrate the operator seed / API keys, and there is no `environment:` gate or required-reviewer step anywhere in the file to stop it. This directly contradicts the requesters own stated design goal in the linked Slack thread ("give a role to Github thats only allowed to initiate a deploy, it can never approve, only humans should approve") and runs ahead of the phased plan (PRS-516/519/520/521), whose Enabler ticket (PRS-521, still Backlog) is supposed to provision scoped CI keys + a Turnkey policy before any auto-approve happens from CI.

scripts/smoke.sh
```diff
@@ -0,0 +1,169 @@
...
+    exit 1
+  fi
+  # A genuine transport/network error reaching the endpoint is a pre-existing
+  # outage, not the deploy's fault -> skip. Match only connection-level errors.
+  if grep -qiE \
+    'connection refused|connection reset|no such host|dial tcp|i/o timeout|timeout|tls handshake|\bEOF\b|context deadline|network is unreachable|server misbehaving|temporary failure in name resolution' \
+    "$ERRFILE"; then
+    echo "SKIP: dev endpoint unreachable / outage — not a regression" >&2
+    exit 0
+  fi
```
pepe-anchor, line 143: must: The outage-vs-regression classifier greps stderr for generic terms including bare `timeout`, `context deadline`, and `\bEOF\b`, and treats any match as a pre-existing outage (`exit 0`, reported as a pass/skip). A genuinely broken deployment, e.g. an enclave that hangs on the gRPC health check or crashes mid-response, is very likely to surface as exactly `context deadline exceeded` or an EOF from the clients perspective. Since this step runs immediately after `set-live` specifically to catch a broken deploy, a deploy broken in a way that manifests as a timeout or dropped connection is misclassified as an unrelated outage and CI reports smoke PASS, defeating the checks purpose at the moment it matters most.

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+        let transient = msg.contains("not yet available")
+            || msg.contains("try again")
+            || msg.contains("not found")
+            || msg.contains("zero healthy replicas");
+        if transient && start.elapsed() < timeout {
+            sleep(POLL_INTERVAL);
+            continue;
+        }
+        bail!("set-live failed: {}", msg.trim());
```
pepe-anchor, line 274: should: `set_live` lowercases combined stdout+stderr from `tvc app set-live-deploy` and treats a bare substring match on "already" (line 259) as success, regardless of the actual failure mode. If `tvc` ever emits an unrelated error containing "already" (e.g. a retry-exhaustion message), this would silently report success on a failed deploy in a fully automated, no-human-in-the-loop path. No current evidence this string appears in a real failure message today, so this is a latent risk rather than a proven break.

.github/workflows/tvc-deploy.yml
```diff
@@ -0,0 +1,150 @@
...
+        default: "latest"
+  pull_request:
+    # Apply the `tvc-deploy-test` label to a PR to run a CI validation deploy
+    # against the dev TEST app (uses repo vars.TVC_TEST_*; never the live app).
+    types: [labeled]
+
+# Avoid overlapping deploys to the same target.
+concurrency:
+  group: tvc-deploy-${{ github.ref }}
+  cancel-in-progress: false
```
pepe-anchor, line 63: should: For the PR-label path, `github.ref` is per-PR (`refs/pull/<n>/merge`), so two different PRs both labeled `tvc-deploy-test` at the same time land in different concurrency groups and can race each others create/approve/set-live against the same `vars.TVC_TEST_APP_ID`. Same issue between two `workflow_dispatch` runs targeting the same `app_id` from different branches. Blast radius is limited to the dev/test app today, not prod.

```suggestion
Key the group on the resolved app id instead, e.g. `tvc-deploy-${{ inputs.app_id || vars.TVC_TEST_APP_ID }}`.
```

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,375 @@
...
+        if let Some(rest) = t.strip_prefix("Deployment:") {
+            in_block = rest.trim() == deploy_id;
+        } else if in_block {
+            if let Some(rest) = t.strip_prefix("Healthy / Desired Replicas:") {
+                return rest.split_whitespace().next().map(str::to_owned);
+            }
+        }
+    }
+    None
+}
```
pepe-anchor, line 327: should: These are pure functions scraping freeform `tvc` CLI output (e.g. `Healthy / Desired Replicas: X/Y`) with zero test coverage, and the workflow only runs `cargo build --release`, never `cargo test`. If `tvc` ever reformats that text, `deployment_health` silently stops matching and `poll_health` spins until timeout with an "unknown" status, with nothing catching the regression before a real deploy run.

scripts/smoke.sh
prasanna-anchorage: Fixed in c1fb6bef: the outage classifier no longer treats bare `timeout`/`EOF`/`context deadline` as a benign outage-skip. It now only skips on unambiguous never-connected errors (connection refused/reset, no such host, dial tcp, tls handshake, network unreachable, etc.); timeout/EOF/context-deadline on an otherwise-reachable endpoint now falls through to an explicit `FAIL` (exit 1) instead of `SKIP`, since those are exactly what a hung or crashed enclave would surface as right after set-live.

tools/tvc-deploy/src/main.rs
prasanna-anchorage: Fixed in c1fb6bef — `set_live` now requires both `"already"` and `"live"` in the combined stdout+stderr before treating it as success, so an unrelated failure that happens to contain "already" (e.g. a retry-exhaustion message) can no longer be misreported as a live deploy.

.github/workflows/tvc-deploy.yml
prasanna-anchorage: Fixed in c1fb6bef — took the suggested fix, concurrency group is now keyed on `inputs.app_id || vars.TVC_TEST_APP_ID` instead of `github.ref`.

tools/tvc-deploy/src/main.rs
prasanna-anchorage: Fixed in c1fb6bef — added a `#[cfg(test)]` module covering `deployment_health`, `validate_digest`, and `parse_after` (8 unit tests), and added a `cargo test` step to the workflow before the build step (previously it only ran `cargo build`), so a regression in the freeform-output parsing now fails CI instead of silently spinning to an "unknown" status.

.github/workflows/tvc-deploy.yml
prasanna-anchorage: Agreed, and this is the exact concern #404 is designed around: splitting `deploy` into separate `initiate` / `approve` / `promote` subcommands (behind a `TvcOps` trait, with unit tests) so CI can be scoped to initiate-only and a human/operator key does the approve. That said, #404 currently only has the `main.rs` split + design docs landed — the workflow-level fix (retiring this labeled-PR auto-approve path in favor of the planned `Release`/`Promote` workflows, Phase 4) hasn't shipped, and PRS-521 (scoped CI keys + Turnkey policy) is still Backlog. So this stays an open risk on the actual `tvc-deploy-test` label path in this PR until that phase lands. Tracking there rather than duplicating the fix here — leaving this thread open until Phase 4 ships or we explicitly decide to gate/disable the labeled-PR path in the interim.

scripts/smoke.sh
```diff
@@ -0,0 +1,181 @@
...
+if [ -n "${TURNKEY_CLIENT:-}" ]; then
+  CLIENT="$TURNKEY_CLIENT"
+elif docker image inspect "$IMAGE" >/dev/null 2>&1 || docker pull "$IMAGE" >/dev/null 2>&1; then
+  CLIENT="$CONTAINER_CLIENT"
+elif [ -n "$CLIENT_PATH" ]; then
+  echo "container image $IMAGE unavailable; using local fallback client: $CLIENT_PATH" >&2
+  CLIENT="$(resolve_fallback_client "$CLIENT_PATH")" || exit $?
+else
+  CLIENT="$CONTAINER_CLIENT"
+fi
```
shahan-khatchadourian-anchorage, line 108: Minor, non-blocking: this fallback still has a gap in the "harness failure vs. endpoint outage" distinction the comment above (lines 96-98) says it maintains.

If `docker image inspect`/`docker pull` both fail (e.g. GHCR unreachable from the runner) and no `--turnkey-client-path` is given, we fall through to `CLIENT="$CONTAINER_CLIENT"` here and try it anyway at line 116. `docker run` will attempt its own implicit pull, hit the same network failure, and emit text like `dial tcp: lookup ghcr.io: ...` or `temporary failure in name resolution` into `$ERRFILE` — which matches the SKIP regex at line 139, not the "harness couldn't run" path. So a CI-runner-side GHCR outage gets reported as `SKIP: dev endpoint unreachable` (exit 0) instead of the `exit 2` harness failure this comment says it should be, and a real deploy regression underneath it wouldn't get caught.

Would be worth either treating a failed pull explicitly (bail with exit 2 before ever invoking `$CLIENT`) or having the SKIP/FAIL classifier distinguish "docker itself couldn't run" from "the target endpoint refused/timed out."

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,456 @@
...
+    // pre-existed with broader perms, so the secret is never world-readable.
+    f.set_permissions(Permissions::from_mode(0o600))
+        .with_context(|| format!("chmod {}", path.display()))?;
+    f.write_all(contents.as_bytes())
+        .with_context(|| format!("write {}", path.display()))
+}
+
+fn deploy(sh: &Shell, flags: &HashMap<String, String>) -> Result<()> {
+    let app_id = req(flags, "app-id")?;
+    let image = req(flags, "image-url")?;
```
shahan-khatchadourian-anchorage, line 125: Minor, non-blocking: `--image-url` isn't validated as digest-pinned (e.g. `...@sha256:...`), only `--expected-digest`'s hex format is checked (`validate_digest`, line 329). The workflow's own doc comment asks callers to pass a pinned reference, but nothing here enforces it.

If a mutable tag is passed, `verify_image_digest` (line 211) hashes whatever that tag resolves to *right now* and the gate passes — but the same tag string is also written into `pivotContainerImageUrl` for the TVC deploy config, so if TVC ever re-resolves that tag later (redeploy, restart) it could pull a different image than the one actually verified. A cheap `image.contains("@sha256:")` check before the digest gate would close this without much code.

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,456 @@
...
+            String::from_utf8_lossy(&out.stderr)
+        )
+        .to_lowercase();
+        if msg.contains("already") && msg.contains("live") {
+            println!("{deploy_id} already live (auto-targeted)");
+            return Ok(());
+        }
+        let transient = msg.contains("not yet available")
+            || msg.contains("try again")
+            || msg.contains("not found")
```
shahan-khatchadourian-anchorage, line 268: Minor, non-blocking: `msg.contains("not found")` in the transient-retry set is fairly broad — it'd also match a permanent error like a typo'd `--operator-id` or `--deploy-id` (e.g. "operator not found in manifest set"), causing `set_live` to retry every 15s for the full 300s timeout before bailing with a generic message, instead of failing fast with something actionable.

Separately (very minor): the loop checks `start.elapsed() < timeout` *before* sleeping (line 270) rather than after, so on a transient message near the deadline it can overshoot `SETLIVE_TIMEOUT` by up to one `POLL_INTERVAL` (~15s) — `poll_health`'s loop checks elapsed right before bailing and doesn't have this gap.

tools/tvc-deploy/src/main.rs
```diff
@@ -0,0 +1,456 @@
...
+    let _ = std::fs::remove_file(&cfg_path);
+
+    let deploy_id = outcome?;
+    println!("deployment {deploy_id} is healthy and live");
+    Ok(())
+}
+
+/// Extract `/parser_app` from the image and sha256 it; it MUST equal the
+/// submitted `--expected-digest`. Ties the deployed digest to the real binary.
+fn verify_image_digest(sh: &Shell, image: &str, expected: &str) -> Result<()> {
```
shahan-khatchadourian-anchorage, line 211: Minor, non-blocking, efficiency nit: `verify_image_digest` spawns 4 subprocesses + a host temp file (`docker create` → `docker cp` → `sha256sum` → `docker rm`) to hash one file already inside the image. Since `parser_app`'s final image stage is `stagex/core-busybox` (which ships `sha256sum`), `docker run --rm --entrypoint sha256sum {image} /parser_app` would get the identical guarantee (still hashing the real image content, not a rebuilt binary) in one subprocess, with `--rm` handling cleanup and no host temp file needed at all.

shahan-khatchadourian-anchorage (approved):
Nice, thorough hardening across the review history here — the digest gate, seed scrubbing, and the recent set-live/concurrency/test-coverage fixes all look solid, and I independently re-verified `cargo build`/`cargo test`/`cargo clippy` all pass clean for `tools/tvc-deploy`.

Ran a second pass looking for anything not already covered in this thread. The one structural gap (no human-approval checkpoint on the label-triggered path) is already tracked as a `must` above and deferred to #404/PRS-521, so not re-raising it. Found a few smaller things below — none of these block merge, flagging as fast-follow candidates.

PR #400, opened 2026-06-25
draft: feat(skills): dispatch solana-add-idl scaffolding as Sonnet subagent

### LINKED ISSUE #397: feat(skills): set a fixed default model for solana-add-idl
## Problem

`solana-add-idl` currently inherits the user's active session model, so the model it runs on is unpredictable. The skill should have a fixed default so its behavior is consistent regardless of what model the user happens to have selected.

## Task

Decide which model the skill should target, then wire it in via the `model:` frontmatter field.

The skill does formulaic but Rust-specific scaffolding — it needs to get import lists, API shapes, and naming conventions exactly right. The candidates are `sonnet` and `haiku`; `sonnet` is the safer starting point.

**To validate a candidate model:**
- Run the skill end-to-end against at least one real program ID
- Confirm generated `config.rs`, `mod.rs`, and test stubs match the expected patterns from the codebase
- Check that Rust-specific details are correct: `BTreeMap` not `HashMap`, wire-data context API (`context.resolve_program_id()` / `context.resolve_accounts()` / `context.data()`), imports at module top not inside functions

## Acceptance criteria

- [ ] Candidate model tested end-to-end on a real program
- [ ] `model: <chosen>` added to `SKILL.md` frontmatter
- [ ] Skill output reviewed for correctness of Rust patterns

shahan-khatchadourian-anchorage:
## Summary

- Restructures `solana-add-idl` into two phases so the model is hard-enforced at the API layer rather than advisory
- Orchestrator phase (session model): gathers program address, name, and VisualizerKind interactively from the user
- Subagent phase (`model: "sonnet"` in the Agent dispatch call): fetches IDL, scaffolds all files, registers the preset, and verifies

The `model:` frontmatter field is advisory only — the subagent dispatch with an explicit model parameter is the only mechanism that actually enforces a specific model independent of the user's active session model.

Closes #397

## Test plan

- [ ] Invoke `/solana-add-idl` and confirm scaffolding subagent runs on Sonnet
- [ ] Confirm generated `config.rs`, `mod.rs`, and tests match expected codebase patterns
- [ ] Confirm both diagnostics feature flag configurations compile and pass

### DIFF
--- a/.claude/skills/solana-add-idl/SKILL.md
+++ b/.claude/skills/solana-add-idl/SKILL.md
@@ -8,6 +8,11 @@ user-invocable: true
 
 You are orchestrating the scaffolding of a new Solana program visualizer preset.
 
+## Mode Selection
+
+- **Standard** (`/solana-add-idl`): gather inputs, dispatch one Sonnet subagent to scaffold the preset
+- **Compare** (`/solana-add-idl compare`): gather inputs, dispatch Sonnet and Opus subagents in parallel to separate temp dirs, then diff their outputs to surface capability gaps
+
 ## Step 1: Gather Information
 
 Ask the user for:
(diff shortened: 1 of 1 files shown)


shahan-khatchadourian-anchorage:
Closing in favor of #401.

#401 carries forward everything this PR does to `solana-add-idl/SKILL.md` (orchestrator + Sonnet subagent dispatch, compare mode) and supersedes the arg-rendering approach here: instead of embedding the `format_arg_value` / `charset_safe` / `bytes_as_hex` bodies into the skill for each preset to copy, #401 extracts them into `crate::core::arg_rendering` and has the skill import from there ("do not copy the body"). It also does the corresponding refactor across the 18 existing presets, fixes the charset / field-explosion bugs, and validates the updated skill end-to-end by scaffolding the Marginfi preset with a Sonnet subagent — which satisfies the #397 acceptance criterion that this PR's test plan left unchecked.

Keeping both open would conflict on `SKILL.md` and re-introduce the per-preset helper duplication that #401 removes. #401 will close #397.

PR #409, opened 2026-07-03
Update QOS version to v2026.6.11

prasanna-anchorage:


shahan-khatchadourian-anchorage:
Closing as a duplicate of #408, which does the same QoS version bump to v2026.6.11 but also fills in the resolved qos commit (`3e4d73c4d2b9bb270c457bf7eae2e76a7f420998`) and CI run link in the marker/docs instead of leaving them as `<commit TODO>`.

PR #411, opened 2026-07-05
fix(scripts): extract-nitro-pcrs.sh buildx OCI export + --expected-kernel-hash

prasanna-anchorage:
## Summary
- Fix `scripts/extract-nitro-pcrs.sh` failing on hosts whose default `docker` buildx driver can't produce qos's `--output type=oci` build ("OCI exporter is not supported for the docker driver"). The script now provisions a scoped `docker-container` driver builder (`qos-oci-builder`) on demand and points the build at it via `BUILDX_BUILDER`, without touching the host's default builder or `daemon.json`.
- Add `--expected-kernel-hash <hex>` to assert the reproduced PCR0 and PCR1 both equal a given value (e.g. pulled from a live attestation), so this can be used to verify a deployment by rebuilding locally instead of eyeballing hex diffs. Exits non-zero on mismatch.

## Context
Related to PR #408, which bumps the `# qos-deployment-rev` marker for the `v2026.6.11` QoS release. That PR's chosen rev doesn't reproduce Turnkey's attested PCR0/PCR1 (verified independently) — this tool is what's needed to audit candidate revs against a real attestation value going forward.

## Test plan
- [x] `bash -n scripts/extract-nitro-pcrs.sh`
- [x] Ran the script end-to-end locally; confirmed it hits the buildx OCI-export error without the fix and builds successfully with it (rev `d866f2c6cbc58cc08c24eab4828f0824ad16a226`, PCR0/1 `d787eb65...22253ab`, matching the value documented in the original PR #300 commit message)
- [x] Unit-verified `check_expected_kernel_hash` against a sample `.pcrs` file: exits 0 on match, exits 1 with a clear mismatch message on mismatch

### DIFF
--- a/scripts/extract-nitro-pcrs.sh
+++ b/scripts/extract-nitro-pcrs.sh
@@ -14,6 +14,7 @@ SKOPEO_IMAGE="quay.io/skopeo/stable@sha256:4d60d6c00b62b463d0a99a7aeedc49358a32c
 QOS_REMOTE="https://github.com/tkhq/qos.git"
 QOS_DIR=""
 REV=""
+EXPECTED_KERNEL_HASH=""
 
 QOS_DIR_AUTO=0
 STAGE_DIR=""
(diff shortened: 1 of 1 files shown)


shahan-khatchadourian-anchorage (approved):
Validated manually succesfully against the git rev provided in #408

PR #413, opened 2026-07-14
feat(tools): Turnkey invite/policy management for tvc-deploy

prasanna-anchorage:
## Summary

Adds Turnkey org-admin subcommands to the standalone `tools/tvc-deploy` helper, built up incrementally to support batch-inviting a team and managing their access. Migrated to `clap` (matching the actual `tvc` CLI's own convention) partway through, replacing a hand-rolled lexopt+HashMap flag parser.

**Invitations**
- `invite` — single (`--user-name`/`--email`) or `--file` batch invite via one `create_invitations` activity (so a whole team needs only one consensus approval, if the org requires it)
- Before sending, fetches the org's current members and skips any invitee whose **canonical** email (lowercased, `+suffix` stripped, so `alice+dev1@co.com` and `alice@co.com` match) equals an existing member's — printing why, instead of silently re-inviting. Override per-invitee or file-wide with `invitees.json`'s own `"allowExisting"` field (not a CLI flag — bypass intent lives in the file you already curated); `--include-existing` disables the check entirely.
- `invitees.json` supports file-level defaults for `accessType`, `tags`, and `allowExisting`, each overridable per invitee.
- `dismiss-invite` — delete an invitation

**Activity consensus** (general Turnkey org consensus, distinct from `tvc deploy approve`'s QOS manifest signing)
- `approve-activity` / `reject-activity` — fetch an activity's fingerprint and vote on it

**Tags & users**
- `create-tag` — new user tag, optionally seeded with existing user ids
- `update-tag` — add/remove existing users from a tag, or rename it (needed since `invite`'s existing-member check means already-registered people can't pick up a new tag via invite — they have to be added directly)
- `list-tags` / `list-users` — look up tag ids and user ids by name/email

**Policies**
- `list-policies`, `create-policy` (single), `create-policies` (batch, from a template with `{{PLACEHOLDER}}` substitution via `--vars` + `--dry-run` to preview without submitting)
- `templates/releaser-initiator-policies.json` — mirrors the releaser/initiator policies already live in dev, parameterized by tag id for reuse in other environments
- `templates/readonly-policy.json` — a basic/read-only policy template: `EFFECT_DENY` on every activity type for a tagged user (deny always wins over conflicting allows), since Turnkey has no built-in "read-only" access level — only `accessType` (web/api/all, which controls login surface, not permissions) and policy-driven scoping

**CLI ergonomics**
- Migrated the whole CLI to `clap` derive: real `--help` at every level (was previously unimplemented), real boolean flags for `--dry-run`/`--include-existing`/`--allow-existing` (no more `--flag true`), and help text generated from doc comments instead of a hand-maintained `USAGE` string that had drifted out of sync with the code
- `--org` now accepts either the config alias or the org's raw UUID (falls back to matching `OrgEntry::id` if the alias lookup misses)
- `tools/tvc-deploy/README.md` documents the batch-invite and batch-policy workflows

Auth resolves the same way as the `tvc` CLI: `TVC_ORG_ID`/`TVC_API_KEY_PUBLIC`/`TVC_API_KEY_PRIVATE` env vars, else `~/.config/turnkey/tvc.config.toml` from `tvc login`.

## Review feedback addressed
- Copilot: stale module doc comment (only mentioned invite/dismiss-invite) → rewritten to cover the full subcommand surface
- Copilot: `TVC_API_BASE_URL` only takes effect alongside the other 3 env vars → kept (intentionally mirrors the official `tvc` CLI's own `client.rs`), documented why via comment
- Copilot: `--notes` shown as required in the old USAGE string but optional in code → resolved by the clap migration (shows `[default: ""]` in `--help`, matching the code)
- Copilot: `--help`/`-h` didn't work at all → resolved by the clap migration
- `/simplify` 4-angle review: extracted a shared `block_on` helper for the repeated tokio-runtime-per-subcommand boilerplate (11 call sites), and an `OrgArgs::as_deref()` convenience method to collapse `args.org.org.as_deref()` down to `args.org.as_deref()`. (Reviewed and intentionally skipped: joining `get_users`+`get_whoami` concurrently — real but marginal win not worth the added branching for a one-shot CLI; a small helper to dedupe the `ActivityRequiresApproval` println across 6 match arms — reviewer itself flagged this as low-confidence.)

## Test plan
- [x] `cargo build`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test` — 35/35 passing (canonical-email/alias filtering, file-level default/override merging for tags/accessType/allowExisting, template rendering against the checked-in policy templates, org alias-or-id resolution, a clap `debug_assert()` self-check)
- [x] Live-verified against the real dev org: `list-tags`, `list-users`, `list-policies` (read-only queries); `create-policies --dry-run` output matched the real dev policies byte-for-byte after placeholder substitution; `create-tag`/`update-tag`/`invite`/`dismiss-invite`/`approve-activity` all exercised live, including a real batch invite (existing-member filtering correctly skipped an already-registered invitee, who was added to the tag directly via `update-tag` instead); confirmed `--help` now works at every level post-clap-migration


tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,937 @@
+//! Turnkey org invitation management: `invite` and `dismiss-invite` subcommands.
+//!
+//! Auth resolves the same way as the official `tvc` CLI: env vars
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed — module doc comment now lists all subcommands (invite/dismiss-invite, activity approve/reject, tag CRUD, user listing, policy CRUD) in c01e3396.

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,937 @@
...
+
+/// Resolve org_id/api_base_url/api_key from env vars, or from
+/// ~/.config/turnkey/tvc.config.toml (selecting `--org` or the active org).
+fn resolve_auth(org_override: Option<&str>) -> Result<Auth> {
+    let creds = match resolve_from_env()? {
+        Some(v) => v,
+        None => resolve_from_config(org_override)?,
+    };
+
+    let api_key = match creds.curve {
```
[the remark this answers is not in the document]
prasanna-anchorage: Kept the current behavior but clarified why via comment: this intentionally mirrors the official `tvc` CLI's own `load_credentials_from_env_vars`, which has the identical shape (TVC_API_BASE_URL only applies once the org id + both keys are set). TVC_API_BASE_URL exists for the fully-env-var-driven CI path, not to override one field of a config-file-backed org — see the doc comment added on `resolve_from_env` in c01e3396.

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,937 @@
...
+    per-invitee accessType overrides the file-level default, which overrides --access-type)\n  \
+    tvc-deploy dismiss-invite --invitation-id <id> [--org <alias>]\n  \
+    tvc-deploy approve-activity --activity-id <id> [--org <alias>]\n  \
+    tvc-deploy reject-activity --activity-id <id> [--org <alias>]\n  \
+    tvc-deploy list-tags [--org <alias>]\n  \
+        (prints user-tag id + name pairs; use the id in invitees.json \"tags\", not the name)\n  \
+    tvc-deploy list-policies [--org <alias>]\n  \
+    tvc-deploy create-policy --name <name> --effect allow|deny --notes <text> \
+    [--condition <tql>] [--consensus <tql>] [--org <alias>]\n  \
+    tvc-deploy create-policies --file <policies.json> [--vars <vars.json>] [--dry-run true] \
```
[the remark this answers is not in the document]
prasanna-anchorage: Resolved as part of migrating the whole CLI to clap (c01e3396) — the hand-written USAGE string (which had drifted, as you found) is gone entirely. `--notes` now shows as optional with `[default: ""]` in `create-policy --help`, generated directly from the same clap default that the code already used.

tools/tvc-deploy/src/main.rs
```diff
@@ -82,7 +93,7 @@ fn parse_args() -> Result<(String, HashMap<String, String>)> {
             other => return Err(other.unexpected().into()),
         }
     }
-    let subcmd = subcmd.ok_or_else(|| anyhow!("missing subcommand\n{USAGE}"))?;
+    let subcmd = subcmd.ok_or_else(|| anyhow!("missing subcommand\n{USAGE}\n{}", invite::USAGE))?;
```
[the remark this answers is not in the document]
prasanna-anchorage: Fixed by migrating the whole CLI from a hand-rolled lexopt+HashMap parser to clap derive (c01e3396) — `--help`/`-h` now work at both the top level and per-subcommand, matching the actual `tvc` CLI's own convention.

pepe-anchor (commented):
[AI Review]

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,1653 @@
...
+        .into_iter()
+        .filter(|name| !vars.contains_key(name))
+        .collect();
+    if !missing.is_empty() {
+        bail!(
+            "{path} references placeholders with no --vars entry: {}",
+            missing.join(", ")
+        );
+    }
+    let rendered = render_template(&content, &vars);
```
pepe-anchor, line 830: must: load_policies_file only checks that placeholders found in the raw template text (line 804) have a matching key in --vars; it never re-scans the rendered output. A --vars entry whose value itself contains {{...}} (e.g. {"FOO": "{{BAR}}"} for a template referencing {{FOO}}) passes this check, and render_template (line 814) bakes the literal unresolved {{BAR}} into the policy that gets submitted to Turnkey. For templates/readonly-policy.json, a broken consensus expression this way means the EFFECT_DENY guardrail never matches any real user and silently never fires, the opposite of the deny-always-wins guarantee the template relies on. No test covers a nested placeholder.

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,1653 @@
...
+            args.allow_existing,
+        )])
+    }
+}
+
+/// Lowercase, and drop any `+suffix` from the local part, so
+/// `Alice+dev1@Co.com` and `alice@co.com` compare equal -- plus-addressed
+/// aliases of the same mailbox map to the same canonical identity.
+fn canonical_email(email: &str) -> String {
+    let email = email.to_lowercase();
```
pepe-anchor: should: canonical_email lowercases and strips a +suffix but never trims, so canonical_email(" alice@co.com") does not equal canonical_email("alice@co.com"). A padded email in invitees.json (a realistic spreadsheet/Slack paste artifact) silently defeats partition_existing, sending a duplicate invite instead of the intended skip.

```suggestion
    let email = email.trim().to_lowercase();
```

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,1653 @@
...
+                receiver_user_name: user_name,
+                receiver_user_email: email,
+                receiver_user_tags: tags,
+                access_type,
+                sender_user_id: String::new(),
+            },
+            args.allow_existing,
+        )])
+    }
+}
```
pepe-anchor, line 545: should: build_invitations takes the --file branch unconditionally whenever --file is set (line 478) and never reads args.user_name/args.email/args.tags/args.allow_existing in that branch. invite --file batch.json --user-name Alice --tags <uuid> --allow-existing silently drops all four flags and sends the batch from the file as is; only the --tags file-mode-only behavior is hinted at in --help text, the rest are not.

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,1653 @@
...
+            .client
+            .update_user_tag(
+                auth.org_id.clone(),
+                current_timestamp_ms(),
+                UpdateUserTagIntent {
+                    user_tag_id: args.tag_id.clone(),
+                    new_user_tag_name: args.name.clone(),
+                    add_user_ids,
+                    remove_user_ids,
+                },
```
pepe-anchor, line 1313: should: update_tag builds and submits an UpdateUserTagIntent even when --add-user-ids, --remove-user-ids, and --name are all omitted, since none are required and there is no check that at least one is set. update-tag --tag-id <uuid> --org dev alone creates a genuine Turnkey activity, possibly needing a live consensus approval from another org member, for zero actual effect on the tag.

tools/tvc-deploy/templates/releaser-initiator-policies.json
```diff
@@ -0,0 +1,25 @@
...
+      "consensus": "approvers.any(user, user.tags.contains('{{RELEASER_TAG_ID}}'))",
+      "notes": ""
+    },
+    {
+      "policyName": "Allow users with releaser tag to update and delete deployments",
+      "effect": "allow",
+      "condition": "activity.type == 'ACTIVITY_TYPE_UPDATE_TVC_APP_LIVE_DEPLOYMENT' || activity.type == 'ACTIVITY_TYPE_DELETE_TVC_DEPLOYMENT'",
+      "consensus": "approvers.any(user, user.tags.contains('{{RELEASER_TAG_ID}}'))",
+      "notes": ""
+    },
```
pepe-anchor, line 16: should: This single ALLOW policy grants the releaser tag both UPDATE_TVC_APP_LIVE_DEPLOYMENT and the irreversible DELETE_TVC_DEPLOYMENT, with notes left empty. RESTORE_TVC_DEPLOYMENT is not granted, so this role can destroy a deployment but not restore one. The PR says this mirrors what is already live in dev, but this PR turns that into a template explicitly meant for reuse in other environments (per the README), propagating an unreviewed privilege bundle with no stated justification.

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,1653 @@
...
+                )
+                .collect();
+
+            let (kept, skipped) = partition_existing(invitations_with_allow, &existing_emails);
+            for s in &skipped {
+                println!(
+                    "skipping {} <{}>: matches an existing member or pending invitation in org {} \
+                     (set \"allowExisting\": true for this invitee, or at the top level of \
+                     the file, to invite it anyway)",
+                    s.receiver_user_name, s.receiver_user_email, auth.org_id
```
pepe-anchor, line 620: nit: This skip message always says to set allowExisting in the file or its top level, even when the invocation used --user-name/--email with no --file at all, pointing the operator at a file that does not exist for that invocation.

tools/tvc-deploy/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
+error: app <app-id> already has 1 deployment activity(ies) awaiting consensus: <activity-id>
+approve or reject the existing one first (tvc-deploy approve-activity / reject-activity --activity-id <id>), or pass --force to submit anyway
+```
+
+Resolve the existing activity (approve or reject it) and re-run, or pass
+`--force` to submit anyway. The check uses the active org by default; pass
+`--org <alias>` on `deploy` if the deployment's org differs from it.
+
+### Basic / read-only access
```
pepe-anchor, line 125: nit: This subsection covers accessType vs. policy-driven scoping and the readonly-policy.json walkthrough, unrelated to deploying. It reads more naturally under Batch-creating policies from a template (line 97) or as its own top-level heading.

tools/tvc-deploy/src/invite.rs
```diff
@@ -0,0 +1,2114 @@
...
+                let access_type = match &e.access_type {
+                    Some(s) => parse_access_type(s)?,
+                    None => default_access,
+                };
+                let receiver_user_tags = e.tags.unwrap_or_else(|| default_tags.clone());
+                let allow_existing = e.allow_existing.unwrap_or(default_allow_existing);
+                Ok((
+                    InvitationParams {
+                        receiver_user_name: e.user_name,
+                        receiver_user_email: e.email,
```
pepe-anchor, line 514: should: The recent trim fix hardened the existing-member skip check but not the value actually sent, so a padded address passes the skip check and is then submitted raw, risking a rejected atomic batch or a stored invite that never matched the normalized form the tool compared against.

<details>

`canonical_email` (line 551) does `email.trim().to_lowercase()` for the dedup comparison, but `build_invitations` sets `receiver_user_email` from the raw input in both modes: batch `receiver_user_email: e.email` (line 514) and single `receiver_user_email: email` (line 537). A value like `"  alice@example.com  "` (the spreadsheet/Slack paste artifact commit `dbb5d86` targeted) canonicalizes fine, passes `partition_existing`, then the `InvitationParams` still carries the padded string. `create_invitations` bundles all invitees into one atomic activity, so one padded entry that fails server-side `tk_email` validation fails the whole batch after the tool reported the skip check passed. The fix is half-done: it normalized the comparison but not the payload. Trim `receiver_user_email` (and ideally `receiver_user_name`) at build time so what is sent matches what was canonicalized.

</details>

tools/tvc-deploy/src/main.rs
```diff
@@ -120,32 +181,32 @@ fn write_secret_file(path: &Path, contents: &str) -> Result<()> {
...
-        None => 3000,
-    };
+fn deploy(sh: &Shell, args: &DeployArgs) -> Result<()> {
+    validate_digest(&args.expected_digest)?;
+
+    if !args.force {
+        // Turnkey has no dedup for create_tvc_deployment: submitting the same
+        // deploy twice while the first is still ConsensusNeeded creates a
+        // second, independent activity instead of reusing it (see README).
+        let pending = invite::find_pending_deployments(args.org.as_deref(), &args.app_id)?;
```
pepe-anchor, line 191: should: On `deploy`, `--org` selects the org the pending-activity dedup guard queries, but it is not forwarded to the `tvc deploy create/approve` shellouts, so the guard can inspect a different org than the one actually deployed to and pass while a duplicate is already pending in the real target org.

<details>

`find_pending_deployments(args.org.as_deref(), &args.app_id)` (line 191) queries the org named by `--org` via the direct Turnkey API. The actual deploy, `tvc deploy create` (line 250) and `tvc deploy approve` (line 257), passes no org, so `tvc` resolves its own active org / `TVC_ORG_ID`. When `--org` differs from tvc active org the guard checks org A while the deploy lands in org B: it can report clean and still submit the duplicate the guard exists to prevent. The README explicitly tells operators to `pass --org on deploy if the deployment org differs`, which makes this path likely to be hit. The all-env-var CI path (no `--org`) is unaffected since both sides read the same env org. Fix: forward the resolved org to the `tvc` shellout, or reject `--org` on `deploy` (and drop the README advice) so the guard can only ever match tvc own active org.

</details>

pepe-anchor (approved):

PR #414, opened 2026-07-16
feat(parser): route machine-readable intermediate_output alongside SignablePayload

prasanna-anchorage:
## Summary

Surfaces the structured decode the parser already computes (Solana instruction name, discriminator, named accounts, program call args) in the output envelope, so downstream policy engines can validate against it instead of only the human-readable, flattened text.

## Changes

- **proto**: add `bytes intermediate_output = 5` to `ParsedTransactionPayload`, and an opt-in `bool include_intermediate_output` on `ParseRequest` (default `false`).
- **chain-agnostic plumbing**: thread a `ConversionResult { payload, intermediate_output }` through the converter trait, registry, and every chain. Written once — only Solana populates it today; other chains return `None` and can add a schema later with no further envelope changes.
- **Solana schema**: `visualsign_solana::intermediate` (published, borsh-serializable, deterministic `BTreeMap` + canonical JSON) with a `schema_version` drift guard so consumers can pin the shape.
- **CLI**: `--with-intermediate` dumps the raw borsh bytes for inspection.

## Backward compatibility

The new field is excluded from the derived Borsh encoding (`#[borsh(skip)]`). The signing path derives the digest from the payload's own field and appends the intermediate bytes **only when present**, so the no-intermediate case is byte-for-byte identical to the prior four-field digest. Emission is opt-in per request, so existing clients and signatures are unaffected until they opt in.

## Testing

- `make -C src test` — all pass.
- `cargo clippy --all-targets -- -D warnings` — clean.
- Unit tests assert the empty case equals the legacy digest and the present case appends `borsh(Vec<u8>)`.


src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+fn canonical_args_json(args: &serde_json::Map<String, Value>) -> String {
+    // Re-key into a BTreeMap so JSON output is alphabetized regardless of
+    // upstream insertion order. We hold references to avoid cloning Values.
+    let ordered: BTreeMap<&String, &Value> = args.iter().collect();
+    // serde_json::to_string never fails on a Map<String, Value>; on the
+    // off-chance it does we fall back to an empty object so the surrounding
+    // borsh encoding stays well-formed.
+    serde_json::to_string(&ordered).unwrap_or_else(|_| "{}".to_string())
```
pepe-anchor: should: `canonical_args_json` alphabetizes only top-level keys, but the module doc and its test claim canonical alphabetized JSON, and with `serde_json/preserve_order` active in the workspace nested object keys serialize in insertion order, so a consumer that mirrors this published schema and relies on canonical JSON can mismatch.

<details>

The returned string is stored as `program_call_args_json: String` (line 112), borsh-encoded into `intermediate_output` and signed into the digest. Line 186 re-keys only the top-level `Map` into a `BTreeMap`; nested `serde_json::Value` objects serialize per serde_json build features. `cargo tree -e features` shows `serde_json feature preserve_order` enabled transitively in the workspace graph, so in a unified build nested keys follow insertion order rather than sorted order. The doc lines 17-19 (canonical JSON, keys are alphabetized) and the test `canonical_args_json_alphabetizes_keys` (line 341, top-level only) overstate the guarantee. This is the exact serde/borsh ordering-determinism class we have been bitten by before. Fix: either canonicalize recursively (walk the Value tree re-keying nested Maps) or correct the doc and test to state that only top-level keys are sorted and nested determinism relies on the upstream parser.

</details>

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -327,28 +329,75 @@ impl VisualSignConverter<SolanaTransactionWrapper> for SolanaVisualSignConverter
...
+            return Ok(ConversionResult::new(payload));
+        }
+
+        let idl_registry = create_idl_registry_from_options(&options)?;
+        Ok(
+            match build_intermediate_bytes(&message_hex, &idl_registry) {
+                Some(bytes) => ConversionResult::with_intermediate(payload, bytes),
+                None => ConversionResult::new(payload),
+            },
+        )
```
pepe-anchor, line 384: should: No test anywhere sets `include_intermediate_output: true`, so the Solana emit path (`build_intermediate_bytes` then `extract_solana_intermediate_output` then borsh then `ConversionResult::with_intermediate`) is never exercised end to end, on a path whose output is signed into the digest and consumed by policy.

<details>

`rg include_intermediate_output` returns zero `true` occurrences across the repo, and no test reads `.intermediate_output` off a converted result. Adjacent unit tests exist (`parse.rs` covers `signing_digest_bytes` with hand-built payloads; `intermediate.rs` covers the From-impl projection and borsh round-trips), but the production branch that parses a real message, applies custom IDLs, borsh-encodes, and wraps `with_intermediate` is untested, including the best-effort `None` fallback. Add at least one converter or gRPC test with `include_intermediate_output: true` asserting non-empty `intermediate_output` for a known Solana tx (ideally cross-checking a transfer amount against the `SignablePayload`), plus one asserting graceful empty on an undecodable message. `tests/common/mod.rs::options_with_idl` (the only helper wiring `idl_mappings`) and `integration/tests/parser.rs::parser_solana_native_transfer_e2e` are the natural places.

</details>

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -327,28 +329,75 @@ impl VisualSignConverter<SolanaTransactionWrapper> for SolanaVisualSignConverter
...
+/// Best-effort: if `solana_parser::parse_transaction_with_idls` cannot parse
+/// the message (e.g. an obscure variant we still display via fallback paths)
+/// we drop the intermediate output rather than fail the whole conversion. The
+/// SignablePayload is still returned so visual signing keeps working; only
+/// policy-engine evaluation degrades to "no metadata".
+fn build_intermediate_bytes(
+    message_hex: &str,
+    idl_registry: &crate::idl::IdlRegistry,
+) -> Option<Vec<u8>> {
+    match extract_solana_intermediate_output(message_hex, false, idl_registry) {
```
pepe-anchor, line 426: comment: The machine-readable `intermediate_output` is produced by `solana_parser::parse_transaction_with_idls`, a separate decoder from the `instructions::decode_instructions` path that builds the human-readable `SignablePayload`, so the two can diverge on the same bytes and a policy engine may validate a decode that differs from what the user visually verified.

<details>

For a visual-signing product this is the classic user-sees-X, machine-consumes-Y surface. Mitigations already in place and worth acknowledging: both are bound into the signed digest, the intermediate is best-effort (dropped to `None` on parse failure so policy degrades to no-metadata rather than wrong-metadata), and it is opt-in, so the real signing boundary is preserved. Transfers and accounts stay faithful because `decode_transfers` also routes through `solana_parser`. The code comment at `build_intermediate_bytes` already flags unifying to a single decoder as tracked work. Worth confirming there is a plan or drift test comparing the two decoders before the policy engine trusts these bytes.

</details>

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -327,28 +329,75 @@ impl VisualSignConverter<SolanaTransactionWrapper> for SolanaVisualSignConverter
...
+            SolanaTransactionWrapper::Legacy(transaction) => {
+                let payload = convert_to_visual_sign_payload(
+                    transaction,
+                    options.decode_transfers,
+                    options.transaction_name.clone(),
+                    &options,
+                    #[cfg(feature = "diagnostics")]
+                    &lint_config,
+                )?;
+                let hex = hex::encode(transaction.message.serialize());
```
pepe-anchor: nit: `message.serialize()` plus `hex::encode` run in both match arms before the `include_intermediate_output` early return, so every conversion on the default path pays a serialize and hex-encode it never uses.

<details>

The PR keeps the default path byte-identical in output, but this adds a full message serialize plus hex on the hot signing path (a third serialize, since `decode_transfers` already serializes the message). Defer both `let hex = ...` computations into the `if options.include_intermediate_output` branch.

</details>

pepe-anchor (commented):
[AI Review]

src/parser/cli-core/src/output.rs
```diff
@@ -216,5 +217,17 @@ pub fn parse_and_display(
...
+    // be captured verbatim as a downstream (e.g. C++) test fixture. Decoding
+    // into a typed schema is chain-specific and intentionally left to the
+    // consumer; the CLI stays chain-agnostic and passes the bytes through.
+    if let Some(bytes) = &conversion.intermediate_output {
+        println!(
+            "\n=== intermediate_output ({} bytes, borsh) ===",
+            bytes.len()
+        );
+        println!("{}", hex::encode(bytes));
+    }
```
pepe-anchor, line 236: nit: With `--output json --with-intermediate`, the `=== intermediate_output ===` header and hex are printed to stdout after the JSON via `println!`, so stdout is no longer valid JSON for a scripted consumer.

<details>

The human-format hint elsewhere in this function uses `eprintln!` (stderr) to keep stdout clean; this dump uses `println!`. Emit the hex block to stderr, or document that `--with-intermediate` is intended for interactive fixture capture rather than piping JSON.

</details>

src/parser/cli/src/serve.rs
pepe-anchor, line 1: must: The non-default `serve` feature no longer compiles because `decode_file` serializes a `ConversionResult`, which is not `Serialize`, and CI only builds default features so it will not catch this.

<details>

This PR changed `TransactionConverterRegistry::convert_transaction` to return `ConversionResult` (derives only `Debug, Clone`, see `visualsign/src/vsptrait.rs:36-40`) instead of `SignablePayload`. At `serve.rs:238`, `decode_file` still does `serde_json::to_value(&payload)` where `payload` is now a `ConversionResult`, so `cargo check -p parser_cli --features serve` fails with `the trait Serialize is not implemented for ConversionResult`. `serve` is a non-default feature (`cli/Cargo.toml:13`; `default = [solana, ethereum, tron, diagnostics]` omits it) and `make -C src test/lint` builds `parser_cli` with default features only, so the break is invisible to CI. Fix: serialize the inner payload, `serde_json::to_value(&payload.payload)`. The line is outside this PR diff hunks (the changed hunks are the two `prepare_runtime` call sites) so this is file-anchored, but it is a direct consequence of the trait change in this PR.

</details>

src/chain_parsers/visualsign-solana/src/intermediate.rs
prasanna-anchorage: Resolved in 04c686bd. `canonical_args_json` now canonicalizes recursively — the `serde_json::Value` tree is walked via `canonicalize_value`/`canonicalize_map`, re-keying every nested object into sorted order so the output is byte-deterministic regardless of `preserve_order`. The module doc and `program_call_args_json` field doc were updated to state the guarantee accurately. Added `canonical_args_json_alphabetizes_nested_keys` covering top-level, nested-object, and array-embedded-object key ordering with differing insertion orders.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
prasanna-anchorage: Resolved in 04c686bd. Expanded the `build_intermediate_bytes` doc with a `# Dual-decoder divergence` section documenting the risk, the mitigations in place (both bound into the signed digest, opt-in emission, transfers route through `solana_parser`, fail-to-`None` not wrong-metadata), and that a drift test is the guard until the structured decode becomes the single source of truth. The new `intermediate_output_emitted_for_known_transfer` test serves as the initial drift guard — it cross-checks the intermediate's transfer amount against the human-readable `SignablePayload`.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
prasanna-anchorage: Resolved in 04c686bd. Added `intermediate_output_emitted_for_known_transfer` — converts a known System Program transfer with `include_intermediate_output: true`, decodes the emitted borsh bytes back into the published `SolanaIntermediateOutput` schema, and cross-checks `decoded.transfers[0].amount == "1000000000"` against the `SignablePayload`. Also added `build_intermediate_bytes_returns_none_on_undecodable` asserting the graceful `None` fallback on garbage input (`"deadbeef"`).

src/parser/cli-core/src/output.rs
prasanna-anchorage: Resolved in 04c686bd. Both `println!` calls for the intermediate-output hex dump were switched to `eprintln!`, so stdout stays a clean JSON document even with `--with-intermediate --output json`. This mirrors the human-format hint earlier in the function, which also uses `eprintln!` to keep stdout clean.

src/parser/cli/src/serve.rs
prasanna-anchorage: Resolved in 04c686bd. `decode_file` now serializes the inner payload (`&payload.payload`) instead of the `ConversionResult` wrapper. Verified with `cargo check -p parser_cli --features serve` (was failing before the fix) and `cargo test -p parser_cli --features serve` (10 serve tests pass). CI now also passes — `ubuntu` check is green.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
prasanna-anchorage: Resolved in 04c686bd. The `message.serialize()` + `hex::encode` are now deferred into the `if options.include_intermediate_output` branch (after the early return), so the default signing path pays nothing for a blob it never uses. The match above now only builds the `payload`; a second match below the opt-in gate computes `message_hex` only when needed.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -327,28 +329,111 @@ impl VisualSignConverter<SolanaTransactionWrapper> for SolanaVisualSignConverter
...
+///
+/// The eventual architecture (tracked, not yet implemented — see
+/// [`extract_solana_intermediate_output`]) is to make the structured decode the
+/// single source of truth from which the `SignablePayload` is generated and
+/// pass these bytes through as-is, eliminating the re-parse. Until that lands,
+/// a drift test that cross-checks the intermediate against the `SignablePayload`
+/// for known transactions is the guard against silent divergence; see
+/// `intermediate_output_emitted_for_known_transfer` for an initial cross-check
+/// (transfer amount) and add more cases as the schema grows.
+fn build_intermediate_bytes(
```
pepe-anchor, line 422: should: The machine-readable `intermediate_output` is decoded by a separate path (`solana_parser::parse_transaction_with_idls`) from the human-readable `SignablePayload` (`decode_instructions`); binding both into one signed digest makes divergence tamper-evident but does not make them agree, and the only cross-check today is one System-transfer amount, so a decoder drift on the fields a policy engine actually consumes would ship uncaught.

<details>

The sole drift test is `intermediate_output_emitted_for_known_transfer` (line 740), which asserts a single System transfer amount. There is no case covering SPL transfers, `program_call_args_json` from an IDL-decoded instruction, named accounts, or a versioned/ALT transaction, which are exactly the fields PROF-355 wants policy to validate. Since the intermediate is the artifact policy trusts, add per-instruction-class drift cases that cross-check the intermediate against the same values rendered in the SignablePayload as the schema grows.

</details>

src/integration/tests/parser.rs
```diff
@@ -207,6 +209,7 @@ async fn parser_solana_native_transfer_e2e() {
         );
         tracing::debug!("Solana transaction: {}", solana_tx);
         let parse_request = ParseRequest {
+            include_intermediate_output: false,
```
pepe-anchor, line 212: should: All five `ParseRequest` literals hard-code `include_intermediate_output: false`, including `parser_solana_native_transfer_e2e`, so the end-to-end path that threads the opt-in flag through the gRPC request, populates the bytes, and surfaces them is never covered, and a regression in the flag wiring (`gateway/src/main.rs:236`) or the base64 response encoding would pass CI.

<details>

Solana is the only chain that populates the field today, so `parser_solana_native_transfer_e2e` (line 199) is the natural place to flip the flag to `true` and assert the response carries a non-empty `intermediate_output`. Library-level emit (`intermediate_output_emitted_for_known_transfer`) and gateway serde unit tests cover the two ends, but the request-flag-to-emission wiring in the middle is untested end to end.

</details>

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -0,0 +1,502 @@
...
+#[derive(BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
+pub struct SolanaParsedInstructionDataIo {
+    pub instruction_name: String,
+    pub discriminator: String,
+    pub named_accounts: BTreeMap<String, String>,
+    /// Canonical JSON string with alphabetized keys at every nesting level.
+    /// Built by recursively re-keying the `serde_json::Value` tree into sorted
+    /// order, so byte-identical inputs produce byte-identical encodings
+    /// regardless of `serde_json`'s `preserve_order` build feature.
+    pub program_call_args_json: String,
```
pepe-anchor, line 118: question: `program_call_args_json` is built from `serde_json::Value` numbers, and with no `arbitrary_precision` in the workspace a `u128`/`i128` program arg would be coerced to lossy `f64` before a policy engine ever validates it, unlike transfer amounts which are carried as strings.

<details>

SOL/SPL transfer amounts use `String` and are lossless. But `program_call_args` arrives from `solana_parser` as `serde_json::Value`; `serde_json::Number` without `arbitrary_precision` (confirmed absent in `src/Cargo.lock`) holds only i64/u64/f64. Is every policy-relevant numeric IDL arg guaranteed <= u64, or can a token/DeFi arg be u128-scale? If the latter, a policy engine keying on that arg validates a rounded value; consider normalizing numeric IDL args to strings the way transfers already are.

</details>

src/parser/app/src/routes/parse.rs
```diff
@@ -120,13 +127,40 @@ pub(crate) fn parse_with_registry(
...
+/// Backward-compatibility contract: `intermediate_output` is excluded from the
+/// derived Borsh encoding (`#[borsh(skip)]`, applied by codegen). When it is
+/// empty, this returns exactly `borsh(payload)` — byte-for-byte identical to
+/// the pre-feature four-field encoding, so existing signatures/consumers are
+/// unaffected. When it is non-empty, its Borsh `Vec<u8>` encoding (u32-LE
+/// length prefix followed by the raw bytes, matching a downstream borsh
+/// encoder) is appended. Presence is out-of-band (no tag byte), so an
+/// intermediate-unaware consumer that also appends nothing agrees on the empty
+/// case and correctly rejects the non-empty case it cannot verify.
+fn signing_digest_bytes(payload: &ParsedTransactionPayload) -> Vec<u8> {
```
pepe-anchor, line 145: comment: The signed bytes are `borsh(4 fields, field 5 skipped)` optionally followed by `borsh(Vec<u8>)` of the intermediate, a non-standard scheme the code comment says an intermediate-unaware verifier will reject on the non-empty case, so the HSM consumer (anchorage#223146) must implement the identical borsh-skip-then-append logic or signatures silently fail once a client opts in.

<details>

This is a cross-repo coordination note, not a bug. The digest is `sha256(borsh(msg with field 5 skipped) || (non-empty ? borsh(Vec<u8>) : nothing))`, not `sha256(borsh(full message))`. A verifier that borsh-encodes the whole payload including field 5, or appends the raw bytes without the u32-LE length prefix, computes a different digest. Worth pinning this in a shared spec/fixture (the CLI `--with-intermediate` hex dump helps) rather than only in the `signing_digest_bytes` doc comment.

</details>

pepe-anchor (approved):

PR #416, opened 2026-07-21
fix(scripts): add turnkey API key import helper for smoke.sh

shahan-khatchadourian-anchorage:
## Summary
- `scripts/smoke.sh` reads its auth key from `~/.config/turnkey/keys/<VSP_SMOKE_KEY>.{public,private}` (default key name `dev`), but `tvc login` (the documented onboarding step in the TVC deploy runbook) only ever populates `~/.config/turnkey/orgs/<org>/api_key.json` — a different, non-interchangeable local credential store. A workstation that's only run `tvc login` hits a plain "no such file or directory" with no path forward, and the only way to get unstuck is to hand-derive the tkcli key-store format from `go-sdk`'s `pkg/store/local` source.
- Adds `scripts/import-turnkey-api-key.sh --org <org-name> [--key-name <name>]`, which converts an org's `api_key.json` into the `keys/<name>.{public,private,meta}` format smoke.sh expects (verified byte-for-byte against an existing hand-provisioned key: same public/private file lengths, same `<hex>:<curve>` private-key suffix, same meta JSON shape and `SIGNATURE_SCHEME_TK_API_*` scheme string). Refuses to overwrite existing key files, matching `go-sdk`'s own `Store.Store()` behavior.
- Documents the two-credential-store distinction, the `jq` runtime dependency, and points at the new script from `smoke.sh`'s own header comment, so the next person hitting this error finds the fix without re-deriving it from scratch.

## Test plan
- [x] `bash -n` on both scripts
- [x] `scripts/import-turnkey-api-key.sh` run against a fixture `HOME` (fake org + `tvc.config.toml`): produced `keys/testorg.{public,private,meta}` matching the real tkcli format exactly, resolved the org id from `tvc.config.toml`, and correctly refused to run again without overwriting
- [x] Live-verified against a real dev-org credential: converted, then ran `VSP_SMOKE_KEY=<name> ./scripts/smoke.sh --turnkey-client-path ~/projects/visualsign-turnkeyclient` end-to-end — PASS (render + attestation + signature all verified against the live enclave)
- [x] `smoke.sh -h` / `import-turnkey-api-key.sh -h` still print correctly after the doc changes

### DIFF
--- a/scripts/import-turnkey-api-key.sh
+++ b/scripts/import-turnkey-api-key.sh
@@ -45,6 +45,17 @@ done
 [ -n "$ORG" ] || { echo "ERROR: --org is required" >&2; exit 2; }
 [ -n "$KEY_NAME" ] || KEY_NAME="$ORG"
 
+# Both values become path components below; reject anything that could escape
+# the intended orgs/keys directories (e.g. --key-name ../../etc/foo).
+case "$ORG" in
+  */* | .*) echo "ERROR: --org must not contain '/' or start with '.': $ORG" >&2; exit 2 ;;
+esac
+case "$KEY_NAME" in
+  */* | .*) echo "ERROR: --key-name must not contain '/' or start with '.': $KEY_NAME" >&2; exit 2 ;;
+esac
+
+command -v jq >/dev/null 2>&1 || { echo "ERROR: jq is required" >&2; exit 2; }
+
 SRC="$HOME/.config/turnkey/orgs/$ORG/api_key.json"
 [ -f "$SRC" ] || {
   echo "ERROR: no such org credential: $SRC" >&2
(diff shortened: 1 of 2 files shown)


prasanna-anchorage (approved):
Verified: path-traversal validation and jq error handling from Copilot's review are addressed in the second commit. No key material is logged; correct file permissions (0600 private/meta, 0644 public). CI green (6/6). LGTM.

scripts/import-turnkey-api-key.sh
```diff
@@ -0,0 +1,118 @@
...
+
+SRC="$HOME/.config/turnkey/orgs/$ORG/api_key.json"
+[ -f "$SRC" ] || {
+  echo "ERROR: no such org credential: $SRC" >&2
+  echo "       (run \`tvc login\` first, or check ~/.config/turnkey/tvc.config.toml for the org name)" >&2
+  exit 2
+}
+
+KEYS_DIR="$HOME/.config/turnkey/keys"
+mkdir -p "$KEYS_DIR"
```
pepe-anchor, line 67: should: If `api_key.json` is truncated/missing `public_key` or `private_key`, `jq -r` emits the string `null`, so the script writes garbage key files that only fail cryptically at `smoke.sh` auth time, and the overwrite-refusal at lines 61-63 then blocks a clean re-run. Use `jq -re` so an absent field aborts the import.

<details>

At lines 65-67, `PUBLIC_KEY="$(jq -r ... )"` returns the 4-char string `null` for an absent field, which is non-empty and passes the go-sdk `LocalStore` `Size >= 1` guard; the resulting `.private` of `null:p256` then fails inside `FromTurnkeyPrivateKey` with an opaque decode error far from the cause, and because `keys/<name>.*` now exist the user must manually delete them before re-importing. `CURVE` is already protected by the `case ... *)` default at line 73, so only `public_key`/`private_key` are exposed, but applying `-re` uniformly is harmless. `jq -re` exits non-zero when the last output is null/false, and `set -e` turns that into a clean import-time failure. Well-formed `tvc` output always carries all three fields, so this only bites on truncated/partial credential files.

</details>

```suggestion
PUBLIC_KEY="$(jq -re '.public_key' "$SRC")"
PRIVATE_KEY="$(jq -re '.private_key' "$SRC")"
CURVE="$(jq -re '.curve' "$SRC")"
```

pepe-anchor (approved):

PR #420, opened 2026-07-27
feat(tvc-deploy): add verify-digest subcommand

pepe-anchor:
## Why

`deploy` already verifies that the image's `/parser_app` hashes to the expected digest, and refuses to proceed on a mismatch. But that gate only fires once `deploy` is running.

Some workflows have to record the expected digest somewhere else first, before the deployment is ever created. Those have no way to check the digest before committing to it, and by the time `deploy` would catch a wrong one, the record is already made.

## What

`tvc-deploy verify-digest --image-url <url> --expected-digest <sha256>` exposes that same gate as a standalone command, so it can be run as a precondition.

It reuses `validate_digest` + `verify_image_digest` rather than reimplementing the check, so there stays exactly one definition of what "the digest matches" means, and the failure message is identical wherever it fires. No existing code path changed.

Needs only Docker. No Turnkey auth, no org selection.

## Testing

Ran against the real `v0.819.0` GHCR image.

Matching digest:

```
$ cargo run -- verify-digest \
    --image-url ghcr.io/anchorageoss/parser_app:v0.819.0@sha256:<...> \
    --expected-digest <digest from the release notes>
✓ digest matches
```

Wrong digest fails with the existing mismatch message, and a malformed digest is rejected by `validate_digest` before Docker is touched.

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.

## Rollback

Additive: a new subcommand and a README section. Revert the commit; nothing else reads it.

### DIFF
--- a/tools/tvc-deploy/README.md
+++ b/tools/tvc-deploy/README.md
@@ -160,6 +160,24 @@ Resolve the existing activity (approve or reject it) and re-run, or pass
 `--force` to submit anyway. The check uses the active org by default; pass
 `--org <alias>` on `deploy` if the deployment's org differs from it.
 
+### Verifying the digest on its own
+
+`deploy` runs the digest gate itself, so a normal deploy needs nothing extra.
+`verify-digest` exposes the same gate as a standalone command:
+
+```
+tvc-deploy verify-digest \
+  --image-url "$IMAGE_URL" --expected-digest "$PIVOT_DIGEST"
+```
+
+Useful when the expected digest has to be recorded somewhere else before the
+deployment is created. `deploy` verifies the digest, but only once it is already
+running, which is too late to stop a wrong digest from being committed
+elsewhere first. Run this, record the digest, then deploy.
+
+No Turnkey auth needed, just Docker: it creates a container from the image,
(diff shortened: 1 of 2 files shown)


shahan-khatchadourian-anchorage:
## Review

Looks good — small, additive, and the reuse claims in the description check out. No blocking issues.

### Strengths

- **Genuine reuse, single source of truth** — `verify_digest` is a two-line composition of the existing `validate_digest` + `verify_image_digest`; format validation, extraction, hashing, cleanup-on-error, and the failure message all stay in one place.
- **Ordering matches `deploy`** — `validate_digest` runs before Docker is touched, so a malformed digest fails fast.
- **Existing tests already cover it** — `cli_parses_all_subcommands` exercises the new clap definition, and `validate_digest` has accept/reject unit tests. The composition adds no new logic to test.
- **Clear README section** — explains when to use it and what it does mechanically (Docker only, no Turnkey auth).

shahan-khatchadourian-anchorage (approved):
LGTM. No blocking comments.

prasanna-anchorage (approved):

PR #422, opened 2026-07-28
feat(prs-556): make ABI trust posture an explicit policy type

pepe-anchor:
## Why

[PRS-556](https://linear.app/anchorlabs/issue/PRS-556/move-abi-trust-posture-from-per-request-behavior-to-a-deploy-time-flag), part 1 of 2. Library groundwork only, no flags and no deployment changes here. #421 stacks on top of this and adds the deploy-time flags.

Split out because this half carries the one real behaviour change in PRS-556, and it deserves reading on its own rather than buried under a PR whose headline is "add two cmdline flags".

Today the parser's willingness to honour caller-supplied ABI mappings that carry no signature is an unconditional property of the code. #406 made unsigned acceptance the default to fix a real bug (unsigned proxy to unsigned implementation was silently dropped), but that left every caller able to omit the signature and get an unverified ABI decoded and rendered, with no way for anyone to choose otherwise.

Making that choice explicit needs a type to represent it first. That is all this PR does.

## What

`visualsign::signing::MetadataTrustPolicy`, a chain-neutral XOR enum:

```rust
pub enum MetadataTrustPolicy {
    AcceptUnsigned,
    RequireAllowlistedSigner(SignerAllowlist),
}
```

No `Default` impl, deliberately. A default is exactly the implicit choice PRS-556 exists to remove, so the type refuses to let a caller skip the decision.

`EthereumVisualSignConverter::with_policy` takes it, and `try_extract_from_chain_metadata` threads it down to `validate_abi_signature`, which now takes `Option<&SignerAllowlist>` instead of a bare allowlist.

Every existing caller passes the equivalent of today's behaviour, so this PR is behaviour-preserving except for the one item below. `parser_cli` moves to require-signed against the local dev key it already signs its own ABI files with.

`try_extract_from_chain_metadata` now returns `AbiExtraction` (registry plus counts) instead of `Option<AbiRegistry>`. A request whose every mapping the posture refused used to be byte-identical to a request that supplied no mappings at all: both return no registry, both render the raw selector, and the only thing separating them was a `log::warn!`. That is not a channel anyone can read from the enclave, since `parser_app` declares no `log` dependency and initialises no logger, so those calls are compiled-in no-ops there rather than log lines nobody reads. The counts now survive the return so the follow-up that renders them has something to read.

### The one behaviour change

Under accept-unsigned, a signature that *is* present is still verified, but its signer is no longer checked against an allowlist.

The old combination was incoherent. In production (no `dev-signing` feature, no env var) the allowlist is empty, so a correctly signed ABI was rejected while the same ABI with the signature stripped was accepted. An attacker facing an identity check just drops the signature, so the check bought nothing while making the strictly worse input the accepted one. Pinned by a test so a change back to the old behaviour fails loudly.

Being precise about what the surviving integrity check buys, because an earlier draft of these doc comments overstated it: the signature is verified against the public key carried alongside it in the same untrusted `SignatureMetadata`. It catches a tamperer who mutates `abi.value` and leaves the signature and key alone. It does not catch one who controls the whole entry and re-signs with a key of their own. Worth keeping, but it is not provenance, and the comments now say so.

Because identity is no longer enforced in that posture, the aggregated unverified counter now counts "identity was not enforced" rather than "signature absent". Under accept-unsigned a self-signed entry is exactly as unattributed as an unsigned one, and counting only the latter would report zero unverified mappings for a request whose entire decode came from an unattributed caller ABI.

### The intermediate state is deliberate

`parser_app` builds its converter through `EthereumVisualSignConverter::new()`, which is accept-unsigned, and this PR does not change `parser/app/src/registry.rs`. Merging this alone therefore moves the production enclave binary from rejecting an ABI entry signed by a non-allowlisted key to accepting it, with no flag to opt out until #421 lands.

This is not an escalation of attacker capability: the same render was already reachable on `main` by omitting the signature, which is the whole reason the old pairing was incoherent. But it is a real change to what the deployed binary does, so it should not be discovered later:

- Merging both PRs together is fine, and #421 targets this branch.
- Merging this one alone is also fine, and lands the accept-unsigned posture explicitly rather than by accident.

Either way, `create_registry_runs_the_accept_unsigned_abi_posture` in `parser/app/src/registry.rs` now pins which posture the deployed binary runs, so moving it takes a deliberate test edit instead of compiling clean and leaving the suite green.

## Testing

`make lint`, `make test`, and `cargo fmt --all -- --check` all clean on this branch alone:

```
FMT_EXIT=0     cargo fmt --all -- --check
LINT_EXIT=0    make lint    (clippy -D warnings, incl. parser_cli + diagnostics variants)
TEST_EXIT=0    make test    (38 suites, 0 failures)
```

Unit coverage in `abi_metadata` for both postures on the same fixture: unsigned accepted under `AcceptUnsigned` and rejected under `RequireAllowlistedSigner`, a mixed request dropping only the unsigned half, and a tampered signature rejected under both postures.

Each of the four assertions this PR relies on was mutation-checked (revert the production line, confirm the test goes red):

| Test | Mutation it catches |
|---|---|
| `create_registry_runs_the_accept_unsigned_abi_posture` | `create_registry` switching posture |
| `test_register_installs_require_signed_posture` | `EthereumPlugin::register` reverting to `new()` |
| `test_accept_unsigned_counts_self_signed_entry_as_unverified` | counter going back to `signature.is_none()` |
| `test_total_refusal_is_distinguishable_from_absent_metadata` | rejection count not being carried out |

The first two go through the registry rather than through the helper that feeds it. Asserting on a `cli_trust_policy()`-shaped helper in isolation proves nothing about what gets registered: editing the `register` call back to `new()` leaves such an assertion passing.

The end-to-end evidence (identical request against two `parser_app` instances differing only in cmdline) lands in #421, since it needs the flags.

## Rollback

Plain revert. No state, schema, wire-format, or deployment changes.

Correcting an earlier claim in this section: this is **not** invisible from outside the library. `parser_app` picks up the accept-unsigned posture described above, so reverting after deploy means redeploying the enclave binary to restore the previous (incoherent) rejection of signed-by-unlisted entries. Nothing persists, so the revert is otherwise clean.

## Backwards compatibility

No proto or payload changes. Behaviour-preserving apart from the allowlist-identity item called out above, which is a deliberate correction of an incoherent combination rather than a regression.

Breaking for out-of-tree embedders of `visualsign-ethereum`, none known in the tree:

- `EthereumVisualSignConverter::with_signers` and `with_registry_and_signers` are removed. Replaced by `with_policy` and `with_registry_and_policy`, which take a `MetadataTrustPolicy` instead of a bare `SignerAllowlist`. Migration is `with_signers(allow)` to `with_policy(RequireAllowlistedSigner(allow))`.
- `with_registry` and `Default`/`new` keep compiling unchanged but now resolve to `AcceptUnsigned`, so an embedder that was relying on identity enforcement loses it silently. Those call sites should move to `with_registry_and_policy` / `with_policy`.
- `try_extract_from_chain_metadata` returns `AbiExtraction` rather than `Option<AbiRegistry>`. The registry is now `.registry`.

## Not in scope

The deploy-time flags, `tvc-deploy` manifest changes, and the workflow changes are in #421, which targets this branch. The cmdline parser that turns flag occurrences into a `MetadataTrustPolicy` moves there too, so the flag and its parser land together and the empty-allowlist contract gets settled where it is actually observable, rather than shipping a parser here with no non-test caller.

Solana IDL mappings still carry the exact pairing this PR removes from Ethereum (unsigned IDLs accepted, correctly signed IDLs from unlisted signers rejected). `MetadataTrustPolicy` is chain-neutral so extending it is mechanical, but it is out of scope here and wants its own ticket. The comments in `visualsign-solana` that described this as "parity with the Ethereum ABI path" are corrected here to say it is a known gap, so a Solana reader does not conclude the pairing is deliberate.

Surfacing verified/unverified status in the rendered payload is still not done. PRS-555 was closed Done on 2026-07-14, but `abi_registry.rs` carries no verified or unsigned field, so the gap it described is real and now has no open ticket pointing at it. This PR carries the counts out of the extractor so the eventual fix has something to read, and does not add a payload field.

### DIFF
--- a/src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
+++ b/src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
@@ -27,11 +27,46 @@ enum AbiSignatureError {
     Validation(String),
 }
 
+/// Outcome of extracting caller-supplied ABI mappings from `ChainMetadata`.
+///
+/// The counts travel with the registry because the rendered payload cannot tell
+/// these two requests apart: one that supplied no ABI mappings at all, and one
+/// whose every mapping this deployment's posture refused. Both fall back to the
+/// raw selector, and the only thing separating them used to be a `log::warn!`.
+/// That is not a channel anyone can read from the enclave: `parser_app` declares
+/// no `log` dependency and initialises no logger, so those calls are compiled-in
+/// no-ops there rather than log lines nobody happens to look at.
+///
+/// Surfacing the counts in the payload itself needs a field on the payload (the
+/// follow-up to PRS-555, which shipped without one). Until that exists, this at
+/// least lets the caller distinguish the two cases instead of discarding the
+/// difference inside the extractor.
+#[derive(Default)]
+pub struct AbiExtraction {
+    /// The accepted entries, or `None` when nothing was registered, whether
+    /// because the request carried no usable mapping or because every mapping
+    /// was refused. `rejected_by_policy` tells those apart.
+    pub registry: Option<AbiRegistry>,
+    /// Entries refused by the deployment's trust posture: no signature at all
+    /// under require-signed, or a signature that failed to verify or whose
+    /// signer is not allowlisted. Entries dropped for a malformed address,
+    /// oversized JSON, or unparseable ABI are not counted here; those are the
+    /// caller's own malformed input rather than a trust decision.
+    pub rejected_by_policy: usize,
+    /// Accepted entries whose signer identity was not established. Under
+    /// require-signed this is always zero. Under accept-unsigned it counts every
+    /// accepted entry, signed or not: see the counting note in
+    /// [`try_extract_from_chain_metadata`] for why a present signature does not
+    /// reduce it.
+    pub unverified: usize,
+}
+
 /// Extract and validate ABIs from `ChainMetadata`, if present.
 ///
 /// Navigates `ChainMetadata -> Ethereum -> abi_mappings` and registers each ABI
-/// with its contract address. Returns `None` if the metadata doesn't contain
-/// any Ethereum ABI mappings.
+/// with its contract address. Returns an [`AbiExtraction`] whose `registry` is
+/// `None` when nothing was registered; its counts say why, so total refusal is
+/// distinguishable from metadata that carried no ABI mappings at all.
 ///
 /// The `chain_id` is needed to register address-to-ABI mappings in the registry.
 ///
@@ -50,9 +85,13 @@ enum AbiSignatureError {
 ///   malformed and unauthorized signatures are all rejected. An empty allowlist
 ///   rejects everything (fail-closed).
 /// - **Under [`MetadataTrustPolicy::AcceptUnsigned`]**, an entry with
-///   `signature: None` is registered rather than dropped; unsigned entries are
-///   counted and reported in a single aggregated warning once per request (no
-///   per-entry flag or marker is stored in the registry, tracked in PRS-555). An
+///   `signature: None` is registered rather than dropped. Every accepted entry is
+///   counted as unverified in this posture, signed or not, since a signature whose
+///   signer is never checked establishes no provenance either; the count is
+///   returned in [`AbiExtraction::unverified`] and reported in a single aggregated
+///   warning once per request. No per-entry flag or marker is stored in the
+///   registry, so a decode backed by caller metadata is still indistinguishable
+///   from a built-in one in the rendered payload. An
 ///   entry that DOES carry a signature must still verify against the public key
 ///   supplied alongside it in the same (untrusted) `SignatureMetadata`, since a
 ///   present-but-invalid signature is a stronger signal of tampering than simply
@@ -91,17 +130,21 @@ pub fn try_extract_from_chain_metadata(
     chain_metadata: Option<&ChainMetadata>,
     chain_id: u64,
     policy: &MetadataTrustPolicy,
-) -> Option<AbiRegistry> {
-    let chain_metadata = chain_metadata?;
-    let chain_metadata::Metadata::Ethereum(ethereum) = chain_metadata.metadata.as_ref()? else {
-        return None;
+) -> AbiExtraction {
+    let Some(chain_metadata) = chain_metadata else {
+        return AbiExtraction::default();
+    };
+    let Some(chain_metadata::Metadata::Ethereum(ethereum)) = chain_metadata.metadata.as_ref()
+    else {
+        return AbiExtraction::default();
     };
     if ethereum.abi_mappings.is_empty() {
-        return None;
+        return AbiExtraction::default();
     }
 
     let mut registry = AbiRegistry::new();
-    let mut unsigned_count: usize = 0;
+    let mut unverified_count: usize = 0;
+    let mut rejected_by_policy: usize = 0;
     for (address, abi) in &ethereum.abi_mappings {
         // Validate address first (cheap) before expensive signature/ABI operations
         let parsed_address = match address.parse::<alloy_primitives::Address>() {
@@ -127,7 +170,14 @@ pub fn try_extract_from_chain_metadata(
         // both postures a signature that IS present must validate, since a
         // present-but-invalid signature signals tampering rather than simply an
         // unsigned source.
-        let is_unsigned = abi.signature.is_none();
+        //
+        // What gets counted as unverified is "identity was not enforced", not
+        // "signature absent". Under accept-unsigned the signer is never checked
+        // against an allowlist, so an entry an attacker signed with a key of
+        // their own is exactly as unattributed as one they left unsigned;
+        // counting only the latter would report zero unverified mappings for a
+        // request whose entire decode came from an unverified caller ABI.
+        let identity_unverified = policy.signer_allowlist().is_none() || abi.signature.is_none();
         if let Some(proto_sig) = abi.signature.as_ref() {
             let signature = convert_proto_signature(proto_sig);
             if let Err(e) = validate_abi_signature(
@@ -140,6 +190,7 @@ pub fn try_extract_from_chain_metadata(
                 log::warn!(
                     "Skipping ABI mapping for '{address}': signature validation failed: {e}"
                 );
+                rejected_by_policy += 1;
                 continue;
             }
         } else if !policy.accepts_unsigned() {
@@ -147,6 +198,7 @@ pub fn try_extract_from_chain_metadata(
                 "Skipping ABI mapping for '{address}': this deployment requires \
                  signed ABI mappings"
             );
+            rejected_by_policy += 1;
             continue;
         }
 
@@ -165,8 +217,8 @@ pub fn try_extract_from_chain_metadata(
                     abi_kind,
                     implementation,
                 );
-                if is_unsigned {
-                    unsigned_count += 1;
+                if identity_unverified {
+                    unverified_count += 1;
                 }
             }
             Err(e) => {
@@ -174,15 +226,21 @@ pub fn try_extract_from_chain_metadata(
             }
         }
     }
-    if unsigned_count > 0 {
+    if unverified_count > 0 {
         log::warn!(
-            "Accepted {unsigned_count} unsigned ABI mapping(s): integrity/provenance unverified"
+            "Accepted {unverified_count} ABI mapping(s) with no verified signer: provenance unestablished"
         );
     }
-    if registry.list_abis().is_empty() {
-        return None;
+    if rejected_by_policy > 0 {
+        log::warn!(
+            "Refused {rejected_by_policy} ABI mapping(s) under this deployment's trust posture"
+        );
     }
-    Some(registry)
+    AbiExtraction {
+        registry: (!registry.list_abis().is_empty()).then_some(registry),
+        rejected_by_policy,
+        unverified: unverified_count,
+    }
 }
 
 /// Resolve the `AbiKind` and (for proxies) the implementation address from a proto
@@ -983,7 +1041,11 @@ mod tests {
 
     #[test]
     fn test_try_extract_no_metadata() {
-        assert!(try_extract_from_chain_metadata(None, 1, &test_require_signed()).is_none());
+        assert!(
+            try_extract_from_chain_metadata(None, 1, &test_require_signed())
+                .registry
+                .is_none()
+        );
     }
 
     #[test]
@@ -996,7 +1058,9 @@ mod tests {
             })),
         };
         assert!(
-            try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed()).is_none()
+            try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed())
+                .registry
+                .is_none()
         );
     }
 
@@ -1009,7 +1073,9 @@ mod tests {
             })),
         };
         assert!(
-            try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed()).is_none()
+            try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed())
+                .registry
+                .is_none()
         );
     }
 
@@ -1027,6 +1093,7 @@ mod tests {
             })),
         };
         let registry = try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed())
+            .registry
             .expect("should contain ABI");
         assert!(registry.list_abis().contains(&TEST_ADDRESS));
     }
@@ -1059,6 +1126,7 @@ mod tests {
         };
 
         let registry = try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed())
+            .registry
             .expect("tampering with abi_type must NOT invalidate the body signature");
         assert!(
             registry.list_abis().contains(&TEST_ADDRESS),
@@ -1077,6 +1145,7 @@ mod tests {
             1,
             &MetadataTrustPolicy::AcceptUnsigned,
         )
+        .registry
         .expect("unsigned ABI entries must be registered under accept-unsigned");
         assert!(registry.list_abis().contains(&TEST_ADDRESS));
     }
@@ -1089,7 +1158,9 @@ mod tests {
     fn test_try_extract_unsigned_abi_rejected_under_require_signed() {
         let metadata = unsigned_metadata();
         assert!(
-            try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed()).is_none(),
+            try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed())
+                .registry
+                .is_none(),
             "require-signed must reject an unsigned ABI mapping, even though the \
              accept-unsigned posture registers the very same request"
         );
@@ -1124,6 +1195,7 @@ mod tests {
             1,
             &MetadataTrustPolicy::AcceptUnsigned,
         )
+        .registry
         .expect("should contain both entries");
         assert!(
             registry.list_abis().contains(&unsigned_address),
@@ -1159,6 +1231,7 @@ mod tests {
             })),
         };
         let registry = try_extract_from_chain_metadata(Some(&metadata), 1, &test_require_signed())
+            .registry
             .expect("the signed entry must still register");
         assert!(
             !registry.list_abis().contains(&unsigned_address),
@@ -1195,6 +1268,7 @@ mod tests {
                 1,
                 &MetadataTrustPolicy::AcceptUnsigned
             )
+            .registry
             .is_none(),
             "a present-but-invalid signature must be rejected even under accept-unsigned"
         );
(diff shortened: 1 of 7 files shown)


src/chain_parsers/visualsign-ethereum/src/lib.rs
```diff
@@ -165,60 +165,65 @@ impl EthereumTransactionWrapper {
...
     pub fn new() -> Self {
-        let (contract_registry, visualizer_builder) =
-            registry::ContractRegistry::with_default_protocols();
-        Self {
-            registry: Arc::new(contract_registry),
-            visualizer_registry: visualizer_builder.build(),
-            abi_signers: abi_metadata::authorized_abi_signers(),
-        }
+        Self::with_policy(MetadataTrustPolicy::AcceptUnsigned)
     }
```
pepe-anchor, line 227: should: `parser_app` builds its converter through `new()`, so merging this PR alone moves the production TEE binary from rejecting an ABI entry signed by a non-allowlisted key to accepting it, with no flag to opt out until #421 lands and no test asserting which posture the deployed binary runs; pin `create_registry`'s posture in a test, and either land the two PRs together or say in the body that the intermediate state is deliberate.

<details>

`src/parser/app/src/registry.rs:20` is the only production construction site and is unchanged by this PR. On `main`, `new()` set `abi_signers: abi_metadata::authorized_abi_signers()`; `dev-signing` is enabled only by `src/parser/cli/Cargo.toml:10`, and `VISUALSIGN_ETH_ABI_SIGNERS` appears in no workflow, Makefile, or deploy manifest, so that allowlist is empty in production and every signed entry was rejected. After this change `policy.signer_allowlist()` returns `None` and step 7 of `validate_abi_signature` is skipped, so a self-signed entry is accepted.

Repro: one `abi_mappings` entry with a valid signature over the correct prehash under an attacker-chosen key renders as a raw selector on `main` and as a decoded, attacker-labelled call after this PR.

This is not an escalation of attacker capability, since the same render was already reachable on `main` by omitting the signature, which is why the coherence argument in the PR body holds. Two things it does invalidate:

1. **Rollback section:** "Nothing in this PR is observable from outside the library, so nothing needs a redeploy" is false for `parser_app`.
2. **The no-`Default` argument:** `MetadataTrustPolicy` refuses a `Default` so a caller cannot skip the decision, but `new()` and `impl Default for EthereumVisualSignConverter` (lib.rs:288) reinstate the skipped decision one layer up, at exactly the call site production uses. Nothing in the type system or the suite signals that `registry.rs` needs revisiting: it compiles unchanged and green.

Cheapest pin: expose `#[must_use] pub fn accepts_unsigned(&self) -> bool` on the converter and assert it against `parser_app::registry::create_registry`, mirroring the `!super::cli_trust_policy().accepts_unsigned()` assertion this PR added for the CLI.

Also undeclared in the Backwards compatibility section: `with_signers` and `with_registry_and_signers` are removed `pub fn`s, and `with_registry` silently drops identity enforcement for out-of-tree embedders.

</details>

src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
```diff
@@ -107,22 +121,33 @@ pub fn try_extract_from_chain_metadata(
...
                 );
                 continue;
             }
+        } else if !policy.accepts_unsigned() {
+            log::warn!(
+                "Skipping ABI mapping for '{address}': this deployment requires \
+                 signed ABI mappings"
+            );
+            continue;
         }
```
pepe-anchor, line 203: should: This rejection is a `log::warn!` and nothing else, so once #421 flips a deployment to require-signed against a caller that does not sign, every entry is dropped, `try_extract_from_chain_metadata` returns `None`, and the payload is byte-identical to "no ABI was supplied" with no signal anywhere a human looks; carry the rejected count out of this function so the caller can distinguish total rejection from absent metadata.

<details>

The Slack thread that produced PRS-556 (`#proj-open-source-vsp`, 2026-07-01) is explicit about this channel: "logs are helpful but currently turnkey doesn't show the logs, so it'll just get silently ignored". The ticket repeats it under "Why the current warning log isn't enough". This PR adds a second rejection path on that same channel.

It is worse than the ticket assumed: `src/parser/app/Cargo.toml` has no `log` dependency and `src/parser/app/src/main.rs` initialises no logger, so these `warn!` calls are compiled-in no-ops in the deployed binary, not merely unsurfaced.

Concrete sequence:
1. #421 lands, deployment starts with require-signed and one authorized pubkey.
2. A caller sends `abi_mappings` with `signature: None` (what everything except `parser_cli` does today).
3. Every entry hits `continue` at line 150, `registry.list_abis().is_empty()` is true at 182, `None` returns at 183.
4. Rendering falls back to the raw selector. The gRPC response carries no error, no counter, no marker, and the signer sees exactly what a contract with no metadata looks like.

Minimal fix that stays in scope: return the rejected count alongside the registry, or fold it into the aggregated summary already built at lines 177 to 181, so the enclave output can distinguish "no ABI supplied" from "all ABIs refused".

</details>

src/chain_parsers/visualsign-ethereum/src/cli_plugin.rs
```diff
@@ -522,14 +533,22 @@ mod tests {
...
         );
 
+        // Pin the CLI's posture itself. A signed entry survives under either posture
+        // (accept-unsigned still integrity-checks a signature that is present), so the
+        // extraction assertion below cannot tell the two apart. Without this, flipping
+        // `cli_trust_policy` to accept-unsigned would leave the whole suite green.
+        assert!(
+            !super::cli_trust_policy().accepts_unsigned(),
+            "the CLI must register the require-signed posture",
+        );
```
pepe-anchor: should: This asserts on `cli_trust_policy()` in isolation rather than on what `register` installs, so editing line 63 back to `EthereumVisualSignConverter::new()` leaves the entire suite green and silently reverts the CLI to accept-unsigned; assert through the registered converter instead of through the helper.

<details>

Both new assertions route around `register`: line 548 calls the helper directly, and the extraction check at line 562 passes `&super::cli_trust_policy()` straight into `try_extract_from_chain_metadata`. Nothing in the suite observes the converter that `register` actually builds.

Mutation proof: change line 63 to `crate::EthereumVisualSignConverter::new()`. Line 548 still passes because `cli_trust_policy()` is unchanged; line 562 still passes because it never touches the converter. The one production posture change this PR makes is therefore unpinned at its only real call site, which is the same gap commit `cf62a78` was written to close ("nothing asserted what it was ... flipping the CLI to accept-unsigned left the whole suite green").

A test that closes it: build a `TransactionConverterRegistry`, call `EthereumPlugin::new(EthereumArgs::default()).register(&mut registry)`, then convert a transaction whose only decoder is an unsigned metadata ABI and assert it is not decoded. `test_converter_honours_stored_require_signed_posture` in `tests/lib_test.rs` already does the equivalent for the constructors and is a good template.

</details>

src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
```diff
@@ -371,6 +414,31 @@ pub fn authorized_abi_signers() -> SignerAllowlist {
...
+/// startup instead of silently producing a smaller (or empty, fail-closed)
+/// allowlist that quietly drops every caller-supplied ABI.
+///
+/// # Errors
+/// Returns `Err` if `entries` is empty, or if any entry is not a valid secp256k1
+/// public key.
+pub fn signer_allowlist_from_hex(entries: &[String]) -> Result<SignerAllowlist, String> {
+    if entries.is_empty() {
+        return Err("no authorized ABI signer public keys were provided".to_string());
+    }
```
pepe-anchor: should: `signer_allowlist_from_hex` has zero non-test callers in the tree, in a PR whose body says "Library groundwork only, no flags and no deployment changes here", so nobody can review whether its error strings and empty-input rule match #421's clap definition without reading #421; move it there alongside the flag it parses.

<details>

Evidence: `grep -rn signer_allowlist_from_hex --include=*.rs .` matches only the definition, four unit tests in the same file, and two doc references (`abi_metadata.rs:390`, `cli_plugin.rs:48`). Its own doc calls the flag "planned". `Display for MetadataTrustPolicy` (`signing.rs:180-191`) is in the same position: documented as backing "the one-line startup log", pinned by `test_policy_display_names_the_posture`, and emitted by no production code, and `parser_app` has no logger for it to reach.

The empty-input rule also contradicts the type it feeds, three references to one:
- `signing.rs:149-150`: "An empty allowlist rejects everything (fail-closed)."
- `signing.rs` test `test_require_signed_with_empty_allowlist_authorizes_nothing` exercises that state as valid.
- `abi_metadata.rs:383-384`: "An empty result (no dev key, no env entries) rejects all signed ABIs (fail-closed)."
- versus line 430 here, where empty is an error.

So a deployment that wants the strictest posture, reject every caller-supplied ABI, cannot express it through the parser this PR ships. Settle which contract is intended. Secondary: it returns `Err(String)` where the module already has `AbiSignatureError`, and duplicates `authorized_abi_signers`'s parsing with inverted error semantics (warn-and-skip there, hard error here).

</details>

src/visualsign/src/signing.rs
```diff
@@ -120,6 +120,76 @@ impl FromIterator<Vec<u8>> for SignerAllowlist {
     }
 }
 
+/// Deploy-time trust posture for caller-supplied metadata (Ethereum ABI
+/// mappings today; the Solana IDL path still uses a bare [`SignerAllowlist`]).
```
pepe-anchor, line 124: nit: This PR declares "accept unsigned but reject a correctly signed entry from an unlisted signer" strictly worse than either coherent posture, yet `visualsign-solana/src/core/visualsign.rs` keeps that exact pairing and still tells the reader it is at "parity with the Ethereum ABI path", which stops being true with this commit; fix that one comment so a Solana reader does not conclude the pairing is deliberate.

src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
pepe-anchor, line 1: should: `is_unsigned` keys off `signature.is_none()`, but under `AcceptUnsigned` a present signature no longer establishes provenance either, so a self-signed entry is registered without being counted and the aggregated warning reports zero unsigned mappings for a request whose entire decode came from an unverified caller ABI; count "identity was not enforced" rather than "signature absent" when the posture is accept-unsigned.

<details>

This state is newly reachable in production because of this PR: on `main` the production allowlist is empty, so a signed-by-unlisted entry was rejected and never reached the counter. Now it is accepted at line 168 with `is_unsigned == false`.

Fix: `let unverified = policy.signer_allowlist().is_none() || abi.signature.is_none();` and rename the counter and the message at lines 177 to 181 accordingly.

Related staleness in the block you are already rewriting: line 55 says the missing per-entry marker is "tracked in PRS-555", but PRS-555 was closed **Done** on 2026-07-14 and `abi_registry.rs` still carries no verified or unsigned field, so that gap is now pointed at a closed ticket. The PR body's "PRS-555 ... is still open" has the same problem. Whatever inherits this predicate will label attacker-self-signed entries as verified.

</details>

src/chain_parsers/visualsign-ethereum/src/lib.rs
pepe-anchor: Applied, and the PR body now says the intermediate state is deliberate rather than leaving it to be discovered.

`create_registry_runs_the_accept_unsigned_abi_posture` in `parser/app/src/registry.rs` pins the posture the deployed binary runs. Mutation-checked: switching `create_registry` to `with_policy(RequireAllowlistedSigner(..))` turns it red.

One deviation from the suggested fix. I added `accepts_unsigned()` on the converter first and then removed it: the registry stores `Box<dyn VisualSignConverterAny>`, so there is no way to get the converter back out and assert on it. Asserting the accessor against a separately constructed converter reproduces exactly the vacuity you flag in the `cli_plugin` thread, one file over. So the test goes through `create_registry()` and converts a transaction whose only possible decoder is an unsigned caller ABI, which observes the real thing and avoids adding public API with no caller.

Body changes: rollback section now says `parser_app` picks up the posture and a post-deploy revert needs a redeploy; backwards compatibility now declares the removed `with_signers` / `with_registry_and_signers` and that `with_registry`, `new` and `Default` keep compiling while silently resolving to `AcceptUnsigned`.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
pepe-anchor: Applied. `try_extract_from_chain_metadata` now returns `AbiExtraction { registry, rejected_by_policy, unverified }` instead of `Option<AbiRegistry>`.

Confirmed the worse-than-assumed part before changing anything: `parser/app/Cargo.toml` has no `log` dependency and `main.rs` initialises no logger, so those `warn!` calls are compiled-in no-ops in the deployed binary. The doc comment on `AbiExtraction` says so, since that is the reason the count cannot just stay a log line.

Malformed entries (bad address, oversized JSON, unparseable ABI) are counted separately from policy refusals, so the count keeps meaning "this deployment refused you" rather than "something went wrong". `test_total_refusal_is_distinguishable_from_absent_metadata` pins refused-everything (`registry: None`, `rejected_by_policy: 1`) against absent metadata (`registry: None`, `rejected_by_policy: 0`), and is mutation-checked against dropping the increment.

Stopping short of rendering it: that needs a field on the payload, which is the PRS-555 follow-up and out of scope here. `extract_metadata_abi` returns the whole struct so that follow-up has something to read, with a comment at the call site saying why the counts are not consumed yet.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

src/chain_parsers/visualsign-ethereum/src/cli_plugin.rs
pepe-anchor: Applied, using the test you sketched. `test_register_installs_require_signed_posture` builds a `TransactionConverterRegistry`, calls `EthereumPlugin::new(EthereumArgs::default()).register(&mut registry)`, converts a transaction whose only decoder is an unsigned metadata ABI, and asserts it renders the raw selector rather than `customFoo`. An accept-unsigned registry converts the same fixture as a control, so the refusal cannot be a false negative.

Ran your mutation proof: changing line 63 back to `EthereumVisualSignConverter::new()` now fails that test, and only that test.

The isolated `cli_trust_policy()` assertion is gone. The extraction check in `test_proxy_mapping_synthesizes_entry_when_no_abi_file` still passes `&cli_trust_policy()`, but its comment now says it is about proxy synthesis and points at the new test for the posture, since a signed entry survives either posture and that assertion cannot tell them apart.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
pepe-anchor: Applied, removed here so it lands with the flag in #421. Confirmed your grep: definition, four unit tests, two doc references, no production caller.

That also settles the contract question by deletion rather than by decision, which I think is the right order. Empty-is-an-error contradicted three places that document empty as fail-closed (`signing.rs:149-150`, `test_require_signed_with_empty_allowlist_authorizes_nothing`, `abi_metadata.rs:383-384`), and the deployment that wants "refuse every caller-supplied ABI" is a real posture that the parser shipped here could not express. Settling it where the clap definition lives means the error strings and the empty rule get reviewed against the flag they belong to.

Kept `Display for MetadataTrustPolicy`. It is a trait impl on the type this PR introduces rather than a parser for a flag that does not exist yet, it carries no contradictory contract, and it costs nothing if the startup log lands late. Happy to pull it too if you would rather this PR ship only what has a caller.

Doc references at `abi_metadata.rs:390` and `cli_plugin.rs:48` no longer name the function; they say the flag and its parser land together in the follow-up.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

src/visualsign/src/signing.rs
pepe-anchor: Fixed. There were three of them, not one:

- `visualsign.rs:142`, the "Mirrors `try_extract_from_chain_metadata`" header, now says it is structured after it but no longer at parity on trust posture.
- `visualsign.rs:258`, the step 4 inline comment, now says the pairing is a known gap rather than parity.
- `visualsign.rs:1785`, the doc on `test_extract_idl_mappings_accepts_unsigned_idl`, which made the same claim.

The security notes on `extract_idl_mappings` now spell out why the pairing is worse than either coherent posture and point at `MetadataTrustPolicy`, so the next person to read it gets the reasoning rather than just "known gap". Comments only, no behaviour change here.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

src/chain_parsers/visualsign-ethereum/src/abi_metadata.rs
pepe-anchor: Applied, with your predicate: `let identity_unverified = policy.signer_allowlist().is_none() || abi.signature.is_none();`. Counter and aggregated message renamed to match ("Accepted N ABI mapping(s) with no verified signer: provenance unestablished"), and it is surfaced as `AbiExtraction::unverified` rather than only logged.

`test_accept_unsigned_counts_self_signed_entry_as_unverified` signs with `FOREIGN_SIGNER_SEED` so the dev key does not mask it, and is mutation-checked: reverting to `abi.signature.is_none()` turns it red. `test_require_signed_registers_nothing_unverified` is the counterpart, so a predicate that counted every accepted entry unconditionally does not pass either.

PRS-555 staleness fixed in both places. Confirmed it is closed Done, 2026-07-14, and that `abi_registry.rs` still carries no verified or unsigned field, so the gap is real with no open ticket behind it. The doc comment now describes the missing per-entry marker without pointing at a ticket, and the PR body says the same instead of "still open".

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

shahan-khatchadourian-anchorage (approved):
Solid, well-scoped security fix. `MetadataTrustPolicy` correctly makes the trust decision explicit (no `Default`, deliberately), and the core judgment — an identity check an attacker can bypass by just dropping the signature buys nothing — is sound. `AbiExtraction` fixing the "total refusal looks identical to no metadata" gap is a good catch.

Test coverage is thorough and clearly mutation-checked: posture divergence on identical input, self-signed-but-unlisted counted as unverified, malformed input not conflated with policy rejection, and end-to-end pinning of both `parser_cli`'s and `parser_app`'s actual registered posture (not just the helper functions) — this closes exactly the gaps flagged in the earlier self-review threads.

No dangling references to the removed `with_signers`/`with_registry_and_signers` APIs anywhere in the tree. Solana IDL path is correctly left untouched with a doc-only correction rather than scope creep.

CI green, no correctness issues found. Approving.

prasanna-anchorage (approved):

PR #423, opened 2026-07-28
fix(smoke): correct the container key mount, support prod, drop org ids

pepe-anchor:
## Why

`scripts/smoke.sh` could not run as written, and the failure got worse the day the client image was published.

The container invocation mounted the Turnkey key pair at `/root/.config/turnkey/keys`, but the published `visualsign-turnkeyclient` image runs as `nonroot` with `HOME=/home/nonroot`, and the client resolves its keys from `$HOME`. Client resolution prefers the GHCR container over any local binary, so once the image went live every container run died with `failed to read public key file` and exited 2.

The verify call also hardcoded `--dev-path`, so the script could only smoke dev and staging (they share one Turnkey app and org). Prod had no harness at all: step 4 of the deploy runbook was a raw client invocation with the JSON eyeballed, so none of the render/verify assertions and none of the outage-vs-regression exit-code classification applied. That is exactly the gap that let the mount bug sit unnoticed.

Last, the dev org id was baked in as a default. This repo is public, and app ids and operator ids already come from CI secrets rather than the tree, so that default was the outlier.

## What

- Mount the keys at the image's `HOME` (`/home/nonroot/...`) instead of `/root`.
- Add `--target dev|prod` (`VSP_SMOKE_TARGET`), defaulting to `dev`, which covers staging. The target picks the endpoint path (`--dev-path` -> `/visualsign-dev`, or the canonical `/visualsign`) and the default key name (`dev`, `default`). An unrecognized target is a usage error (exit 2), never a silent fall back to dev, which would smoke the wrong deployment and report it as a pass.
- Take org ids from the environment for both targets: `VSP_SMOKE_ORG_DEV` / `VSP_SMOKE_ORG_PROD`, with `VSP_SMOKE_ORG` winning for whichever target is selected (that is what CI passes, from `secrets.TVC_ORG_ID`). A missing id exits 2 rather than smoking whatever org happens to be configured.
- Announce `target`/`org`/`key` at the start of a run, and carry the target into the `PASS:` line, so a prod run is self-evidencing in CI logs.
- `.gitignore`: add `.claude/runbooks/` and `.claude/worktrees/`, per-agent scratch dirs that must never land in a commit (same reasoning as the existing `.claude/plans/` entry).

CI is unaffected: `.github/workflows/tvc-deploy.yml` calls bare `./scripts/smoke.sh` and supplies `VSP_SMOKE_ORG`, so it still resolves to the dev target.

The deploy runbook is updated to match: both step 4s now lead with `scripts/smoke.sh` (`--target prod` for prod), the raw `$CLIENT verify` form is kept only for flags the script does not pass (`--debug`, `--include-intermediate-output`, another `--chain`), and the required org env vars are listed in the prerequisites.

## Testing

Dev smoke against the live endpoint, on this branch (org id redacted here, it is in the diff as a deletion):

```
$ ./scripts/smoke.sh
smoking target=dev org=<dev org> key=dev
✓ Attestation document verified successfully
✓ PCRs verified: 32 PCRs found
✓ Raw manifest hash matches UserData in attestation
✓ Signature verified successfully
PASS: turnkey-client verify succeeded on target=dev (org=<dev org>); V0+ALT rendered
(17622 chars, no "Cannot render V0"); attestation + signature cryptographically verified
(executed in AWS Nitro enclave); moduleId=i-0275639cc084e73c8-enc019fa971401dc03c
$ echo $?
0
```

Prod argv checked through a stub client, so no prod traffic. Correctly drops `--dev-path` and picks the prod org and key:

```
ARGV: verify --host https://api.turnkey.com --organization-id <prod org> --key-name default \
  --unsigned-payload <payload> --chain CHAIN_SOLANA
```

Guardrails:

```
$ env -u BASH_ENV -u VSP_SMOKE_ORG -u VSP_SMOKE_ORG_DEV ./scripts/smoke.sh
ERROR: no organization id for --target dev; set VSP_SMOKE_ORG_DEV (or VSP_SMOKE_ORG). ...  # exit 2
$ ./scripts/smoke.sh --target staging
--target must be dev or prod (got: staging)                                                # exit 2
```

`VSP_SMOKE_ORG_DEV` set alone does not satisfy `--target prod` (no cross-target leak), and the CI shape (`VSP_SMOKE_ORG` only) still resolves. `shellcheck` and `bash -n` are clean.

**Not verified here:** the prod endpoint itself. That needs the prod API key, so the prod path is proven at the argv level only. First prod use should be a deliberate run by someone holding that key.

## Backwards compatibility

Backwards compatible. `--target` defaults to `dev`, which is the previous behavior, and the only in-repo caller (the CI smoke step) passes no target. The one new requirement is that the dev org id now comes from the environment; CI already provides it via `VSP_SMOKE_ORG`, and local runs need `VSP_SMOKE_ORG_DEV` exported (see the runbook prerequisites).

## Rollback

Revert the commits. `scripts/smoke.sh` is a developer/CI script that only reads from the deployed parser, so there is no production behavior to unwind and no data migration. Reverting restores the broken `/root` mount and the dev-only hardcoded org, so prefer a forward fix.

## Linear ticket

None; found while running the weekly dev/staging deploy smoke.

### DIFF
--- a/.github/workflows/tvc-deploy.yml
+++ b/.github/workflows/tvc-deploy.yml
@@ -130,8 +130,7 @@ jobs:
       # cryptographically VERIFIES (AWS Nitro attestation + enclave signature).
       # The API key is reconstructed from the same TVC_API_KEY_* secrets. Pins an
       # approved turnkey-client image version rather than :latest.
-      # NOTE: requires the turnkey-client image on GHCR (visualsign-turnkeyclient
-      # docker.yml / #23); until that's published this step can't pull the image.
+      # smoke.sh defaults to --target dev, which is what this workflow deploys.
       - name: Smoke (dev verify endpoint)
         if: success()
         env:
--- a/.github/workflows/tvc-deploy.yml
+++ b/.github/workflows/tvc-deploy.yml
@@ -130,7 +130,8 @@ jobs:
       # cryptographically VERIFIES (AWS Nitro attestation + enclave signature).
       # The API key is reconstructed from the same TVC_API_KEY_* secrets. Pins an
       # approved turnkey-client image version rather than :latest.
-      # smoke.sh defaults to --target dev, which is what this workflow deploys.
+      # Passes --target dev explicitly so the invariant holds here rather than
+      # resting on the script default and on VSP_SMOKE_TARGET being absent.
       - name: Smoke (dev verify endpoint)
         if: success()
         env:
@@ -146,7 +147,7 @@ jobs:
           mkdir -p "$HOME/.config/turnkey/keys"
           printf '%s' "$TVC_API_KEY_PUBLIC"       > "$HOME/.config/turnkey/keys/dev.public"
           printf '%s:p256' "$TVC_API_KEY_PRIVATE" > "$HOME/.config/turnkey/keys/dev.private"
-          ./scripts/smoke.sh
+          ./scripts/smoke.sh --target dev
 
       - name: Scrub secrets
         if: always()
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,10 @@ docs/superpowers/
 .claude/shell-snapshots/
 .claude/statsig/
 .claude/settings.local.json
+.claude/runbooks/
 
 # Coverage artifacts (see TESTING.md)
 lcov.info
 *.profraw
 *.profdata
+.claude/worktrees/
--- a/.gitignore
+++ b/.gitignore
@@ -14,9 +14,9 @@ docs/superpowers/
 .claude/statsig/
 .claude/settings.local.json
 .claude/runbooks/
+.claude/worktrees/
 
 # Coverage artifacts (see TESTING.md)
 lcov.info
 *.profraw
 *.profdata
-.claude/worktrees/
--- a/scripts/smoke.sh
+++ b/scripts/smoke.sh
@@ -132,10 +132,15 @@ if [ -z "$ORG" ]; then
   exit 2
 fi
 IMAGE="ghcr.io/anchorageoss/visualsign-turnkeyclient:${CLIENT_VERSION}"
-# The published image runs as `nonroot` with HOME=/home/nonroot, and the client
-# resolves the key pair from $HOME/.config/turnkey/keys, so the mount target has
-# to match that HOME (not /root, which the client never looks in).
-CONTAINER_CLIENT="docker run --rm -v $HOME/.config/turnkey/keys:/home/nonroot/.config/turnkey/keys:ro $IMAGE"
+# The client resolves its key pair from $HOME/.config/turnkey/keys, so the mount
+# target has to match the container's HOME (not /root, which it never looks in).
+# Two further wrinkles, both of which this invocation handles:
+#   - run as the invoking uid, so a key file with sane 0600 permissions is
+#     readable. The image's own user is `nonroot`, which cannot read a key owned
+#     by you unless you widen it to 0644; a prod credential should not need that.
+#   - mount under a neutral HOME rather than the image's /home/nonroot, which is
+#     not traversable by any other uid, so --user alone would still hit EACCES.
+CONTAINER_CLIENT="docker run --rm --user $(id -u):$(id -g) -e HOME=/tkhome -v $HOME/.config/turnkey/keys:/tkhome/.config/turnkey/keys:ro $IMAGE"
 
 # Resolve a local fallback path to a runnable client: an executable is used
 # directly; a directory is treated as the turnkey-client source and built
--- a/scripts/smoke.sh
+++ b/scripts/smoke.sh
@@ -42,13 +42,17 @@
 # Env (the target's org id is REQUIRED, the rest are optional):
 #   VSP_SMOKE_ORG_DEV             org id for --target dev (dev + staging share it)
 #   VSP_SMOKE_ORG_PROD            org id for --target prod
-#   VSP_SMOKE_ORG                 org id for whichever target is selected; wins
-#                                 over the two above (this is what CI passes)
+#   VSP_SMOKE_ORG                 org id fallback for whichever target is selected,
+#                                 used only when that target's own var above is
+#                                 unset (this is what CI passes). The per-target
+#                                 vars win so a VSP_SMOKE_ORG left exported from a
+#                                 dev run cannot point the dev org at prod.
 #   VSP_SMOKE_HOST                 API host    (default https://api.turnkey.com)
 #   VSP_SMOKE_TARGET              dev|prod (same as --target)
-#   VSP_SMOKE_KEY                 key name under ~/.config/turnkey/keys/<key>.{public,private}
-#                                 for whichever target is selected; wins over the
-#                                 two below
+#   VSP_SMOKE_KEY                 key name fallback under
+#                                 ~/.config/turnkey/keys/<key>.{public,private},
+#                                 used only when the selected target's own key var
+#                                 below is unset
 #   VSP_SMOKE_KEY_DEV             key name for --target dev   (default: dev)
 #   VSP_SMOKE_KEY_PROD            key name for --target prod  (default: default)
 #
@@ -75,8 +79,9 @@
 set -euo pipefail
 
 HOST="${VSP_SMOKE_HOST:-https://api.turnkey.com}"
-# Org/key defaults depend on --target, so they are resolved after arg parsing;
-# an explicit env value still wins over the target's default.
+# Org/key defaults depend on --target, so they are resolved after arg parsing.
+# These two are only the target-agnostic fallback: the selected target's own
+# VSP_SMOKE_{ORG,KEY}_{DEV,PROD} wins over them.
 ORG="${VSP_SMOKE_ORG:-}"
 KEY="${VSP_SMOKE_KEY:-}"
 TARGET="${VSP_SMOKE_TARGET:-dev}"
@@ -112,15 +117,21 @@ done
 #
 # Org ids are NOT baked in. This repo is public and the ids identify Anchorage's
 # Turnkey orgs, so they come from the environment (CI secrets, or a private local
-# env file); the deploy runbook lists which id goes with which target.
+# env file); the parser_app deploy runbook in Notion lists which id goes with
+# which target.
+#
+# The target's own var wins over the target-agnostic VSP_SMOKE_ORG /
+# VSP_SMOKE_KEY: an operator who keeps those exported for dev runs would
+# otherwise point the dev org and dev key at prod's canonical /visualsign
+# endpoint, and the resulting failure would read as a bad deploy.
 case "$TARGET" in
   dev)
-    ORG="${ORG:-${VSP_SMOKE_ORG_DEV:-}}"
-    KEY="${KEY:-${VSP_SMOKE_KEY_DEV:-dev}}"
+    ORG="${VSP_SMOKE_ORG_DEV:-$ORG}"
+    KEY="${VSP_SMOKE_KEY_DEV:-${KEY:-dev}}"
     PATH_ARGS=(--dev-path) ;;
   prod)
-    ORG="${ORG:-${VSP_SMOKE_ORG_PROD:-}}"
-    KEY="${KEY:-${VSP_SMOKE_KEY_PROD:-default}}"
+    ORG="${VSP_SMOKE_ORG_PROD:-$ORG}"
+    KEY="${VSP_SMOKE_KEY_PROD:-${KEY:-default}}"
     PATH_ARGS=() ;;
   *) echo "--target must be dev or prod (got: $TARGET)" >&2; exit 2 ;;
 esac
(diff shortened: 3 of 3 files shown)


scripts/smoke.sh
```diff
@@ -122,8 +186,9 @@ PAYLOAD="$(tr -d '[:space:]' < "$DIR/testdata/solana_v0_alt.b64")"
 ERRFILE="$(mktemp)"
 trap 'rm -f "$ERRFILE"' EXIT
 
+echo "smoking target=$TARGET org=$ORG key=$KEY" >&2
 set +e
-OUT="$($CLIENT verify --dev-path --host "$HOST" --organization-id "$ORG" \
+OUT="$($CLIENT verify "${PATH_ARGS[@]}" --host "$HOST" --organization-id "$ORG" \
```
pepe-anchor: must: Expanding the empty `PATH_ARGS` array under `set -u` is an unbound-variable error on bash before 4.4, so `--target prod`, the only new execution path in this PR, dies before the client is ever invoked for any operator whose bash is macOS stock /bin/bash (3.2.57).

<details>

Reproduced with the real script against the `bash:3.2` image, which is what macOS ships and what `#!/usr/bin/env bash` resolves to unless a newer bash sits earlier in PATH:

```
$ docker run --rm -e TURNKEY_CLIENT=/bin/echo -e VSP_SMOKE_ORG_PROD=fake-org \
    -v "$PWD":/w -w /w bash:3.2 ./scripts/smoke.sh --target prod
smoking target=prod org=fake-org key=default
/w/scripts/smoke.sh: line 192: PATH_ARGS[@]: unbound variable
ERROR: smoke could not run the turnkey-client; this is not an endpoint outage
```

`--target dev` is unaffected because `PATH_ARGS=(--dev-path)` is non-empty, which is why the dev run in the testing section passes, and neither `shellcheck` nor `bash -n` catches it (both are clean on my machine too). The exit-code classification stays honest (2, not a false pass), but prod never gets smoked, and prod is exactly the path with no automated coverage.

I verified the suggested one-line guard: shellcheck rc=0, `bash -n` clean, and on `bash:3.2` the prod path now clears the expansion and reaches the client. Folding the subcommand into the array instead (`VERIFY_ARGS=(verify --dev-path)` for dev, `VERIFY_ARGS=(verify)` for prod) is equivalent if you prefer how it reads.

</details>

```suggestion
OUT="$($CLIENT verify ${PATH_ARGS[@]+"${PATH_ARGS[@]}"} --host "$HOST" --organization-id "$ORG" \
```

scripts/smoke.sh
```diff
@@ -73,13 +99,51 @@ while [ "$#" -gt 0 ]; do
...
+# Org ids are NOT baked in. This repo is public and the ids identify Anchorage's
+# Turnkey orgs, so they come from the environment (CI secrets, or a private local
+# env file); the deploy runbook lists which id goes with which target.
+case "$TARGET" in
+  dev)
+    ORG="${ORG:-${VSP_SMOKE_ORG_DEV:-}}"
+    KEY="${KEY:-${VSP_SMOKE_KEY_DEV:-dev}}"
+    PATH_ARGS=(--dev-path) ;;
+  prod)
+    ORG="${ORG:-${VSP_SMOKE_ORG_PROD:-}}"
```
pepe-anchor: should: `VSP_SMOKE_ORG` and `VSP_SMOKE_KEY` are target-agnostic and win for every target, so an operator who keeps them exported for dev runs and then runs `--target prod` points the dev org and dev key at the canonical /visualsign endpoint: prod is never smoked, and the resulting failure is attributed to the deployment rather than to the config.

<details>

Evidence that this shell exists: the testing section had to run `env -u BASH_ENV -u VSP_SMOKE_ORG -u VSP_SMOKE_ORG_DEV ./scripts/smoke.sh` to trigger the missing-org error, so `VSP_SMOKE_ORG` is exported in the very environment prod smokes will run from. The guardrail that was tested (`VSP_SMOKE_ORG_DEV` alone not satisfying prod) does not cover this direction, and the only signal is the new `smoking target=... org=...` line. Depending on the client error text the run then exits 1 as `FAIL: deployed parser rejected a tx it should render (regression)` (line 204) or 2, immediately after a prod set-live.

The header comment at lines 31 to 33 already names this hazard, which argues for enforcing it rather than documenting it. Minimal fix, and CI is unaffected because it sets only `VSP_SMOKE_ORG`:

```
  dev)  ORG="${VSP_SMOKE_ORG_DEV:-$ORG}";  KEY="${VSP_SMOKE_KEY_DEV:-${KEY:-dev}}" ;;
  prod) ORG="${VSP_SMOKE_ORG_PROD:-$ORG}"; KEY="${VSP_SMOKE_KEY_PROD:-${KEY:-default}}" ;;
```

</details>

scripts/smoke.sh
```diff
@@ -73,13 +99,51 @@ while [ "$#" -gt 0 ]; do
...
+  *) echo "--target must be dev or prod (got: $TARGET)" >&2; exit 2 ;;
+esac
+if [ -z "$ORG" ]; then
+  # Exit 2, not 1: nothing was smoked, so this is a harness/config failure and
+  # must not be reportable as either a pass or a parser regression.
+  case "$TARGET" in
+    dev) var=VSP_SMOKE_ORG_DEV ;;
+    *) var=VSP_SMOKE_ORG_PROD ;;
+  esac
+  echo "ERROR: no organization id for --target $TARGET; set $var (or VSP_SMOKE_ORG). See the parser_app deploy runbook for the id." >&2
```
pepe-anchor: should: The one failure a fresh operator is guaranteed to hit tells them to see the parser_app deploy runbook for the org id, but no runbook exists in the tree and this same PR adds `.claude/runbooks/` to `.gitignore`, so the pointer resolves to nothing for anyone who is not the author.

<details>

`git ls-files | grep -i runbook` returns nothing, and the word appears only in the three files this PR touches (`.gitignore`, `.github/workflows/tvc-deploy.yml:10`, `scripts/smoke.sh:115` and `:134`). `.claude/runbooks/` was never tracked (`git log -- .claude/runbooks/*` is empty). The PR body also says the deploy runbook is updated to match, but no doc appears in the diff, so that claim is unreviewable.

The runbook does exist in Notion and already carries the org-to-target mapping, so the cheap fix is to make the pointer resolvable rather than to move the doc: name Notion explicitly in the message (and in the line 115 comment), or state the private env-file path the operator is expected to have.

</details>

scripts/smoke.sh
```diff
@@ -122,8 +186,9 @@ PAYLOAD="$(tr -d '[:space:]' < "$DIR/testdata/solana_v0_alt.b64")"
 ERRFILE="$(mktemp)"
 trap 'rm -f "$ERRFILE"' EXIT
 
+echo "smoking target=$TARGET org=$ORG key=$KEY" >&2
```
pepe-anchor, line 200: nit: The start line and the PASS line now carry the target, but the outage skip at line 216 still hardcodes dev and exits 0, so a prod smoke that skips names the wrong endpoint in the log an on-call reads.

<details>

Line 216 is outside this diff, hence the anchor on the new target echo. Use `$TARGET` there as the other three messages now do.

</details>

.github/workflows/tvc-deploy.yml
```diff
@@ -130,8 +130,7 @@ jobs:
       # cryptographically VERIFIES (AWS Nitro attestation + enclave signature).
       # The API key is reconstructed from the same TVC_API_KEY_* secrets. Pins an
       # approved turnkey-client image version rather than :latest.
-      # NOTE: requires the turnkey-client image on GHCR (visualsign-turnkeyclient
-      # docker.yml / #23); until that's published this step can't pull the image.
+      # smoke.sh defaults to --target dev, which is what this workflow deploys.
```
pepe-anchor: nit: The new comment asserts that the workflow deploys dev and the smoke step checks dev, but line 149 calls bare `./scripts/smoke.sh`, so the invariant rests on the script default and on `VSP_SMOKE_TARGET` being absent from the runner environment.

<details>

Change line 149 to `./scripts/smoke.sh --target dev` and the comment becomes self-enforcing at no cost. Not reachable as a false pass today: the deploy authenticates with the dev-scoped `secrets.TVC_ORG_ID` and `TVC_API_KEY_*` and the smoke step is `if: success()`, so a non-dev `app_id` fails the deploy first.

</details>

.gitignore
```diff
@@ -13,8 +13,10 @@ docs/superpowers/
 .claude/shell-snapshots/
 .claude/statsig/
 .claude/settings.local.json
+.claude/runbooks/
 
 # Coverage artifacts (see TESTING.md)
 lcov.info
 *.profraw
 *.profdata
+.claude/worktrees/
```
pepe-anchor: nit: `.claude/worktrees/` is agent scratch, not a coverage artifact, so it sits under the wrong section header and away from `.claude/runbooks/` added six lines earlier in the same PR; both entries are also unrelated to the smoke fix this PR is about.

<details>

Move it up next to line 16, under the existing `.claude/` block whose comment already carries the rationale.

</details>

scripts/smoke.sh
pepe-anchor: Fixed, and confirmed against `bash:3.2` the same way you reproduced it. Went with the one-line guard rather than folding the subcommand into the array, since it keeps the change to the line that was broken:

```
OUT="$($CLIENT verify ${PATH_ARGS[@]+"${PATH_ARGS[@]}"} --host "$HOST" ...
```

`scripts/smoke.sh:202`. Verification on `bash:3.2`, prod path now clears the expansion and reaches the client (it only fails afterwards because that image has no `jq`):

```
$ docker run --rm -e TURNKEY_CLIENT=/bin/echo -e VSP_SMOKE_ORG_PROD=fake-org \
    -v "$PWD":/w -w /w bash:3.2 ./scripts/smoke.sh --target prod
smoking target=prod org=fake-org key=default
/w/scripts/smoke.sh: line 253: jq: command not found
```

No more `PATH_ARGS[@]: unbound variable`. Dev is unchanged and still passes `--dev-path`. shellcheck rc=0, `bash -n` clean.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

scripts/smoke.sh
pepe-anchor: Applied your fix, the target's own var now wins and the generic one is the fallback (`scripts/smoke.sh:126-135`). Also updated the header block and the inline comment, which documented the old precedence.

Worth noting your premise held on my machine too: a test run with only `VSP_SMOKE_ORG` set still picked up the real dev org, because `VSP_SMOKE_ORG_DEV` is exported in my shell. That is exactly the environment prod smokes would run from.

Verified with a stub client:

- stale `VSP_SMOKE_ORG`/`VSP_SMOKE_KEY` for dev + `--target prod` with the prod vars set -> `smoking target=prod org=prod-org key=prod-key`
- CI shape (clean env, only `VSP_SMOKE_ORG`) -> `org=ci-org key=dev` for dev, `org=ci-org key=default` for prod, so CI is unaffected

One thing this does not close: with *only* `VSP_SMOKE_ORG` set, `--target prod` still uses it. Enforcing per-target ids outright would break the CI invocation, so I left it as the documented hazard.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

scripts/smoke.sh
pepe-anchor: Right, the pointer was dangling. Made it resolvable rather than moving the doc, as you suggested, naming Notion in both places:

- `scripts/smoke.sh:145` (the error an operator actually hits): `The id is in the parser_app deploy runbook in Notion (search "parser_app deploy"), or in your private local env file.`
- `scripts/smoke.sh:118-121` (the comment): `the parser_app deploy runbook in Notion lists which id goes with which target`

The PR body claim about the runbook being updated is about the Notion doc, not a tracked file, so it stays unreviewable here. Fair hit.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

scripts/smoke.sh
pepe-anchor: Fixed, the skip now uses `$TARGET` like the other three (`scripts/smoke.sh:230`). Dropped the em dash in that string while I was on the line, for the ASCII-only rule.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

.github/workflows/tvc-deploy.yml
pepe-anchor: Done, line 150 is now `./scripts/smoke.sh --target dev` and the comment says so instead of asserting the default. Agreed it was not a reachable false pass, just an invariant that did not enforce itself.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

.gitignore
pepe-anchor: Moved it up next to `.claude/runbooks/` under the `.claude/` block, so it inherits that comment's rationale and is out of the coverage section. `git check-ignore -v` confirms both still match (`.gitignore:16` and `:17`).

Leaving both entries in this PR rather than splitting them out, they are one-line ignores and pulling them into their own PR costs more than it saves.

---
*Addressed by Claude, applied. React 👎 if this doesn't land and I'll flag it for follow-up.*

prasanna-anchorage (approved):

PR #424, opened 2026-07-29
fix(solana): report unsupported transaction versions instead of a bincode EOF

pepe-anchor:
## Why

[SIMD-0296](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0296-larger-transactions.md) raises Solana's max transaction size from 1232 to 4096 bytes, but only for the new `v1` format specced separately in [SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md). Both are `status: Review` with no feature key assigned. We support legacy and v0 only, so the question was whether a `v1` payload could reach a user rendered under a wrong interpretation.

It can't: a `v1` envelope fails both of our decoders. But the error we returned for that (and for any unsupported message version) was `Decode error: io error: unexpected end of file`, which points at nothing. `from_string` discarded solana-sdk's precise diagnostic with `if let Ok(..)` and reported the legacy decoder's failure instead, after that decoder had already misread the version prefix as a header byte.

## What

Both decoders run first. Only if both fail do we inspect the payload, and then only the leading byte:

- If it is the `v1` marker (`0x80 | 1`), return `UnsupportedVersion` naming SIMD-0385.
- Otherwise return `DecodeError` carrying the **versioned** decoder's error verbatim. solana-sdk already emits `invalid value: integer `N`, expected a valid transaction message version` for an unknown version and `off-chain messages are not accepted` for version 127, so there is nothing left to reconstruct.

No behaviour change for legacy or v0: the only payloads whose outcome changes are ones that already failed.

### Two defects this PR went through before landing here

Worth recording, because the first version of this change got the reasoning wrong and review caught it empirically.

**Ordering.** The `v1` marker check originally ran *before* either decoder, justified by a comment claiming no legacy/v0 transaction could produce a leading `0x81`, on the grounds that doing so needs ≥129 signatures (≥8256 bytes) and Solana caps transactions at 1232. That cap is a **network** rule this parser does not enforce (our limits are 10 MB CLI / 25 MB gRPC). A 129-signature legacy transaction decodes fine here and was being rejected as `v1`, regressing a payload that previously worked. Running the check last removes the need for the argument entirely.

**Byte-walking version probe.** A helper located the message version byte to name unsupported versions. `decode_shortu16_len` succeeds for any first byte below `0x80` regardless of whether the input is envelope-shaped, so random bytes could land on an offset whose byte had the high bit set and be reported as a specific unsupported version. A fix for misleading errors that produced its own. Deleted in favour of solana-sdk's error text.

### Accepted residual

Garbage that starts with `0x81` and fails both decoders is still labeled `v1` (e.g. a *truncated* 129-signature legacy tx). It is rejected either way, so this is diagnostic imprecision, never a wrong accept, and much narrower than what it replaces.

### Out of scope

Implementing `v1`. It is a new wire format (fixed-width counts, trailing signatures, no address lookup tables) and its `TransactionConfigMask` moves priority fee / CU limit / heap into a header bitmask while ComputeBudget instructions become ignored-but-executed no-ops. That last part is a display trap for our `compute_budget` preset (a `v1` tx could show a fee the validator ignores) and deserves its own change once SIMD-0385 is accepted.

## Rollback

Revert the two commits. The change is confined to `SolanaTransactionWrapper::from_string` plus its tests in one file; nothing persists state or changes a wire format.

## Backwards compatibility

Compatible. Legacy and v0 parse exactly as before, pinned by tests including the 129-signature case. The error *variant* for some already-failing payloads changes; nothing in the workspace matches on `TransactionParseError` variants, both stringify through the same gRPC error path.

## Test evidence

`unsupported_versions` module in `core/visualsign.rs`:
- `rejects_simd385_v1_with_an_unsupported_version_error`
- `rejects_v1_across_the_whole_legal_envelope` — sweeps the SIMD-0385 caps (1-12 signatures, 2-64 addresses, 0-3000 bytes instruction data, up to the 4096-byte limit); payloads traverse both decoders rather than short-circuiting at byte 0
- `accepts_legacy_with_129_signatures_despite_leading_v1_marker_byte` — the regression from the ordering defect
- `reports_the_version_for_an_unknown_versioned_message_prefix` — asserts solana-sdk's wording survives
- `names_version_127_as_an_off_chain_message`
- `still_accepts_v0`

```
cargo test -p visualsign-solana            # 328 passed, 0 failed (9 binaries)
cargo clippy --all-targets -- -D warnings  # exit 0, workspace-wide
cargo fmt --check                          # clean
```

Reviewed at depth=high (quick-review + pr-review + deep-review) over three quality-loop iterations. `deep-review` verified the ordering fix against `solana-transaction`/`solana-message` source and with a standalone probe crate built on the pinned dependency versions, confirming `bytes.first()` is the outer short-vec signature-count byte and that returning the versioned error hides no better legacy-specific diagnostic.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/visualsign.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/visualsign.rs
@@ -13,8 +13,9 @@ use crate::idl::signature::{
 };
 use base64::{self, Engine};
 use solana_sdk::{
-    message::VersionedMessage,
+    message::{MESSAGE_VERSION_PREFIX, VersionedMessage},
     pubkey::Pubkey,
+    short_vec::decode_shortu16_len,
     transaction::{Transaction as SolanaTransaction, VersionedTransaction},
 };
 use std::collections::BTreeMap;
@@ -37,6 +38,34 @@ use crate::intermediate::extract_solana_intermediate_output;
 /// loading has a wider 10 MB cap, but proto-supplied IDLs arrive per-request
 /// and are deserialized on the hot path, so we apply a tighter bound here.
 const MAX_IDL_JSON_BYTES: usize = 1_024 * 1_024;
+
+/// Leading byte of a SIMD-0385 `v1` transaction (`0x80 | 1`).
+///
+/// `v1` puts its `VersionByte` first and its signatures last, so unlike
+/// legacy/v0 this byte sits at offset 0 of the payload. Legacy and v0 both
+/// start with a compact-u16 signature count instead, and `0x81` there decodes
+/// to `1 + 128 * next_byte` (>= 129) signatures, which needs >= 8256 bytes of
+/// signature data. No legacy/v0 transaction can be that large, so this byte is
+/// an unambiguous `v1` marker rather than a heuristic.
+const TRANSACTION_V1_VERSION_BYTE: u8 = 129;
+
+/// Read the message version prefix of a legacy/v0-framed payload, returning the
+/// version number when it is one we cannot decode.
+///
+/// The envelope is `[compact-u16 signature count][64-byte signatures..][message]`
+/// and a versioned message leads with `MESSAGE_VERSION_PREFIX | version`. Returns
+/// `None` for legacy (no high bit) and for v0, which we both support, and for
+/// payloads too short or malformed to locate the byte in.
+fn unsupported_message_version(bytes: &[u8]) -> Option<u8> {
+    let (sig_count, prefix_len) = decode_shortu16_len(bytes).ok()?;
+    let message_start = prefix_len.checked_add(sig_count.checked_mul(64)?)?;
+    let version_byte = *bytes.get(message_start)?;
+    if version_byte & MESSAGE_VERSION_PREFIX == 0 {
+        return None;
+    }
+    let version = version_byte & !MESSAGE_VERSION_PREFIX;
+    (version != 0).then_some(version)
+}
 
 /// Append decode errors as diagnostics and lint diagnostics to the output fields.
 /// decode::visualizer_error is intentionally not routed through LintConfig --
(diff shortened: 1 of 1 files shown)


src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -1604,6 +1650,223 @@ mod tests {
...
+    /// size cap to 4096 bytes but only for the SIMD-0385 `v1` format, which
+    /// shares no framing with legacy/v0 (fixed-width counts instead of
+    /// compact-u16, and signatures trailing rather than leading). `from_string`
+    /// always tries both real decoders (versioned, then legacy) first; the
+    /// `0x81` leading-byte check for `v1` only runs once both have failed, so
+    /// it can never shadow a payload that is actually decodable. These tests
+    /// pin that behavior: we reject anything neither decoder can read, we say
+    /// why, and a legacy/v0 payload that merely happens to start with `0x81`
+    /// (an outer compact-u16 signature count of 129, 257, ...) still decodes.
+    mod unsupported_versions {
```
nathan-margosian-anchor, line 1665: Are any of the functions in this mod unsupported_versions used anywhere?

nathan-margosian-anchor (approved):
Just a note that SIM-0296 does have a feature gate on Mainnet (even though the SIMD doc on Github doesn't have it)
```
solana feature status -um | grep 0296
6TkHkRmP7JZy1fdM6fg5uXn76wChQBWGokHBJzrLB3mj | inactive                | NA              | SIMD-0296: Raise CPI nesting limit from 4 to 8
```

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
pepe-anchor: Yep, only in the tests tho.

PR #425, opened 2026-07-30
feat(near): add visualsign-near chain crate skeleton

shahan-khatchadourian-anchorage:
First PR of a stack adding NEAR + NEAR Intents support as a first-class chain crate, alongside the other chains.

Stack: **A1 (this): crate skeleton** -> A2: baseline chain parser (adds the `defuse-*`/`near-sdk` pins + decode) -> A3: intents preset -> A4: signature verification -> B: `CHAIN_NEAR` identity + registry/CLI wiring -> C: docs site -> D: wallet-signed token metadata.

## What this PR does

- **`chain_parsers/visualsign-near` skeleton**: manifest with a single dependency, `thiserror = "2.0.12"`, matching every other chain crate's manifest -- plus an initial `NearParserError` enum (`BorshDecode`, `TrailingData`, `UnsupportedAction`). No decode/render code yet, and no `defuse-*`/`near-sdk`/`near-primitives` dependencies -- those land in A2/A3 alongside the code that uses them.
- Registers `visualsign-near` as a new workspace member in `src/Cargo.toml`. No new workspace-level dependency pins.
- **`parser_app` gains a `near` feature (not in `default`)** that gates only the optional dependency. This is required in this PR by design: the narrow-build matrix auto-derives from `chain_parsers/visualsign-*` directories, so the PR that introduces the directory must introduce the feature ("fails loudly at narrow-build-check on the PR that introduces it"). Registration in `create_registry()` and joining `default` come with the wiring PR (B).
- **Labeler entry** for `chain:near`.

## Verification

- `cargo check -p parser_app --features near` -- clean (non-default feature gate holds).
- `make -C src lint`, `make -C src test` -- clean (default features unaffected).

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -14,6 +14,7 @@ members = [
   "parser/grpc-server",
   "visualsign",
   "chain_parsers/visualsign-ethereum",
+  "chain_parsers/visualsign-near",
   "chain_parsers/visualsign-solana",
   "chain_parsers/visualsign-sui",
   "chain_parsers/visualsign-tron",
@@ -73,6 +74,24 @@ tracing = "0.1.41"
 
 # Pin visualsign dependencies
 visualsign = { path = "./visualsign" }
+
+# NEAR: near-sdk's `non-contract-usage` feature routes the defuse crates'
+# env::{sha256, keccak256, ed25519_verify, ecrecover} calls to pure-Rust
+# crypto so intent verification runs natively (no NEAR VM).
+near-sdk = { version = "5.24", features = ["non-contract-usage"] }
+
+# NEAR Intents (defuse) signing/verification crates, from the public
+# near/intents repo at the release tag. One pin point for the whole set.
+defuse-core        = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-crypto      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-deadline    = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-erc191      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-nep413      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-sep53       = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-tip191      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-token-id    = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2", default-features = false }
+defuse-ton-connect = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2", default-features = false, features = ["text"] }
+defuse-webauthn    = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
 
 [workspace.lints.clippy]
 unwrap_used = "deny"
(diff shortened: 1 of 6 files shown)


src/Cargo.toml
```diff
@@ -74,6 +75,24 @@ tracing = "0.1.41"
 # Pin visualsign dependencies
 visualsign = { path = "./visualsign" }
 
+# NEAR: near-sdk's `non-contract-usage` feature routes the defuse crates'
+# env::{sha256, keccak256, ed25519_verify, ecrecover} calls to pure-Rust
+# crypto so intent verification runs natively (no NEAR VM).
+near-sdk = { version = "5.24", features = ["non-contract-usage"] }
+
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Good catch. The gap was real, and it turned out to matter: near-sdk versions before 5.27 call an unresolvable native host symbol from their panic path, which fails to link outside the NEAR VM (this is why the original design called for an unsafe shim). 5.27 restructured that path (the near-sdk-env crate) to a plain Rust panic on native targets, which is what actually makes this workspace's crypto verification path link without a shim at all.

So rather than pin back to 5.24, I tightened the manifest's floor to 5.27 (aed3b53) with a comment explaining why -- no lockfile change, since 5.27.0 already satisfied the tightened requirement.

src/Cargo.toml
```diff
@@ -74,6 +75,19 @@ tracing = "0.1.41"
...
+defuse-core        = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-crypto      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-deadline    = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-erc191      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-nep413      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-sep53       = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-tip191      = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-token-id    = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2", default-features = false }
+defuse-ton-connect = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2", default-features = false, features = ["text"] }
+defuse-webauthn    = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
```
prasanna-anchorage: are all these deps required yet? can we add them as needed in followup PRs so that they're easier to review than batch add unnecessary dependencies that cannot be validated how they're used in the PR?

prasanna-anchorage (changes requested):

src/Cargo.toml
shahan-khatchadourian-anchorage: Adding deps as needed makes sense

shahan-khatchadourian-anchorage:

prasanna-anchorage (approved):
Thanks for narrowing this -- the batch dep add is gone; the crate now pulls only `thiserror = "2.0.12"`, matching every other chain crate manifest. The defuse/`near-sdk` pins can land in A2/A3 alongside the code that uses them, which is exactly what I was after.

Verified locally on this branch: `cargo clippy -p visualsign-near --all-targets -- -D warnings` clean, `cargo fmt --check` clean, and `cargo check -p parser_app --features near` builds, so the non-default feature gate holds.

Non-blocking: the PR description is now stale -- it still says this PR adds the workspace `defuse-*` pins and the `near-sdk >= 5.27` floor, and that the manifest carries "the crate's full dependency set (`near-primitives`/`near-crypto`, the defuse set)". None of that is in the diff anymore. Worth trimming so the merge commit message matches what actually landed; the `near-sdk >= 5.27` rationale is good context to carry forward to whichever PR introduces the pin.

PR #426, opened 2026-07-30
feat(near): add baseline NEAR transaction parser

shahan-khatchadourian-anchorage:
Second PR of the NEAR chain-support stack (stacked on #425). Implements baseline NEAR transaction parsing in this crate's conventions.

Stack: A1 (#425) -> **A2 (this)** -> A3 (intents preset) -> A4 (signature verification) -> B (identity + wiring) -> C (docs site) -> D (wallet-signed token metadata).

## What this PR does

Adds baseline NEAR chain decoding to `chain_parsers/visualsign-near`:

- `tx.rs`: borsh-decodes a NEAR `Transaction`/`SignedTransaction` from hex or base64 input.
- `actions.rs`: renders each `Action` variant (Transfer, FunctionCall with a fail-closed JSON args decoder for `ft_transfer`/`ft_transfer_call`/`ft_withdraw`, generic label fallback for the rest).
- `convert.rs`: assembles the final payload (Network/From/To + per-action fields) via `NearVisualSignConverter`.
- `fmt.rs`/`networks.rs`: yoctoNEAR/Tgas decimal formatting, Mainnet/Testnet display.

Two design decisions worth calling out:

- **`tx.rs`'s input decoding** uses this repo's shared `visualsign::encodings::SupportedEncodings::detect` + `decode_hex` convention (matching `visualsign-solana`'s own `from_string`) rather than hand-rolling `0x`-prefix stripping -- this repo's convention is to reuse the shared mechanism rather than hand-roll per chain. Added a regression test for the `0x`-prefixed case.
- **`convert.rs` omits `execute_intents` decoding for now** -- that code depends on `presets/intents/`, which lands in A3, compiled unconditionally there (no feature gate in the final design, since every other preset in this repo is unconditional too).

A1's placeholder `NearParserError` enum is removed as dead code in favor of `visualsign::vsptrait::TransactionParseError` directly; the now-unused `thiserror` dependency is dropped with it.

## Review fixes folded in

- `NearVisualSignConverter::with_network(NearNetwork)`: external callers (e.g. `parser_app`) can now construct a testnet converter without making the `network` field public.
- Two stale module doc comments corrected: `fmt.rs` claimed a base58 helper that doesn't exist in this file; `networks.rs` claimed "system-account recognition" that was never implemented.
- `payload_type` switched from the generic `"VisualSign"` to `"NearTx"`, matching every other chain crate's convention (`visualsign-solana`: `"SolanaTx"`, `visualsign-tron`: `"TronTx"`).
[...]

PR #428, opened 2026-07-30
feat(near): add NEAR Intents preset (decode + render) and merge converters

shahan-khatchadourian-anchorage:
Third PR of the NEAR chain-support stack (stacked on #426). Adds the NEAR Intents preset -- decode + render for both the signed-envelope and pre-signature flows -- and merges the crate's two input-format handlers into one.

Stack: A1 (#425) -> A2 (#426) -> **A3 (this)** -> A4 (#429) -> B (identity + wiring) -> C (docs site) -> D (wallet-signed token metadata).

## The preset

`chain_parsers/visualsign-near/src/presets/intents/`:

- `mod.rs` -- entry points (`try_decode_execute_intents` for a signed batch, `try_render_single_intent` for the pre-signature view), `NearIntentsError`, `NearTokenRegistry`/`TokenMeta`. The public surface is `visualsign` types plus primitives; `defuse-*` and `near-sdk` types stay behind this module.
- `args.rs` -- JSON deserialization of `execute_intents` args (`{"signed": [...]}`).
- `tokens.rs` -- seeded NEP-141 metadata and amount formatting.
- `render.rs` -- all 11 intent variants, plus the envelope and signature sections.
- `verify.rs` -- signature verification and payload extraction, reported independently of each other.

## The converter merge

`tx.rs`'s `NearTransaction` is an enum:

```rust
pub enum NearTransaction {
    OnChain(near_primitives::transaction::Transaction),
    Intent(String), // validated DefusePayload JSON, kept as text
}
```

`from_string` tries borsh first (via the existing hex/base64 detection); on failure it validates the input as a `DefusePayload` JSON envelope, and rejects anything that is neither. Borsh bytes are never valid JSON, so the discrimination is unambiguous. One chain identity, format-discriminated, rather than a converter per input format under its own custom chain name.

`convert.rs` dispatches on the variant: `OnChain` renders A2's baseline fields, plus the decoded intents when the action is a `FunctionCall` to `intents.near`/`execute_intents`; `Intent` renders via `try_render_single_intent`, titled "NEAR Intent", with no signature section because nothing is signed at that stage.

## What the signing screen states

- **Verification and extraction are independent.** A bad signature still renders the decoded intents, flagged at the signature line; a body that fails to extract surfaces the reason as a diagnostic rather than silently omitting the envelope.
- **Every field that changes what is authorized renders.** The three withdraws' `storage_deposit` (an unconditional, unrefundable wNEAR debit alongside the withdraw) and `msg` (switches the call into its `_transfer_call` form); `Transfer`'s flattened notification, whose `msg` calls `mt_on_transfer` on the receiver; and the NEP-616 `state_init` on both `Transfer`'s notification and `AuthCall`, which initializes the receiver's contract in the same receipt. `state_init`'s code and data have no cheap field-level render, so it is flagged `(not fully decoded)`, matching `actions.rs`'s handling of NEAR's own `DeterministicStateInit`. `min_gas` is the only field omitted, consistently across the four variants that carry it.
- **`mt_withdraw` rejects a `token_ids`/`amounts` length mismatch** rather than rendering the `zip` overlap and dropping the rest.
- **Signature outcomes are three, not two.** `Valid`, `Invalid`, and `MalformedEncoding` -- a signature that is cryptographically sound but carries Ethereum's `v = recovery_id + 27` against the 0-3 this wire format expects. Reporting that as `Invalid` would render "signature verification failed", which reads as tampering. A separate guard rejects out-of-range recovery ids before `verify()`, because near-crypto's native `ecrecover` panics on them rather than returning `None`, and a signing service must treat malformed input as invalid, never as a crash. The variant list is explicit, so a future secp256k1-backed standard is a build break rather than a latent panic.
- **The account-binding fields claim only what was checked.** When `signer_id` has an implicit-account shape, the recovered key's implied account id is compared against it and the result stated literally. Neither outcome settles authorization: the contract accepts a key that either derives the account id *or* sits in the account's on-chain key set (`Account::has_public_key`, `defuse/v0.4.2`), so a derivation match can still be rejected on-chain and a non-match is expected for any account that added keys via `AddPublicKey`. A named account has nothing to compare against, so no claim is made.
- **Deadline expiry is advisory.** `has_expired()` reads the system clock, which is not necessarily trustworthy inside the enclave; the contract enforces the real deadline.
- **Diagnostics carry in both build shapes.** With the `diagnostics` feature a soft finding is a structured `Diagnostic` field; by default it is the same information as a `Warning`-labelled text field, leaving the production payload shape unchanged.

## Token metadata

`SEEDS` is a compiled-in table of entries verified against the token contract's own `ft_metadata`; anything unverified is omitted and falls back to the raw asset id, since a wrong `decimals` silently misrenders amounts. Request-scoped overrides layer on top via `LayeredRegistry`. `decimals` above 38 does not resolve at all: `format_units` scales by `10^decimals`, which exceeds `u128` past that point, so an out-of-range value would either overflow or -- with overflow checks off -- divide by a wrapped-around scale and misrender the amount being signed.

## Dependencies

`defuse-core`/`defuse-crypto`/`defuse-deadline`, `near-sdk`, `chrono`, `bs58` and `hex` land here, where they are first used. near-sdk 5.27+ (what this workspace resolves) fixes the native-linking gap on the panic path, so `visualsign-near` keeps the workspace's `unsafe_code = "forbid"` with no carve-out and no shim.

## Verification

- `cargo build -p visualsign-near` -- clean.
- `cargo test -p visualsign-near` -- 94/94 pass, under both default and `--features diagnostics`.
- `cargo clippy -p parser_app --no-default-features --features near` -- clean (narrow-build variant).
- `make -C src lint` -- clean; `make -C src test` -- clean (workspace suite, including the 9 gRPC integration tests).

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -75,6 +75,12 @@ tracing = "0.1.41"
 # Pin visualsign dependencies
 visualsign = { path = "./visualsign" }
 
+near-sdk = { version = "5.27", features = ["non-contract-usage"] }
+
+defuse-core     = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-crypto   = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+defuse-deadline = { git = "https://github.com/near/intents.git", tag = "defuse/v0.4.2" }
+
 [workspace.lints.clippy]
 unwrap_used = "deny"
 expect_used = "deny"
--- a/src/chain_parsers/visualsign-near/src/convert.rs
+++ b/src/chain_parsers/visualsign-near/src/convert.rs
@@ -1,8 +1,12 @@
-//! `NearVisualSignConverter`: NEAR transaction -> VisualSign payload.
+//! `NearVisualSignConverter`: NEAR input -> VisualSign payload.
+
+use std::sync::Arc;
 
 use near_primitives::action::Action;
+use near_primitives::transaction::Transaction;
 use visualsign::errors::VisualSignError;
 use visualsign::field_builders::{create_address_field, create_text_field};
+use visualsign::registry::LayeredRegistry;
 use visualsign::vsptrait::{
     ConversionResult, VisualSignConverter, VisualSignConverterFromString, VisualSignOptions,
 };
@@ -10,11 +14,12 @@ use visualsign::{SignablePayload, SignablePayloadField};
 
 use crate::actions::render_action;
 use crate::networks::{NearNetwork, extract_network_from_metadata};
+use crate::presets::intents::NearTokenRegistry;
 use crate::tx::NearTransaction;
 
-/// Payload version emitted for NEAR transactions.
+/// Payload version emitted for NEAR payloads.
 const PAYLOAD_VERSION: i64 = 0;
-/// Payload type tag emitted for NEAR transactions, matching the other chain
+/// Payload type tag emitted for NEAR payloads, matching the other chain
 /// crates' convention (`"SolanaTx"`, `"TronTx"`).
 const PAYLOAD_TYPE: &str = "NearTx";
 
@@ -44,8 +49,19 @@ impl VisualSignConverter<NearTransaction> for NearVisualSignConverter {
         transaction: NearTransaction,
         options: VisualSignOptions,
     ) -> Result<ConversionResult, VisualSignError> {
-        let tx = transaction.inner();
+        match transaction {
+            NearTransaction::OnChain(tx) => self.render_on_chain(&tx, &options),
+            NearTransaction::Intent(json) => render_intent_envelope(&json, &options),
+        }
+    }
+}
 
+impl NearVisualSignConverter {
+    fn render_on_chain(
+        &self,
+        tx: &Transaction,
+        options: &VisualSignOptions,
+    ) -> Result<ConversionResult, VisualSignError> {
         if tx.actions().is_empty() {
             return Err(VisualSignError::ValidationError(
                 "NEAR transaction has no actions".to_string(),
@@ -78,6 +94,7 @@ impl VisualSignConverter<NearTransaction> for NearVisualSignConverter {
         let total_actions = tx.actions().len();
         for action in tx.actions() {
             fields.extend(render_action(action, total_actions)?);
+            fields.extend(decode_intents(tx.receiver_id().as_str(), action, options)?);
         }
 
         Ok(ConversionResult::new(SignablePayload::new(
@@ -103,6 +120,48 @@ impl VisualSignConverterFromString<NearTransaction> for NearVisualSignConverter 
         .map_err(VisualSignError::ParseError)?;
         self.to_validated_visual_sign_payload(transaction, options)
     }
+}
+
+/// Decode an `execute_intents` call to `intents.near` and render the signed
+/// intent batch. Any other action or receiver yields no extra fields. A
+/// decode failure surfaces as a conversion error rather than silently
+/// dropping the intents.
+fn decode_intents(
+    receiver_id: &str,
+    action: &Action,
+    options: &VisualSignOptions,
+) -> Result<Vec<SignablePayloadField>, VisualSignError> {
+    if receiver_id != "intents.near" {
+        return Ok(vec![]);
+    }
+    let Action::FunctionCall(fc) = action else {
+        return Ok(vec![]);
+    };
+    if fc.method_name != "execute_intents" {
+        return Ok(vec![]);
+    }
+    let registry = LayeredRegistry::new(Arc::new(NearTokenRegistry::default()));
+    crate::presets::intents::try_decode_execute_intents(&fc.args, &registry, options)
+        .map_err(|e| VisualSignError::ConversionError(e.to_string()))
+}
+
+/// Render the pre-signature intents envelope a user is about to sign: no
+/// signature exists yet, so this is a confirmation view only.
+fn render_intent_envelope(
+    json: &str,
+    options: &VisualSignOptions,
+) -> Result<ConversionResult, VisualSignError> {
+    let registry = LayeredRegistry::new(Arc::new(NearTokenRegistry::default()));
+    let fields =
+        crate::presets::intents::try_render_single_intent(json.as_bytes(), &registry, options)
+            .map_err(|e| VisualSignError::ConversionError(e.to_string()))?;
+    Ok(ConversionResult::new(SignablePayload::new(
+        PAYLOAD_VERSION,
+        "NEAR Intent".to_string(),
+        None,
+        fields,
+        PAYLOAD_TYPE.to_string(),
+    )))
 }
 
 /// Title for the payload: a single action names itself, otherwise a generic
@@ -139,10 +198,10 @@ mod tests {
     use super::*;
     use generated::parser::{ChainMetadata, NearMetadata, chain_metadata};
     use near_crypto::{KeyType, PublicKey, Signature};
-    use near_primitives::action::{CreateAccountAction, TransferAction};
+    use near_primitives::action::{CreateAccountAction, FunctionCallAction, TransferAction};
     use near_primitives::hash::CryptoHash;
     use near_primitives::transaction::{SignedTransaction, TransactionV0};
-    use near_primitives::types::Balance;
+    use near_primitives::types::{Balance, Gas};
     use visualsign::vsptrait::DeveloperConfig;
 
     fn transfer() -> Action {
@@ -156,16 +215,14 @@ mod tests {
     }
 
     fn near_tx_as(signer_id: &str, receiver_id: &str, actions: Vec<Action>) -> NearTransaction {
-        NearTransaction::new(near_primitives::transaction::Transaction::V0(
-            TransactionV0 {
-                signer_id: signer_id.parse().expect("valid account id"),
-                public_key: PublicKey::empty(KeyType::ED25519),
-                nonce: 0,
-                receiver_id: receiver_id.parse().expect("valid account id"),
-                block_hash: CryptoHash::default(),
-                actions,
-            },
-        ))
+        NearTransaction::OnChain(Transaction::V0(TransactionV0 {
+            signer_id: signer_id.parse().expect("valid account id"),
+            public_key: PublicKey::empty(KeyType::ED25519),
+            nonce: 0,
+            receiver_id: receiver_id.parse().expect("valid account id"),
+            block_hash: CryptoHash::default(),
+            actions,
+        }))
     }
 
     #[test]
@@ -290,7 +347,10 @@ mod tests {
     }
 
     fn signed_transfer_hex() -> String {
-        let unsigned = near_tx(vec![transfer()]).inner().clone();
+        let unsigned = match near_tx(vec![transfer()]) {
+            NearTransaction::OnChain(tx) => tx,
+            NearTransaction::Intent(_) => panic!("expected OnChain"),
+        };
         let signed = SignedTransaction::new(Signature::empty(KeyType::ED25519), unsigned);
         hex::encode(borsh::to_vec(&signed).expect("borsh encode"))
     }
@@ -316,5 +376,61 @@ mod tests {
         };
         let result = converter.to_visual_sign_payload_from_string(&signed_transfer_hex(), options);
         assert!(result.is_ok());
+    }
+
+    #[test]
+    fn execute_intents_call_renders_intents() {
+        let inner = r#"{"signer_id":"alice.near","verifying_contract":"intents.near","deadline":"2999-01-01T00:00:00Z","nonce":"XVoKfmScb3G+XqH9ke/fSlJ/3xO59sNhCxhpG821BH8=","intents":[{"intent":"ft_withdraw","token":"wrap.near","receiver_id":"bob.near","amount":"1000000000000000000000000"}]}"#;
+        let args = serde_json::json!({"signed":[{
+            "standard": "raw_ed25519",
+            "payload": inner,
+            "public_key": "ed25519:8rVvtHWFr8hasdQGGD5WiQBTyr4iH2ruEPPVfj491RPN",
+            "signature": "ed25519:3vtbNQJHZfuV1s5DykzyjkbNLc583hnkrhTz57eDhd966iqzkor6Twgr4Loh2C195SCSEsiGfrd6KcxpjNq9ZbVj"
+        }]});
+
+        let txv0 = TransactionV0 {
+            signer_id: "alice.near".parse().unwrap(),
+            public_key: "ed25519:8rVvtHWFr8hasdQGGD5WiQBTyr4iH2ruEPPVfj491RPN"
+                .parse()
+                .unwrap(),
+            nonce: 1,
+            receiver_id: "intents.near".parse().unwrap(),
+            block_hash: CryptoHash::default(),
+            actions: vec![Action::FunctionCall(Box::new(FunctionCallAction {
+                method_name: "execute_intents".to_string(),
+                args: serde_json::to_vec(&args).unwrap(),
+                gas: Gas::from_gas(30_000_000_000_000),
+                deposit: Balance::from_yoctonear(0),
+            }))],
+        };
+        let near_tx = NearTransaction::OnChain(Transaction::V0(txv0));
+        let payload = NearVisualSignConverter::new()
+            .to_visual_sign_payload(near_tx, VisualSignOptions::default())
+            .expect("convert");
+        let json = payload.payload.to_json().expect("json");
+
+        // Generic FunctionCall view + decoded intents both present.
+        assert!(json.contains("execute_intents"), "method missing: {json}");
+        assert!(json.contains("Signer"), "intents envelope missing: {json}");
+        assert!(
+            json.contains("wNEAR"),
+            "resolved ft_withdraw amount missing: {json}"
+        );
+    }
+
+    #[test]
+    fn intent_envelope_renders_without_signature_section() {
+        let swap = r#"{"signer_id":"alice.near","verifying_contract":"intents.near","deadline":"2100-01-01T00:00:00Z","nonce":"XVoKfmScb3G+XqH9ke/fSlJ/3xO59sNhCxhpG821BH8=","intents":[{"intent":"ft_withdraw","token":"wrap.near","receiver_id":"bob.near","amount":"1000000000000000000000000"}]}"#;
+        let payload = NearVisualSignConverter::new()
+            .to_visual_sign_payload(
+                NearTransaction::Intent(swap.to_string()),
+                VisualSignOptions::default(),
+            )
+            .expect("convert");
+        let json = payload.payload.to_json().expect("json");
+        for expected in ["Signer", "alice.near", "wNEAR"] {
+            assert!(json.contains(expected), "missing {expected}: {json}");
+        }
+        assert!(!json.contains("Standard"), "unexpected signature section");
     }
 }
--- a/src/chain_parsers/visualsign-near/src/presets/intents/mod.rs
+++ b/src/chain_parsers/visualsign-near/src/presets/intents/mod.rs
@@ -19,10 +19,6 @@ pub(crate) const NEAR_DECIMALS: u8 = 24;
 pub enum NearIntentsError {
     #[error("input was not valid JSON: {0}")]
     InputNotJson(String),
-    #[error("Unknown intent action variant: {0}")]
-    UnknownAction(String),
-    #[error("Malformed asset identifier: {0}")]
-    MalformedAssetId(String),
     #[error("Failed to render intents: {0}")]
     Render(String),
 }
--- a/src/chain_parsers/visualsign-near/src/presets/intents/mod.rs
+++ b/src/chain_parsers/visualsign-near/src/presets/intents/mod.rs
@@ -1,0 +1,241 @@
+//! NEAR Intents protocol decoder.
+//!
+//! Public API uses only `visualsign` types + plain primitives. The defuse-*
+//! and near-sdk transitive dependencies are confined to this module's
+//! internals.
+
+mod args;
+mod render;
+mod tokens;
+mod verify;
+
+use thiserror::Error;
+
+/// NEAR's native token symbol and decimal scale (yoctoNEAR = 10^-24 NEAR).
+pub(crate) const NEAR_SYMBOL: &str = "NEAR";
+pub(crate) const NEAR_DECIMALS: u8 = 24;
+
+#[derive(Debug, Error)]
+pub enum NearIntentsError {
+    #[error("execute_intents args were not valid JSON: {0}")]
+    ArgsNotJson(String),
+    #[error("Unknown intent action variant: {0}")]
+    UnknownAction(String),
+    #[error("Malformed asset identifier: {0}")]
+    MalformedAssetId(String),
+    #[error("Failed to render intents: {0}")]
+    Render(String),
+}
+
+/// Plain NEP-141 token metadata. The lean baseline chain parser constructs
+/// the registry; this module consumes it via `LayeredRegistry<NearTokenRegistry>`.
+#[derive(Debug, Default)]
+pub struct NearTokenRegistry {
+    pub by_asset_id: std::collections::BTreeMap<String, TokenMeta>,
+}
+
+#[derive(Debug, Clone)]
+pub struct TokenMeta {
+    pub symbol: String,
+    pub decimals: u8,
+}
+
+/// Decode `execute_intents` args and render them as `SignablePayloadField`s.
+///
+/// This is the sole public entry point for the signed-envelope batch.
+pub fn try_decode_execute_intents(
+    args: &[u8],
+    token_registry: &visualsign::registry::LayeredRegistry<NearTokenRegistry>,
+    _options: &visualsign::vsptrait::VisualSignOptions,
+) -> Result<Vec<visualsign::SignablePayloadField>, NearIntentsError> {
+    let payloads = args::decode_args(args)?;
+    let total = payloads.len();
+    let mut fields = Vec::new();
+    for (i, mp) in payloads.iter().enumerate() {
+        fields.extend(
+            render::section(i + 1, total, mp, token_registry)
+                .map_err(|e| NearIntentsError::Render(e.to_string()))?,
+        );
+    }
+    Ok(fields)
+}
+
+/// Render the single intent a user is about to sign, from the JSON of its
+/// `DefusePayload` message (the inner message that gets signed, independent of
+/// which signature standard later wraps it).
+///
+/// This is the user-signing view: it reuses the envelope + per-intent rendering
+/// but skips signature verification (no signature exists at signing time).
+pub fn try_render_single_intent(
+    payload_json: &[u8],
+    token_registry: &visualsign::registry::LayeredRegistry<NearTokenRegistry>,
+    _options: &visualsign::vsptrait::VisualSignOptions,
+) -> Result<Vec<visualsign::SignablePayloadField>, NearIntentsError> {
+    let payload: defuse_core::payload::DefusePayload<defuse_core::intents::DefuseIntents> =
+        serde_json::from_slice(payload_json)
+            .map_err(|e| NearIntentsError::ArgsNotJson(e.to_string()))?;
+    render::render_single(&payload, token_registry)
+        .map_err(|e| NearIntentsError::Render(e.to_string()))
+}
+
+/// Test-only matcher shared across this module's test submodules.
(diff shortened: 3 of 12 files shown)


src/chain_parsers/visualsign-near/src/presets/intents/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+use thiserror::Error;
+
+/// NEAR's native token symbol and decimal scale (yoctoNEAR = 10^-24 NEAR).
+pub(crate) const NEAR_SYMBOL: &str = "NEAR";
+pub(crate) const NEAR_DECIMALS: u8 = 24;
+
+#[derive(Debug, Error)]
+pub enum NearIntentsError {
+    #[error("execute_intents args were not valid JSON: {0}")]
+    ArgsNotJson(String),
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed -- renamed NearIntentsError::ArgsNotJson to InputNotJson with a generic message. The variant is shared between execute_intents args decoding and try_render_single_intent's payload JSON parse, and the old message named only the former.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -0,0 +1,519 @@
...
+    registry: &Reg,
+) -> Result<Fields, VisualSignError> {
+    let (check, extracted) = super::verify::verify_and_extract(mp);
+    let mut fields = vec![
+        create_text_field("Signed Intent", &format!("{index} of {total}"))?.signable_payload_field,
+    ];
+    fields.extend(render_signature(standard_name(mp), &check)?);
+    if let Some(payload) = &extracted {
+        fields.extend(render_single(payload, registry)?);
+    }
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed -- verify_and_extract now returns Result<DefusePayload<..>, String> instead of collapsing extraction failure to None, and section() renders that error as an "extraction" diagnostic alongside the signature status, so a failed extraction is visible instead of silently omitting the envelope/intents.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -0,0 +1,549 @@
...
+
+/// Render the standard + signature-status fields for one signed payload.
+pub(crate) fn render_signature(
+    standard: &str,
+    check: &SignatureCheck,
+) -> Result<Vec<SignablePayloadField>, VisualSignError> {
+    let mut fields = vec![create_text_field("Standard", standard)?.signable_payload_field];
+    match check {
+        SignatureCheck::Valid { recovered_key } => fields.push(
+            create_text_field("Signature", &format!("valid (recovered {recovered_key})"))?
```
prasanna-anchorage: This renders `Signature: valid (recovered <key>)`, and three lines later `render_envelope` renders `Signer: <signer_id>`. Adjacent on a signing screen, those read as one claim: *this account signed this*. But they're independent -- `verify()` recovers whatever key signed the bytes, and nothing checks that key is bound to `signer_id`. `verify.rs` says so explicitly ("NOT a proof of account binding"); the rendered output doesn't.

I reproduced it. Payload claiming `signer_id: alice.near`, signed with a key that has no relationship to that account, withdrawing 5 NEAR to `attacker.near`:

```
Signed Intent          1 of 1
Standard               raw_ed25519
Signature              valid (recovered ed25519:4uA4xtFiedwCyGLuDLn49Pz4AdNnKJFpEuH4LgESHjwK)
Signer                 alice.near
Verifying Contract     intents.near
Deadline               2999-01-01T00:00:00+00:00
To                     attacker.near
Amount                 5 NEAR
```

Everything there is true and the payload is exactly what the parser was handed -- the problem is the word `valid` unqualified next to an account the signature says nothing about. The on-chain contract does enforce key-to-account binding, so this isn't exploitable *against the protocol*; the risk is a reviewer or a wallet UI treating a green "valid" as attestation that `alice.near` authorized it.

The parser can't resolve account keys (no chain access), so the fix is wording, not verification -- make the field say only what was checked:

```rust
create_text_field(
    "Signature",
    &format!("valid for key {recovered_key} (key-to-account binding not verified)"),
)?
```

Or split into a `Signed By Key` field plus a diagnostic when the recovered key can't be tied to `signer_id`. Either way I'd rather the payload under-claim than let `valid` carry an implication the code deliberately doesn't check.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -0,0 +1,549 @@
...
+    push_withdraw_call_details(&mut fields, &w.msg, w.storage_deposit)?;
+    Ok(fields)
+}
+
+fn render_mt_withdraw(w: &MtWithdraw) -> Result<Fields, VisualSignError> {
+    let mut fields = vec![
+        create_text_field("Token", w.token.as_str())?.signable_payload_field,
+        create_text_field("To", w.receiver_id.as_str())?.signable_payload_field,
+    ];
+    for (id, amount) in w.token_ids.iter().zip(w.amounts.iter()) {
```
prasanna-anchorage, line 297: `zip` stops at the shorter iterator, and `MtWithdraw` carries `token_ids: Vec<TokenId>` and `amounts: Vec<U128>` as two independent vectors with no length agreement enforced at deserialize time. A payload with more `token_ids` than `amounts` renders only the overlap -- the extra withdrawals are silently dropped from the signing view.

Reproduced with `{"token_ids":["gold","silver","bronze"],"amounts":["5"]}`:

```
Token                  mt.near
To                     attacker.near
MT Token               gold x5
=> MT Token lines rendered: 1 (token_ids declared: 3)
```

The user sees one token move; the intent names three. This is the same class of gap the `storage_deposit` regression test was written to prevent -- something material to what's being authorized not reaching the screen -- and here it's attacker-controlled input rather than an oversight.

Suggest treating a length mismatch as malformed rather than rendering a partial view:

```rust
if w.token_ids.len() != w.amounts.len() {
    return Err(/* malformed mt_withdraw: N token_ids vs M amounts */);
}
```

Erroring out is defensible since the contract would reject it anyway; a diagnostic plus rendering every `token_id` (amount shown as unknown where missing) would also work. What's not safe is the current silent truncation. Worth a regression test alongside the `storage_deposit` ones.

Unrelated and much smaller, same function: `min_gas` is on `MtWithdraw` and isn't rendered. Probably fine to omit, just flagging it's dropped.

src/chain_parsers/visualsign-near/src/presets/intents/verify.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Rejects secp256k1 signatures whose recovery byte is out of range before
+/// they reach `verify()`: near-crypto's native `ecrecover` backend panics on
+/// an invalid recovery id rather than returning `None`, and a signing service
+/// must treat a malformed payload as invalid input, never as a crash.
+fn has_invalid_secp256k1_recovery_id(payload: &MultiPayload) -> bool {
+    let signature = match payload {
+        MultiPayload::Erc191(signed) => &signed.signature,
+        MultiPayload::Tip191(signed) => &signed.signature,
+        _ => return false,
```
prasanna-anchorage: The guard itself is correct -- I checked `defuse/v0.4.2` and `Erc191` and `Tip191` are the only variants that route to `PublicKey::Secp256k1` / `ecrecover` (`multi.rs:82-83`), so the two arms above cover exactly the panicking paths today.

My concern is the `_`. This function exists because a panic in a signing service is unacceptable, but the catch-all silently pins that guarantee to the current variant list of a git-tagged dependency. If a later bump adds a secp256k1-backed standard, this falls through to `false`, reaches `verify()`, and the panic is back -- with nothing failing at compile time to warn you.

Since the whole point is panic-avoidance, I'd make the compiler enforce it by listing the variants explicitly:

```rust
MultiPayload::Nep413(_)
| MultiPayload::RawEd25519(_)
| MultiPayload::WebAuthn(_)
| MultiPayload::TonConnect(_)
| MultiPayload::Sep53(_) => return false,
```

Then a new variant is a build break at exactly the place someone needs to think about it, instead of a latent crash. A short comment on why each listed variant is panic-safe (ed25519/p256, no recovery id) would make the next bump mechanical.

prasanna-anchorage (commented):
Went through the preset and built it locally against the real `defuse/v0.4.2` crates -- 50 tests pass, clippy `--all-targets -D warnings` clean. The module boundary is well drawn: `defuse-*`/`near-sdk` types genuinely stay behind `presets::intents`, and splitting `verify()` from `extract_defuse_payload()` so a bad signature still renders the decoded intents is the right call. `storage_deposit` and `msg` rendering on all three withdraw variants, with regression tests naming *why*, is exactly the level of care this surface needs.

Two findings below I reproduced by running them, not by reading -- both are cases where the signing screen shows the user less than what they'd be authorizing. Third is a cheap robustness fix.

Smaller note, not worth its own thread: `deadline.has_expired()` reads the system clock (hence the new `chrono` `clock` feature). Inside the enclave the wall clock isn't necessarily trustworthy or even set, so a stale-deadline payload could render with no warning. Not a blocker -- the warning is advisory and the contract enforces the real deadline -- but worth knowing the field is best-effort.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Reworded per your suggestion — the field now reads `valid for key <key> (key-to-account binding not verified)` instead of bare `valid (recovered <key>)`. States exactly what was checked and nothing more.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Fixed — `render_mt_withdraw` now rejects a `token_ids`/`amounts` length mismatch as `VisualSignError::ValidationError` instead of silently rendering the `zip()` overlap. Added a regression test using your exact repro (3 `token_ids`, 1 `amount`). Left `min_gas` unrendered as you suggested.

src/chain_parsers/visualsign-near/src/presets/intents/verify.rs
shahan-khatchadourian-anchorage: Done — replaced the `_ => false` catch-all with the explicit list of the 5 non-secp256k1 `MultiPayload` variants (`Nep413`, `RawEd25519`, `WebAuthn`, `TonConnect`, `Sep53`), with a short comment on why each is panic-safe (ed25519/P256, no recovery id). A future `defuse-core` bump adding a secp256k1-backed standard now fails to compile here instead of silently reaching the panic path.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Implemented here in #428 (not #429), since render_signature/render_envelope live in this PR and needed the restructuring. SignatureCheck::Valid now also carries implied_account_id (via defuse_crypto::PublicKey::to_implicit_account_id()), and render_signature compares it against signer_id when the id has a recognized implicit-account shape (NEAR's 64-hex, or defuse's `0x`+40-hex EVM convention) — a real match/mismatch check, not a hedge. Named accounts (e.g. `alice.near`) still get the original hedge wording, since that binding genuinely isn't checkable offline. A mismatch renders as a diagnostic (a hard finding), not folded into the same wording as the confirmed case. Test coverage covers the full matrix: confirmed/mismatch for both shapes, the named-account hedge, and the shape-detection helper's edge cases directly.

shahan-khatchadourian-anchorage: Reworked the wording once more after checking the contract's rule. `Account::has_public_key` (`defuse/src/contract/accounts/account/mod.rs:88-91`) accepts a key that either derives the account id **or** sits in the account's on-chain `public_keys` set — so a non-match is expected for any account that authorized a key via `AddPublicKey`, and an implicit account that removed its implicit key would take the warning on every payload it signs. A match isn't proof either: `is_implicit_public_key_removed()` can make the contract reject a key that does derive the id.

Kept the diagnostic, dropped the verdict. Both arms now state only the observation — `valid for key <k> (derives signer_id <id>)`, or that it doesn't derive `signer_id` and the account may still have authorized it on-chain, which this parser can't check.

pepe-anchor (commented):
**Peer Review Summary — NEAR stack #428** (COMMENT)\n\n8 findings: 2 MEDIUM, 3 LOW, 3 INFO. Detailed inline comments below.\n\n**Key concern:** No size limit on JSON intent input (DoS). Potential panic on truncated signature.

---

src/chain_parsers/visualsign-near/src/tx.rs
```diff
@@ -29,22 +28,31 @@ impl NearTransaction {
...
-        )))
+        // Validate eagerly so malformed input is rejected at parse time
+        // rather than at render time.
+        serde_json::from_str::<
+            defuse_core::payload::DefusePayload<defuse_core::intents::DefuseIntents>,
+        >(trimmed)
+        .map_err(|e| {
+            TransactionParseError::DecodeError(format!(
+                "input is neither a NEAR borsh transaction nor a DefusePayload JSON envelope: {e}"
+            ))
```
pepe-anchor, line 77: **[MEDIUM] No size limit on JSON intent input (cat 2: Input Validation, cat 10: DoS)**

NearTransaction::from_string_with_options accepts a JSON DefusePayload via serde_json::from_str with no size bound. An attacker can supply an arbitrarily large JSON payload (millions of intents, deeply nested fields, or huge string values) causing unbounded memory allocation. The borsh path has an implicit limit from NEAR's max transaction size, but the JSON intent path has no equivalent guard. The project's established pattern (BoundedReader, 10MB file cap) is not applied here.

---

src/chain_parsers/visualsign-near/src/presets/intents/verify.rs
```diff
@@ -0,0 +1,98 @@
...
+    (check, extracted)
+}
+
+/// Rejects secp256k1 signatures whose recovery byte is out of range before
+/// they reach `verify()`: near-crypto's native `ecrecover` backend panics on
+/// an invalid recovery id rather than returning `None`, and a signing service
+/// must treat a malformed payload as invalid input, never as a crash.
+///
+/// Returns the human-readable reason when rejected, `None` when the
+/// signature's recovery id is in range (or the standard has none). The most
```
pepe-anchor, line 69: **[MEDIUM] Potential panic on truncated ERC191/TIP191 signature byte access (cat 9: Error Handling)**

invalid_secp256k1_recovery_id_reason() does signature[64] without bounds-checking. If a malformed MultiPayload deserializes with a truncated signature (< 65 bytes for Erc191/Tip191), this panics. The defuse library *should* enforce signature length at deserialization, but this code path relies entirely on an external dependency for crash-safety. A signing service that panics on malformed input is a denial-of-service vector. A len() >= 65 guard before the index would make this locally safe regardless of upstream behavior.

---

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -0,0 +1,877 @@
...
+fn render_ft_withdraw(w: &FtWithdraw, registry: &Reg) -> Result<Fields, VisualSignError> {
+    let mut fields = vec![
+        create_text_field("Token", w.token.as_str())?.signable_payload_field,
+        create_text_field("To", w.receiver_id.as_str())?.signable_payload_field,
+        token_amount_field(
+            "Amount",
+            &format!("nep141:{}", w.token),
+            w.amount.0,
+            registry,
+        )?,
```
pepe-anchor, line 263: **[LOW] looks_like_implicit_account uses case-sensitive '0x' prefix (cat 2: Input Validation)**

looks_like_implicit_account() uses strip_prefix("0x") which only matches lowercase 0x, not 0X. However, is_ascii_hexdigit() accepts both cases, and Ethereum addresses with uppercase 0X prefix are technically valid hex. A signer_id like '0X17c5...d025' would fail the 0x-prefix check but pass as 42 hex chars, so it would NOT match either branch of the boolean expression. This means an EVM address using the 0X prefix would not be recognized as implicit, silently skipping the account-binding check where it could have been performed. The rest of the codebase consistently uses visualsign::encodings::split_hex_prefix for case-insensitive prefix handling.

---

src/chain_parsers/visualsign-near/src/tx.rs
```diff
@@ -54,7 +62,10 @@ impl visualsign::vsptrait::Transaction for NearTransaction {
     }
 
     fn transaction_type(&self) -> String {
-        "NEAR".to_string()
+        match self {
```
pepe-anchor, line 89: **[LOW] Double memory allocation for intent payload string (cat 10: Availability)**

from_string_with_options first parses the JSON with serde_json::from_str (allocating the full deserialized structure), then discards the result and stores trimmed.to_string() as Intent(String), duplicating the input in memory. For large payloads, this doubles peak memory. The parsed DefusePayload could be stored instead of re-parsing at render time. However, the current design intentionally keeps defuse types out of the public crate API, so storing the string is a deliberate trade-off.

---

src/chain_parsers/visualsign-near/src/convert.rs
```diff
@@ -100,6 +117,48 @@ impl VisualSignConverterFromString<NearTransaction> for NearVisualSignConverter
     }
 }
 
+/// Decode an `execute_intents` call to `intents.near` and render the signed
+/// intent batch. Any other action or receiver yields no extra fields. A
```
pepe-anchor, line 126: **[LOW] Token registry always empty-default at all intent call sites**

Both decode_intents() and render_intent_envelope() construct LayeredRegistry::new(Arc::new(NearTokenRegistry::default())) with an empty default, and the request-scoped override layer is never populated. The tokens.rs module documents 'request-scoped override layer first (via the registry)' but the current call sites never supply overrides, so only the compiled-in seed table (single entry: wrap.near) is reachable. This limits token resolution to wNEAR only, which may be intentional at this stage, but the parameter exists to accept overrides that nothing yet provides.

---

src/Cargo.toml
```diff
@@ -75,6 +75,12 @@ tracing = "0.1.41"
 # Pin visualsign dependencies
 visualsign = { path = "./visualsign" }
```
pepe-anchor, line 77: **[INFO] defuse-* git dependencies pull in 17+ transitive crates**

The three workspace dependencies (defuse-core, defuse-crypto, defuse-deadline at tag defuse/v0.4.2, commit 6dad94cf) transitively pull in: defuse-auth-call, defuse-bitmap, defuse-borsh-utils, defuse-erc191, defuse-fees, defuse-io-utils, defuse-map-utils, defuse-near-utils, defuse-nep245, defuse-nep413, defuse-nep461, defuse-num-utils, defuse-sep53, defuse-serde-utils, defuse-tip191, defuse-token-id, defuse-ton-connect, defuse-webauthn. Additional new transitive crates: bnum 0.13, array-util 1.0, tlb-ton, schemars 0.8, near-contract-standards. All pinned to an exact git tag and commit -- good. But the supply chain surface expands significantly with crates from outside the existing dependency footprint. The verify.rs module depends on defuse-crypto::SignedPayload::verify() for its core cryptographic operation.

---

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -0,0 +1,877 @@
...
+        fields.push(create_text_field("Referral", referral.as_str())?.signable_payload_field);
+    }
+    Ok(fields)
+}
+
+fn render_transfer(t: &Transfer, registry: &Reg) -> Result<Fields, VisualSignError> {
+    let mut fields = vec![create_text_field("To", t.receiver_id.as_str())?.signable_payload_field];
+    for (token_id, amount) in t.tokens.iter() {
+        fields.push(token_amount_field(
+            "Amount",
```
pepe-anchor, line 209: **[INFO] token_diff with zero delta renders 'Receive 0' label**

In render_token_diff, a delta == 0 entry would render as 'Receive 0 <token>' since the condition uses is_negative() (which is false for 0) to pick 'Send' vs 'Receive'. A zero-delta entry in a token_diff is semantically odd (no net change) but renders as if the user received nothing. Not a security issue, but the rendering is confusing and could mask a malformed diff entry. The i128::unsigned_abs() call is safe for all values including i128::MIN (returns 1 << 127 as u128).

---

src/chain_parsers/visualsign-near/src/presets/intents/verify.rs
```diff
@@ -0,0 +1,98 @@
...
+                implied_account_id: key.to_implicit_account_id().to_string(),
+            },
+            None => SignatureCheck::Invalid,
+        },
+    };
+    let extracted = payload
+        .clone()
+        .extract_defuse_payload()
+        .map_err(|e| e.to_string());
+    (check, extracted)
```
pepe-anchor, line 60: **[INFO] verify_and_extract PROMISES vs DELIVERS: independent verification and extraction**

CONTRACT: verify() and extract_defuse_payload() are independent; both always attempted. DELIVERS: The verify result is computed first (guard + verify), then payload.clone().extract_defuse_payload() is called unconditionally. The clone ensures extraction cannot be affected by verify's potential consumption. Both results returned as a tuple. MalformedEncoding is checked before verify() to prevent a panic in near-crypto's ecrecover backend. Extraction failure carries the error message for rendering. Contract fully satisfied.

---

---

src/chain_parsers/visualsign-near/src/tx.rs
shahan-khatchadourian-anchorage: Not taking this one in the NEAR crate -- same reasoning as the equivalent thread on #426 (3723372998), which is worth reading together with this since both point at `decode_input`.

The CLI path is bounded: `parser/cli-core/src/tx_input.rs` sets `MAX_TRANSACTION_INPUT_SIZE = 10 * 1024 * 1024` and `read_bounded` enforces it. That helper is chain-neutral, so the JSON envelope path inherits it too -- it is the same reader, only the branch after decode differs.

The gRPC path has no per-chain cap for any chain, as you note. Adding `MAX_INPUT_BYTES` inside NEAR's `decode_input` would make NEAR the only chain with a second, lower, differently-worded limit while Ethereum, Solana, Sui and Tron stay unbounded on the same transport. The effective fix is a transport-level bound (tonic `max_decoding_message_size`), which covers every chain in one place.

Happy to file that separately if you agree it's worth doing -- it just shouldn't be a NEAR-shaped patch.

src/chain_parsers/visualsign-near/src/presets/intents/verify.rs
shahan-khatchadourian-anchorage: This one doesn't reproduce -- the index is checked at compile time.

`signature` here is `<Secp256k1 as Curve>::Signature`, which `defuse-crypto` defines as `[u8; 65]` (`crypto/src/curve/secp256k1.rs:18`, pinned at `defuse/v0.4.2`). It's a fixed-size array, not a slice, so `signature[64]` is statically in bounds and rustc const-checks it. A `len() >= 65` guard would be dead code.

The "truncated signature deserializes" precondition can't arise either: deserializing into `[u8; 65]` requires exactly 65 bytes, so a short input fails at decode rather than reaching this function.

Worth flagging that the review summary leads with this and the input-size finding as the key concerns, which overstates the risk -- same pattern as the trailing-data finding on #426 (3723372046), where borsh already enforced what the finding asked for.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Took the convention half, and it turned out to need a second change to be safe -- fixed in 13cd58df.

On the premise first: `signer_id` comes from `DefusePayload::signer_id`, a near-sdk `AccountId`, whose validation only admits lowercase, so `0X...` can't parse into one and never reaches this function. So there's no reachable gap today.

The repo-convention point stands on its own, though, and the function now takes the prefix through `visualsign::encodings::split_hex_prefix` per the unified hex rule in CLAUDE.md.

The part worth calling out: making the prefix case-insensitive on its own would have made things *worse*. `implied_account_id` comes from `to_implicit_account_id()`, which emits lowercase, and the derivation check was a case-sensitive `id == implied_account_id`. Accepting `0X...` into the implicit branch without also comparing case-insensitively would have turned "no claim made" into a false `the signing key does not derive signer_id` -- a scarier statement than staying silent. The comparison now uses `eq_ignore_ascii_case`, with a test asserting a differently-cased spelling of the same account reads as a match and raises no account-binding finding.

src/chain_parsers/visualsign-near/src/tx.rs
shahan-khatchadourian-anchorage: Agreed, and leaving it as-is per your own conclusion -- the trade-off is deliberate.

Storing the parsed `DefusePayload` would keep defuse types in `NearTransaction`'s public shape, which is the boundary this crate is built around (only `presets/intents/render.rs` and `verify.rs` read defuse types). Re-parsing at render time is the cost of that boundary. Token metadata is bounded separately at 1KiB per entry, and the envelope itself is bounded by the CLI reader discussed in the size-limit thread.

src/chain_parsers/visualsign-near/src/convert.rs
shahan-khatchadourian-anchorage: Accurate for this PR read alone -- the override layer is populated one PR up, in #432, at `convert.rs`'s `token_registry_for` via `LayeredRegistry::with_request`, gated by `try_extract_token_metadata_from_chain_metadata` against `authorized_token_metadata_signers` and the `MetadataTrustPolicy`.

Worth noting why it can't land earlier: what makes accepting caller-supplied symbol/decimals safe at all is the signature-verification layer (`token_signature.rs`, ~1200 lines, plus the proto field and the shared prehash/tag additions in `visualsign::signing`). Wiring the layer here without that would mean accepting unauthenticated display strings for token amounts -- the failure class this review round has otherwise been closing.

The delta in #432 at the call site is two lines (`LayeredRegistry::new(...)` -> `token_registry_for(options, trust_policy)`), because the interface this PR ships is the final one -- every `registry: &Reg` signature and `tokens::resolve`'s lookup are unchanged by #432. `tokens.rs`'s own tests already exercise the request layer through `with_request` here.

src/Cargo.toml
shahan-khatchadourian-anchorage: Thanks -- and this is the follow-up you flagged on #426 (3706262026), where you noted the defuse git pins belonged in this review rather than that one.

The closure numbers there are worth reading alongside this: the `near-primitives` investigation on #426 (3723383517) put the full graph at 176 crates, checked `near-api-types` as a slimmer alternative and rejected it because its `Action` discriminants disagree with nearcore at 12/13/14, and found that bumping to 0.37 terminates in a toolchain bump that changes the attested enclave measurement.

Agreed on the framing: exact tag plus commit is the mitigation that's available here, and `defuse-crypto::SignedPayload::verify()` being the core cryptographic dependency is the part worth keeping visible. Happy to fold the defuse surface into the same tracking issue as the `near-primitives` investigations so it outlives these threads.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Good catch, fixed in 13cd58df -- and the contract agrees with you more strongly than the finding suggests.

`defuse_core::intents::token_diff` refuses both cases itself: `if self.diff.is_empty() { return Err(DefuseError::InvalidIntent) }`, and inside the loop `if *delta == 0 { return Err(DefuseError::InvalidIntent) }`. So a zero-delta entry can't execute on-chain, which makes rendering it as `Receive 0 <token>` a line claiming a movement that cannot happen.

`render_token_diff` now refuses both as `VisualSignError::ValidationError` rather than rendering them, matching the call Prasanna's `render_mt_withdraw` length-mismatch comment settled on. Two tests cover it (zero delta names the offending asset id; empty diff reports no entries).

Also confirming your `i128::unsigned_abs()` note -- that's why the refusal is on the delta rather than on the formatting.

src/chain_parsers/visualsign-near/src/presets/intents/verify.rs
shahan-khatchadourian-anchorage: Thanks for working through the contract -- agreed on all of it, and nothing to change.

One small addition to the audit: the `MalformedEncoding`-before-`verify()` ordering you note is load-bearing for the reason your other thread raises, since near-crypto's `ecrecover` backend panics on an out-of-range recovery id rather than returning `None`. So that check isn't just a nicer error, it's what keeps a malformed payload from crashing the signing path.

pepe-anchor (approved):

PR #436, opened 2026-08-04
build(fmt): extend make fmt to the Solana preset tree

### LINKED ISSUE #434: ci: cargo fmt cannot reach the Solana preset tree; 11 preset files are misformatted
## Context

`cargo fmt` does not reach any file under `src/chain_parsers/visualsign-solana/src/presets/*/`.

`presets/mod.rs` declares its preset modules through a generated include:

```rust
include!(concat!(env!("OUT_DIR"), "/generated_presets_mod.rs"));
```

rustfmt cannot follow an `include!` of a path that only exists at build time, so it formats `presets/mod.rs` and stops there. Confirmed with:

```
cargo fmt -p visualsign-solana -- --check --verbose
```

which lists `presets/mod.rs` and no file beneath it.

## Impact

CI's formatting gate is `make generated` followed by an "ensure working tree is clean" check. Because `cargo fmt` never descends into the preset tree, that gate gives zero formatting coverage over every preset, and misformatted preset code passes CI indefinitely.

Running rustfmt directly over the tree on current `main`:

```
11 of 60 preset files fail rustfmt --check

  presets/drift/mod.rs
  presets/exponent_finance/mod.rs
  presets/jupiter_borrow/mod.rs
  presets/jupiter_earn/mod.rs
  presets/jupiter_perps/mod.rs
  presets/kamino_borrow/mod.rs
  presets/meteora_damm_v2/mod.rs
  presets/meteora_dlmm/mod.rs
  presets/neutral_trade/mod.rs
  presets/orca_whirlpool/mod.rs
  presets/swig_wallet/mod.rs
```

Reproduce:

```bash
cd src/chain_parsers/visualsign-solana/src
for f in $(find presets -name '*.rs' | sort); do
  rustfmt --edition 2024 --check "$f" >/dev/null 2>&1 || echo "MISFORMATTED: $f"
done
```

PR #381 hit this: a struct deletion left a stray double blank line, and 17 `create_text_field("Raw Data", ...)` calls stayed wrapped across four lines each after their argument got shorter. CI reported the branch green throughout.

## Proposed fix

1. Extend the fmt target in `src/Makefile` to sweep the preset tree explicitly, e.g. a `rustfmt --edition 2024` pass over `presets/*/*.rs` alongside `cargo fmt --all`, and have `make lint` (or the CI clean-tree step) check the same set.
2. Reformat the 11 files above in the same change so the gate starts from a clean tree.

The `include!(OUT_DIR)` indirection is load-bearing for preset registration, so the fix belongs in the fmt tooling rather than in how modules are declared.

An explicit list or a plain glob is preferable to anything clever here -- the set of preset directories should be visible in the Makefile and in CI logs.

shahan-khatchadourian-anchorage:
## Summary

Closes #434.

`cargo fmt` cannot reach any file under `chain_parsers/visualsign-solana/src/presets/*/`. `presets/mod.rs` declares its submodules through a generated include:

```rust
include!(concat!(env!("OUT_DIR"), "/generated_presets_mod.rs"));
```

rustfmt does not follow an `include!` of a path that exists only after codegen, so it formats `presets/mod.rs` and stops. CI's formatting gate is `make -C src generated` (which calls `make fmt`) followed by an "Ensure working tree is clean" check, so the gate inherits the same blind spot and every preset is unchecked.

This PR:

1. Enumerates the preset sources in the `fmt` target so `make fmt` -- and therefore the existing CI gate -- covers them. No new CI step.
2. Reformats the 11 preset files the gate never saw.
3. Adds `src/rustfmt.toml` pinning `edition = "2024"`, so running `rustfmt <file>` directly on a preset produces the same result as `cargo fmt` and the Makefile. Bare rustfmt otherwise defaults to edition 2015.

`find` rather than a literal file list: presets are scaffolded one directory per program and the set grows with each new one, so a hardcoded list silently drops files out of the gate as it goes stale. The recipe echoes the file count (`rustfmt 60 Solana preset sources`) so a coverage regression is visible in CI logs without dumping 60 paths on one line. The target fails loudly if the directory ever moves.

The `find` is recursive rather than `presets/*/*.rs` because three files live one level deeper, under `presets/*/tests/`.

## Files reformatted

```
drift, exponent_finance, jupiter_borrow, jupiter_earn, jupiter_perps,
kamino_borrow, meteora_damm_v2, meteora_dlmm, neutral_trade,
orca_whirlpool, swig_wallet
```

Every hunk joins lines that now fit within the width limit -- pure re-wrapping, no tokens change. Largest single change is `swig_wallet/mod.rs` at -13 net lines.

## Test plan

- [x] `make -C src generated` then `git status --porcelain` is empty -- reproduces CI's gate exactly and it now passes
- [x] `make -C src fmt` is idempotent (second run produces no diff)
- [x] All 60 preset files pass `rustfmt --check`; before this change 11 failed
- [x] `make -C src lint` -- clippy clean with `-D warnings`
- [x] `cargo test -p visualsign-solana` -- passes

## Verifying the gate actually bites

Introduce a formatting error in any preset and confirm CI now catches it:

```bash
# add a stray blank line to e.g. presets/drift/mod.rs, then:
make -C src generated && git status --porcelain   # non-empty -> gate fails the build
```

## Note

PR #381 is where this surfaced: a struct deletion left a stray double blank line and 17 `create_text_field("Raw Data", ...)` calls stayed wrapped after their argument got shorter, with CI green throughout. That PR touches only `presets/spl_token/mod.rs`, which is not in the list above, so there is no conflict with this branch and no ordering requirement between them.


prasanna-anchorage (approved):
Small, mechanical, LGTM. One nit, non-blocking:

`SOLANA_PRESET_SOURCES` is defined with `:=` (simply-expanded), so the `find` over the presets tree runs at Makefile-parse time on *every* invocation of `make` — `build`, `test`, `lint`, `narrow-build-check` — not just `fmt`, even though only `fmt` references it. Cheap today, but wasted work on every unrelated target and it'll only get pricier as the preset tree grows. Switching to a recursively-expanded `=` variable (or computing it inside the `fmt` recipe) would confine the `find` to when `fmt` actually runs.

Approving.

PR #437, opened 2026-08-04
feat(near): add a parser_cli token-metadata flag for NEAR

shahan-khatchadourian-anchorage:
Adds the NEAR token-metadata flag `parser_cli` was missing, closing the CLI
parity gap with Ethereum (`--abi-json-mappings`) and Solana
(`--idl-json-mappings`).

Depends on #432 (`near-d-token-metadata`), which adds the
`ChainMetadata.near` token-metadata plumbing this flag feeds.

Based on #463 rather than #432 directly. That is merge sequencing, not a
dependency: #463 only adds seed entries and touches no file this PR does, and it
sits lower because it is ready to merge while this one is still in review.

## The gap

`NearPlugin::create_metadata` accepts only `--network`. An asset outside the
compiled-in seed table renders its raw base-unit amount tagged
`unresolved <asset id>`, with no way to supply a symbol and decimals locally:

```
└─ Amount: 250000000 (unresolved nep141:my-token.near)
```

## The flag

```bash
echo '{"symbol":"MYTOKEN","decimals":8}' > mytoken.json

parser_cli decode --chain near --output human \
  --near-token-metadata-mappings 'MyToken@mytoken.json@nep141:my-token.near' \
  -t '{"signer_id":"alice.near", ...}'
```

```
└─ Amount: 2.5 MYTOKEN
```

Two things differ from the sibling flags:

**`@` separates the fields, not `:`.** NEAR Intents asset ids embed their own
colons (`nep141:wrap.near`), so the shared colon-delimited
`mapping_parser::parse_mapping` would truncate the id at its first embedded
colon. `parse_near_mapping` splits on `@` into exactly three parts and takes
the asset id verbatim. A regression test pins this.

**Each entry is signed with the CLI dev key.** The NEAR plugin installs the
strict `MetadataTrustPolicy::RequireAllowlistedSigner` posture, carrying the
env-configured curator allowlist the decode path checks against, so an entry
the parser cannot attribute to a curator is dropped —
signing is what makes the flag do anything at all. This follows Ethereum's
fuller template (mappings + signing) rather than Solana's unsigned-only one.
Three pieces make that work:

- `sign_token_metadata_for_cli` is decoupled from the `dev-signing` cargo
  feature the same way `sign_abi_for_cli` is, so `cli_plugin` still compiles
  in a `cli-plugin`-without-`dev-signing` build (verified explicitly).
- `authorized_token_metadata_signers` enrolls that dev key under
  `dev-signing`/`cfg(test)`, matching `visualsign-ethereum`'s
  `authorized_abi_signers`. Without it the CLI would sign entries its own
  decode path then rejects as an untrusted signer.
- `parser_cli`'s `near` feature enables `visualsign-near/dev-signing`, as its
  `ethereum` feature already does for `visualsign-ethereum`.

`parser_app` enables neither `dev-signing` nor `diagnostics`, so the enclave
binary carries no key material and no allowlist entry trusting the dev key.
`make build` splits `parser_cli` out of the workspace build to keep Cargo
feature unification from crossing that line, and the release image builds from
`parser/app` alone; this PR extends the Makefile comment enumerating those
features, which no longer listed all of them.

CLI-signed entries are always NEAR-origin (`origin_chain` unset).
Ethereum/Solana-origin CLI signing is not wired up.

## Composition with `--network`

The flag composes with `--network` rather than replacing it. An invalid
network still errors before any mapping file is read, so a bad `--network`
can't be masked by a successful mapping load, and metadata is emitted when
either input yields something. `None` is returned only when neither does.

`--network` also decides what the signatures are valid for. A token-metadata
signature is scoped to one NEAR network, so entries are signed for the network
the parser will resolve for the same request: the flag's value when given,
otherwise the network the converter defaults to. `--network NEAR_TESTNET`
therefore produces testnet-scoped entries, and the same mappings signed for one
network do not verify against the other.

## Coverage

- 11 plugin-level cases: composition with `--network` in both directions, the
  colon-in-asset-id regression, duplicate asset ids, and partial failure
  (a malformed mapping and a missing file alongside a good one).
- One end-to-end case proving a CLI-signed entry resolves through the posture
  `register` installs. This is the gate that fails if the dev key leaves the
  allowlist or the signing domain tag drifts — assertions on
  `signature.is_some()` and on unsigned-entry refusal both pass in that case.
- Two `parser_cli` tests drive the real binary, so clap exposure and the
  `dev-signing` feature wiring are covered, not just direct `NearArgs`
  construction.
- Docs: the new flag in `docs/parser-cli.mdx` (whose `--chain` row also did
  not list `near`) and a worked example in `docs/chains/near.mdx`, both run
  against the built binary before being written down.

Known follow-up, raised in review and not addressed here:
`build_token_mappings_from_files` reimplements the load/parse/sign/dedupe/count/
log loop that `parser_cli_core::mapping_parser::load_mappings` provides for
Ethereum, rather than generalizing that helper to take a pluggable delimiter. A
fix to `load_mappings`'s dedup or logging will not reach this copy.

`make lint` and `make test` are green across the workspace.

### DIFF
--- a/docs/chains/near.mdx
+++ b/docs/chains/near.mdx
@@ -128,8 +128,12 @@ cargo run --bin parser_cli -- decode \
   --chain near \
   --output human \
   --near-token-metadata-mappings 'MyToken@mytoken.json@nep141:my-token.near' \
-  -t '{"signer_id":"alice.near", ...}'
+  -t '{"signer_id":"alice.near","verifying_contract":"intents.near","deadline":"2100-01-01T00:00:00Z","nonce":"XVoKfmScb3G+XqH9ke/fSlJ/3xO59sNhCxhpG821BH8=","intents":[{"intent":"ft_withdraw","token":"my-token.near","receiver_id":"bob.near","amount":"100000000"}]}'
 ```
+
+The envelope withdraws `my-token.near`, the asset the mapping covers, so the
+`Amount` field renders as `1 MYTOKEN` rather than `100000000 (unresolved
+nep141:my-token.near)`.
 
 Two details differ from the Ethereum and Solana mapping flags:
 
--- a/docs/chains/near.mdx
+++ b/docs/chains/near.mdx
@@ -117,6 +117,27 @@ The `--network` flag selects the rendered `Network` header — `NEAR_MAINNET` (t
 
 The resolved network also has to agree with the accounts being rendered: an account whose suffix contradicts it (`.testnet` under mainnet, or `.near` under testnet) is refused rather than displayed under a network it does not belong to. This covers the transaction's own accounts and every rendered envelope's `signer_id`/`verifying_contract`. Since the default is mainnet, decoding a `.testnet` envelope needs `--network NEAR_TESTNET` explicitly. Implicit (64-hex) accounts carry no suffix and are not constrained.
 
+### Supplying token metadata locally
+
+Assets the compiled-in seed table doesn't cover render their raw base-unit amount tagged `unresolved <asset id>`. To give the CLI a symbol and decimals for one, pass `--near-token-metadata-mappings`:
+
+```bash
+echo '{"symbol":"MYTOKEN","decimals":8}' > mytoken.json
+
+cargo run --bin parser_cli -- decode \
+  --chain near \
+  --output human \
+  --near-token-metadata-mappings 'MyToken@mytoken.json@nep141:my-token.near' \
+  -t '{"signer_id":"alice.near", ...}'
+```
+
+Two details differ from the Ethereum and Solana mapping flags:
+
+- **`@` separates the fields, not `:`.** NEAR Intents asset ids contain their own colons (`nep141:wrap.near`), so a colon-delimited format would truncate the id at its first embedded colon.
+- **The CLI signs each entry it loads** with its local development key, because it runs the same require-signed posture as the deployed parser — an unsigned entry is dropped rather than trusted. That key is allowlisted only in builds that enable the `dev-signing` feature, which `parser_cli` does and the enclave binary does not.
+
+Repeat the flag for each asset. A mapping that fails to parse or whose file can't be read is reported and skipped; the rest still load.
+
 ### Configuring curator keys
 
 Signer identity is resolved per origin chain from three environment variables, each a comma-separated list of hex public keys:
--- a/docs/chains/near.mdx
+++ b/docs/chains/near.mdx
@@ -134,7 +134,7 @@ cargo run --bin parser_cli -- decode \
 Two details differ from the Ethereum and Solana mapping flags:
 
 - **`@` separates the fields, not `:`.** NEAR Intents asset ids contain their own colons (`nep141:wrap.near`), so a colon-delimited format would truncate the id at its first embedded colon.
-- **The CLI signs each entry it loads** with its local development key, because it runs the same require-signed posture as the deployed parser — an unsigned entry is dropped rather than trusted. That key is allowlisted only in builds that enable the `dev-signing` feature, which `parser_cli` does and the enclave binary does not.
+- **The CLI signs each entry it loads** with its local development key, because it runs the require-signed posture — an unsigned entry is dropped rather than trusted. That key is allowlisted only in builds that enable the `dev-signing` feature, which `parser_cli` does and the enclave binary does not. The posture is a per-deployment choice, not a property of the parser: `parser_cli` pins require-signed, while `parser_app` registers the converter with the permissive default.
 
 Repeat the flag for each asset. A mapping that fails to parse or whose file can't be read is reported and skipped; the rest still load.
 
--- a/docs/chains/near.mdx
+++ b/docs/chains/near.mdx
@@ -114,6 +114,27 @@ Output:
 
 The `--network` flag selects the rendered `Network` header — `NEAR_MAINNET` (the default) or `NEAR_TESTNET`. An unrecognized value errors before anything renders, rather than silently falling back to mainnet.
 
+### Supplying token metadata locally
+
+Assets the compiled-in seed table doesn't cover render their raw base-unit amount tagged `unresolved <asset id>`. To give the CLI a symbol and decimals for one, pass `--near-token-metadata-mappings`:
+
+```bash
+echo '{"symbol":"MYTOKEN","decimals":8}' > mytoken.json
+
+cargo run --bin parser_cli -- decode \
+  --chain near \
+  --output human \
+  --near-token-metadata-mappings 'MyToken@mytoken.json@nep141:my-token.near' \
+  -t '{"signer_id":"alice.near", ...}'
+```
+
+Two details differ from the Ethereum and Solana mapping flags:
+
+- **`@` separates the fields, not `:`.** NEAR Intents asset ids contain their own colons (`nep141:wrap.near`), so a colon-delimited format would truncate the id at its first embedded colon.
+- **The CLI signs each entry it loads** with its local development key, because it runs the same require-signed posture as the deployed parser — an unsigned entry is dropped rather than trusted. That key is allowlisted only in builds that enable the `dev-signing` feature, which `parser_cli` does and the enclave binary does not.
+
+Repeat the flag for each asset. A mapping that fails to parse or whose file can't be read is reported and skipped; the rest still load.
+
 ## Implementation details
 
 Source code available at:
--- a/src/chain_parsers/visualsign-near/src/cli_plugin.rs
+++ b/src/chain_parsers/visualsign-near/src/cli_plugin.rs
@@ -1,19 +1,34 @@
+use std::collections::BTreeMap;
+
 use clap::Args as ClapArgs;
-use generated::parser::{ChainMetadata, NearMetadata, chain_metadata};
+use generated::parser::{ChainMetadata, NearMetadata, TokenMetadataEntry, chain_metadata};
 use visualsign::registry::{Chain, TransactionConverterRegistry};
 
+use parser_cli_core::mapping_parser::load_json_file;
+
 use crate::networks::NearNetwork;
+use crate::presets::intents::sign_token_metadata_for_cli;
 
 /// CLI arguments specific to NEAR.
-///
-/// Currently no NEAR-specific args are needed beyond the global `--network`
-/// flag, which `create_metadata` below turns into `NearMetadata`.
 #[derive(ClapArgs, Debug, Default, Clone)]
-pub struct NearArgs {}
+pub struct NearArgs {
+    /// Map custom token metadata JSON file to a NEAR Intents asset id.
+    ///
+    /// Format: `Name@/path/to/token.json@AssetId`. `@` is the field
+    /// separator here, not `:` (the convention `--abi-json-mappings`/
+    /// `--idl-json-mappings` use), because NEAR Intents asset ids embed
+    /// their own colons (e.g. `nep141:wrap.near`), which would make the
+    /// identifier ambiguous under a colon-delimited format. The file should
+    /// contain `{"symbol":"...","decimals":N}`. Can be used multiple times.
+    #[arg(
+        long = "near-token-metadata-mappings",
+        value_name = "NAME@FILE_PATH@ASSET_ID"
+    )]
+    pub near_token_metadata_mappings: Vec<String>,
+}
 
 /// [`parser_cli_core::ChainPlugin`] implementation for NEAR.
 pub struct NearPlugin {
-    #[allow(dead_code)]
     args: NearArgs,
 }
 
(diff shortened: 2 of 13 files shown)


pepe-anchor (commented):
**Peer Review Summary — NEAR stack #437** (COMMENT)\n\n5 findings: 5 LOW. Detailed inline comments below.\n\n**Key concern:** @ in file paths breaks mapping parsing. Dev-signing edge cases.

---

src/chain_parsers/visualsign-near/src/cli_plugin.rs
```diff
@@ -51,28 +66,160 @@ impl parser_cli_core::ChainPlugin for NearPlugin {
...
+/// `mapping_parser::parse_mapping`, which splits on `:` and would truncate a
+/// NEAR asset id at its first embedded colon).
+fn parse_near_mapping(mapping_str: &str) -> Result<NearMappingComponents, String> {
+    let mut parts = mapping_str.splitn(3, '@');
+    let (Some(name), Some(path), Some(asset_id)) = (parts.next(), parts.next(), parts.next())
+    else {
+        return Err(format!(
+            "Invalid mapping format (expected Name@FilePath@AssetId): {mapping_str}"
+        ));
+    };
```
pepe-anchor, line 106: **[LOW] `@` in file paths breaks `Name@FilePath@AssetId` parsing because `splitn(3, '@')` consumes the path component greedily, and the error message doesn't surface which component is the problem**

`parse_near_mapping` uses `splitn(3, '@')` which splits on the first two `@` characters regardless of position. A file path containing `@` (valid on Linux/macOS) causes mis-parsing: `/tmp/my@dir/token.json` yields `name='/tmp/my', path='dir/token.json', asset_id='<rest>'` instead of a clear error. The format is documented as `Name@FilePath@AssetId` and the help text expects `@/path/to/file.json@AssetId` (implying absolute paths with a leading `@` that disambiguates), but a relative path with an embedded `@` may still confuse users — consider adding a validation step that rejects a path component containing `@` and points to the third `@` as the intended separator.

---

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
```diff
@@ -538,6 +566,56 @@ pub fn sign_token_metadata_secp256k1(
...
+#[cfg(any(test, feature = "dev-signing"))]
+pub fn sign_token_metadata_for_cli(
+    asset_id: &str,
+    value: &str,
+) -> Result<generated::parser::SignatureMetadata, String> {
+    Ok(sign_token_metadata_ed25519(
+        asset_id,
+        value,
+        &DEV_NEAR_SIGNING_KEY_SEED,
+        visualsign::signing::near_token_metadata_prehash,
```
pepe-anchor, line 639: **[LOW] `DEV_NEAR_SIGNING_KEY_SEED` is a deterministic all-0x51 seed — consistent with `DEV_ETHEREUM_SIGNING_KEY_SEED` (0x52) and `DEV_SOLANA_SIGNING_KEY_SEED` (0x53) but worth re-confirming these are intentionally non-random for dev-only use and would never reach a production enclave binary**

The seed `[0x51u8; 32]` is hardcoded, gated behind `#[cfg(any(test, feature = "dev-signing"))]`, and allowlisted only under the same cfg. The split-build Makefile strategy prevents unification, so `parser_app` never links the key. However, if anyone changes the build to a single `cargo build --workspace` invocation (e.g. for a CI optimization), the key and its allowlist entry would silently leak into the production enclave — there's no compile-time assertion or build-script check that `dev-signing` is absent from the `parser_app`/`grpc-server` dependency graph. Consider adding a compile-time guard (e.g. a `compile_error!` in a non-dev-signing `parser_app` module that depends on `visualsign-near` but is `#[cfg(not(dev_signing))]`) or a CI-only check in `narrow-build-check`.

---

src/chain_parsers/visualsign-near/src/cli_plugin.rs
```diff
@@ -51,28 +66,160 @@ impl parser_cli_core::ChainPlugin for NearPlugin {
...
+            Err(e) => {
+                eprintln!(
+                    "  Warning: '{}' could not be signed ({e}); registering unsigned",
+                    components.name
+                );
+                None
+            }
         };
-        if NearNetwork::from_network_id(&network).is_none() {
-            return Err(format!(
```
pepe-anchor, line 58: **[LOW] `create_chain_metadata` returns `Ok(None)` when all mappings fail to load, but `test_cli_near_token_metadata_invalid_file_still_parses` only covers the file-not-found case — add a test for a file that exists but contains invalid JSON (e.g. `{`) to ensure the `serde_json` parse failure path in `load_json_file` is exercised**

`load_json_file` returns `Err(...)` when `serde_json::from_str::<Value>` fails (invalid JSON). The CLI test `test_cli_near_token_metadata_invalid_file_still_parses` uses `/nonexistent/token.json` which exercises the `File::open` error path but not the `serde_json` parse error path. A separate test with a real but syntactically invalid JSON file would close this coverage gap — important because the error message shape (`Invalid JSON in file {path}: ...`) is user-visible and worth snapshotting.

---

src/parser/cli/tests/cli_test.rs
```diff
@@ -634,3 +634,80 @@ fn test_cli_solana_idl_invalid_file_still_parses() {
...
+        token_path.display()
+    );
+
+    let (stdout, stderr) = run_cli_full(&[
+        "decode",
+        "--chain",
+        "near",
+        "--output",
+        "json",
+        "--near-token-metadata-mappings",
```
pepe-anchor, line 670: **[LOW] CLI integration tests exercise the happy path and file-not-found error, but not the `dev-signing`-disabled error path — add a test with an intentionally disabled feature to catch the regression where `sign_token_metadata_for_cli` returns `Err` and the entry loads unsigned then gets silently dropped**

The two new CLI tests (`test_cli_near_token_metadata_mappings` and `test_cli_near_token_metadata_invalid_file_still_parses`) cover happy-path resolution and file-not-found. But neither exercises what happens when `dev-signing` is absent: `sign_token_metadata_for_cli` returns `Err(...)`, `build_token_mappings_from_files` inserts the entry unsigned, and `RequireAllowlistedSigner` drops it. A `#[cfg(not(feature = "dev-signing"))]` test (or a dedicated test binary without the feature) that asserts the entry is dropped (symbol stays unresolved) would prevent a future regression where the feature gate is accidentally removed or broken.

---

---

src/chain_parsers/visualsign-near/src/cli_plugin.rs
shahan-khatchadourian-anchorage: Good catch, fixed in f21237c4 -- and the mis-parse was silent, which made it worse than a bad error message.

`parse_near_mapping` now splits on every `@` and requires exactly three components. A NEAR Intents asset id never contains `@`, so a fourth component can only mean the path did, and the error says so: it names the component count and states that a path containing `@` can't be used in this format.

Before this, `MyToken@/tmp/my@dir/token.json@nep141:wrap.near` parsed as path `/tmp/my` and asset id `dir/token.json@nep141:wrap.near` with no complaint at all -- it would then fail as a missing file, pointing at the wrong thing. Test `parse_near_mapping_rejects_an_at_sign_in_the_path` pins that exact string.

On the leading-`@` reading of the help text: the `@` there is the separator, not part of the path, so absolute and relative paths are both fine -- it's only an embedded `@` that's unrepresentable.

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
shahan-khatchadourian-anchorage: Confirmed non-random and dev-only, and you're right that nothing enforced it -- added a CI gate in f21237c4.

`narrow-build-check` now runs `dev-signing-absent-check`, which fails if any `--workspace` recipe in `src/Makefile` lacks `--exclude parser_cli`. That's the actual invariant: Cargo unifies features within a single invocation, so the only thing keeping `dev-signing` off `parser_app`/`grpc-server` is that `build` compiles `parser_cli` separately. A future `cargo build --workspace` covering both now fails the build rather than quietly linking a dev key into an attested binary.

Worth recording why it isn't the graph probe you suggested, since I tried that first: `cargo tree -p parser_app -e features` does not print enabled features that carry no dependency edges, and `dev-signing = []` is exactly that. The probe matched nothing even against `parser_cli`, which *does* enable it -- so it would have passed unconditionally and looked like coverage. I verified the replacement fires by injecting a violating recipe.

It also covers `visualsign-ethereum/dev-signing`, which has the same exposure and predates NEAR.

src/chain_parsers/visualsign-near/src/cli_plugin.rs
shahan-khatchadourian-anchorage: Agreed, added in f21237c4. `build_token_mappings_skips_a_file_whose_json_is_invalid` writes a real file containing `{`, so `File::open` succeeds and the failure comes from `serde_json` in `load_json_file` -- the branch the existing `/nonexistent/token.json` test can't reach.

Asserts both halves of the outcome: the returned map is empty and `valid_count` is 0, so a future change that logged the error but still registered the entry would fail here.

src/parser/cli/tests/cli_test.rs
shahan-khatchadourian-anchorage: Added in 28344600 -- and my first read of this was wrong, so worth spelling out why it is testable.

You're right that `#[cfg(not(feature = "dev-signing"))]` doesn't work from the crate's own test module: the gate is `cfg(any(test, feature = "dev-signing"))`, so under `cargo test` the signing variant is always linked and its `Err` arm is unreachable. But an **integration test** compiles the library without `cfg(test)`, so with `dev-signing` off the error-returning twin at `token_signature.rs:646` is what links -- no separate test binary needed.

`tests/unsigned_without_dev_signing.rs` drives the real public path (`NearPlugin::create_metadata` with a mapping file) and asserts the entry registers with `signature: None` and its value carried verbatim -- i.e. unsigned rather than dropped, which is the half nothing else covered. I checked it isn't vacuous two ways: it reports `1 passed` rather than being silently `cfg`-ed out, and with the gate temporarily removed and `--features dev-signing` on it fails against a real ed25519 signature.

`make test` already runs it in the right configuration -- the workspace pass excludes `parser_cli`, the only crate that enables `dev-signing`, so unification can't switch it back on.

Scoped deliberately to the unsigned-registration half. Asserting the subsequent drop would need `#439`'s `TokenMetadataExtraction` API, which doesn't exist on this PR -- and that behaviour is already covered by `token_signature.rs`'s tests, which don't depend on this feature.

prasanna-anchorage (dismissed):
Reviewed the `parser_cli` NEAR token-metadata flag, the `@`-delimited mapping parser, and the dev-key signing/allowlist wiring (`dev-signing` feature gating for `cli_plugin`/`parser_cli`, `parser_app` carrying neither `dev-signing` nor `diagnostics`). Checked the colon-in-asset-id regression test and the `--network` composition logic. All correct.

One non-blocking finding: `build_token_mappings_from_files` (with `parse_near_mapping`) in `src/chain_parsers/visualsign-near/src/cli_plugin.rs:232` hand-reimplements the generic load/parse/sign/dedupe/count/log loop that `parser_cli_core::mapping_parser::load_mappings` already provides for Ethereum's `build_abi_mappings_from_files`, instead of generalizing that shared helper to accept a pluggable parse function/delimiter. Not a bug today, but a future fix to `load_mappings`'s dedup/logging behavior won't propagate to this copy, and there's now no signal to a maintainer that a second copy exists.

Approving — worth a follow-up to generalize `load_mappings`, doesn't need to block this PR.

prasanna-anchorage (dismissed):
Re-approving. My 2026-08-25 approval was dismissed by the stale-review rule, but unlike the rebase-only cases in this stack this PR gained two real commits since, so I re-read the whole thing rather than re-issuing.

## The two post-approval commits

**`b4dfbcce` — the `@` fix is correct, and it's a real bug.** `splitn(3, '@')` did absorb a path's `@` into the asset id silently: `MyToken@/tmp/my@dir/token.json@nep141:wrap.near` parsed as path `/tmp/my` and asset id `dir/token.json@nep141:wrap.near`. Splitting on every `@` and rejecting a fourth component is the right call, the error message names the actual fault ("found N components, expected 3 ... A file path containing '@' cannot be used here"), and `parse_near_mapping_rejects_an_at_sign_in_the_path` pins it against exactly the string that used to pass.

I checked the load-bearing assumption rather than taking it on faith: "an asset id never contains one" holds for everything this crate resolves — `grep nep245\|nep171` across `visualsign-near/src` returns nothing, and `tokens.rs` is entirely `nep141:`. Worth keeping in mind if multi-token ids ever reach `token_mappings`, since a NEP-245 token id is contract-defined and `#460` already names `mt_withdraw` intents; not a concern today.

**`b4dfbcce` — the dev-signing gate.** The reasoning in the commit message is right, including the part that's easy to get wrong: `cargo tree -e features` really does omit enabled features that carry no dependency edges, and `dev-signing = []` is exactly that, so a graph probe would have passed vacuously. Asserting on the recipes instead is a reasonable call.

I verified the surface it's guarding:

- The guard passes clean on the current Makefile, and the three real `--workspace` recipes (lines 35, 44, 72) all carry `--exclude parser_cli`.
- `narrow-build-check` is in CI (`main.yml:64`), so this actually runs.
- No `cargo ... --workspace` invocation exists anywhere outside `src/Makefile` — I grepped the workflows, shell scripts and Containerfiles.
- The shipping binary was never at risk regardless: `images/parser_app/Containerfile:32` builds from `/src/parser/app` with `--no-default-features --features "vsock ${CHAIN_FEATURES}"`, and `parser_app`'s `near = ["dep:visualsign-near"]` doesn't pull `dev-signing` the way `parser_cli`'s does.
- The `insert_near_dev_signer` cfg pair is the right shape — a real fn under `cfg(any(test, feature = "dev-signing"))` and a no-op under `cfg(not(...))` — so with the feature off the dev key is genuinely never enrolled, not merely unused.

**`20ad19b3` — the integration test earns its place.** The header explains precisely why it can't live in the crate's test module: `sign_token_metadata_for_cli` is gated on `cfg(any(test, feature = "dev-signing"))`, so under unit tests the signing variant always links and its `Err` arm is unreachable. An integration test compiles the library without `cfg(test)`, so with `dev-signing` off the error-returning variant is what links — the configuration a shipped binary has. That's the distinction most people miss, and the `#![cfg(all(feature = "cli-plugin", not(feature = "dev-signing")))]` gate makes it self-disabling rather than silently vacuous when the feature is on.

`make lint` and `make test` green locally (40 suites, 1689 tests, both exit 0). I also confirmed the gate doesn't make it vacuous — `without_dev_signing_the_cli_registers_the_entry_unsigned ... ok` appears in the run, so it really does execute under `make test` rather than being cfg'd out, along with `parse_near_mapping_rejects_an_at_sign_in_the_path`, `build_token_mappings_skips_a_file_whose_json_is_invalid` and `test_cli_near_token_metadata_mappings`.

## Non-blocking

**1. The guard is anchored to lines that literally start with `cargo`.** `^[[:space:]]*cargo (build|test|check|clippy).*--workspace` misses any recipe with a prefix — `RUSTFLAGS=... cargo build --workspace`, `time cargo build --workspace`, `cd foo && cargo build --workspace`. It also false-positives on a line-continued recipe, which I confirmed:

```
	cargo build --workspace \
		--exclude parser_cli
```

flags as bad even though it's correct. That direction errs safe, so neither is urgent.

If you want the version that can't be routed around, assert on the resolved feature rather than on recipe text — it then holds for the Containerfile and for a hand-run cargo too:

```rust
// visualsign-near/src/lib.rs
pub const DEV_SIGNING_ENABLED: bool = cfg!(feature = "dev-signing");

// parser/app/src/main.rs
#[cfg(feature = "near")]
const _: () = assert!(
    !visualsign_near::DEV_SIGNING_ENABLED,
    "dev-signing must never resolve in parser_app"
);
```

That's a compile error at the moment unification happens, rather than a lint on one file's text. Happy either way — the grep is a fine stopgap and the shipped path is already safe.

**2. The two new tests hand-roll fixed-name temp dirs, while this same PR uses the collision-safe helper elsewhere.** `cli_test.rs` (added here) calls `write_temp_json("vsp_cli_tests", ...)`, whose doc comment says filenames "include the PID and a nanosecond timestamp to avoid collisions when tests run in parallel" and which also checks the resolved path stays inside the temp dir. But `cli_plugin.rs:462` uses `env::temp_dir().join("vsp_near_cli_plugin_invalid_json")` and `unsigned_without_dev_signing.rs:32` uses `.join("vsp_near_no_dev_signing")`, both ending in `remove_dir_all`.

Within one `make test` they don't collide — the five cargo invocations are sequential and the two names differ. The exposure is a concurrent run on the same machine (a developer running `cargo test -p visualsign-near` while `make test` is going): one test's `remove_dir_all` deletes the other's file mid-read. `parser_cli_core::test_utils::write_temp_json` is `pub` and already reachable from both call sites, so this is a two-line change if you want it.

**3. The original non-blocking finding stands**, and this round is mild evidence for it: `parse_near_mapping` still reimplements `parser_cli_core::mapping_parser::load_mappings` rather than generalizing it over the delimiter. The `@`-in-path bug is exactly the class of thing a single shared splitter gets right once instead of twice.

src/chain_parsers/visualsign-near/src/cli_plugin.rs
```diff
@@ -52,28 +67,184 @@ impl parser_cli_core::ChainPlugin for NearPlugin {
...
+/// dropped by the strict posture [`cli_trust_policy`] installs; if signing is
+/// unavailable (binary built without `dev-signing`), the entry is registered
+/// unsigned instead of dropped here.
+fn build_token_mappings_from_files(
+    mappings: &[String],
+    signing_network: NearNetwork,
+) -> (BTreeMap<String, TokenMetadataEntry>, usize) {
+    let mut map = BTreeMap::new();
+    let mut valid_count = 0;
+
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Taken, in `f218c1ce`. `build_token_mappings_from_files` is now a call to `load_mappings`, and the duplicated loop is gone.

Of your two suggested shapes it had to be the pluggable parser, not a delimiter. The two parsers differ in more than the separator: the colon form splits from the right so a Windows drive letter stays in the path, while the NEAR form rejects a fourth `@` component outright, because an asset id never contains `@` and a path that does would otherwise be split into a wrong path and a wrong asset id. A shared split parameterized only by character would have dropped that error.

So `load_mappings` takes `parse` as a closure, and the four descriptive strings became a `MappingFormat` struct -- six positional arguments, four of them interchangeable `&str`, is a signature where transposing two is a silent bug rather than a compile error, and the delimiter-specific format hint needed somewhere to live anyway.

On the drift risk you name: the file-size limit was already shared, since NEAR called `load_json_file` directly. What was genuinely duplicated was the logging format and the dedupe/count semantics, and those are now single-sourced. Solana and Ethereum keep byte-identical behavior; NEAR's output is unchanged too, except that a duplicate now reports as `Duplicate AssetId` rather than `Duplicate asset id`, matching how the shared loop labels the other two chains' identifiers. I checked the parse-failure path by running it, since that is where the `@`-shaped hint had to survive.

docs/chains/near.mdx
```diff
@@ -117,6 +117,27 @@ The `--network` flag selects the rendered `Network` header — `NEAR_MAINNET` (t
...
+
+```bash
+echo '{"symbol":"MYTOKEN","decimals":8}' > mytoken.json
+
+cargo run --bin parser_cli -- decode \
+  --chain near \
+  --output human \
+  --near-token-metadata-mappings 'MyToken@mytoken.json@nep141:my-token.near' \
+  -t '{"signer_id":"alice.near", ...}'
+```
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed in `ae7c433d`. The example now carries a full envelope, withdrawing the asset the mapping covers so it demonstrates the flag rather than merely invoking it: the amount renders as `1 MYTOKEN` instead of `100000000 (unresolved nep141:my-token.near)`.

Verified by running the command as written, rather than by inspection -- which is what the original was missing.

shahan-khatchadourian-anchorage:

src/chain_parsers/visualsign-near/src/cli_plugin.rs
```diff
@@ -52,28 +67,157 @@ impl parser_cli_core::ChainPlugin for NearPlugin {
...
+            ) {
+                Ok(sig) => Some(sig),
+                Err(e) => {
+                    eprintln!(
+                        "  Warning: '{}' could not be signed ({e}); registering unsigned",
+                        components.name
+                    );
+                    None
+                }
+            };
```
pepe-anchor, line 155: should: A signing failure here still counts the entry as loaded and prints "Loaded", unlike Ethereum's identical `--abi-json-mappings` flow where a signing failure rejects the entry and the success count reflects it.

pepe-anchor (approved):

src/chain_parsers/visualsign-near/src/cli_plugin.rs
shahan-khatchadourian-anchorage: Thanks for the review @pepe-anchor 
Addressing the 'should' in #478

PR #439, opened 2026-08-04
feat(near): report refused token metadata as a diagnostic

shahan-khatchadourian-anchorage:
Turns on the `diagnostics` feature for `visualsign-near` and closes the one gap
that needed new code rather than a flag flip: token-metadata entries the parser
refuses are reported to the signer instead of only to an operator log.

Stacked on #437 (`near-e-cli-token-metadata`).

## What was already in place, and what wasn't

`visualsign-near` declares `diagnostics = ["visualsign/diagnostics"]`, and
`render.rs` routes every soft finding (`signature`, `account-binding`,
`deadline`, `extraction`, `unverified-token-metadata`) through one
feature-aware `diagnostic()` helper. But nothing enabled that feature: neither
`parser_cli`'s nor `parser_app`'s `diagnostics` propagated into the crate, so
the structured half of that helper was dead code.

Refused token-metadata entries were the real gap.
`try_extract_from_chain_metadata` dropped them with a `tracing::warn!` and
returned only the survivors. A caller supplying metadata the parser then threw
away learned nothing from the payload: the signer saw an amount in raw base
units against an `unresolved` asset id, or resolved from a seed instead of the
supplied override, and the reason lived in a log the caller cannot read.

## Changes

**Extraction reports refusals.** It returns
`TokenMetadataExtraction { registry, rejected }`. A `reject!` macro binds each
refusal to its `continue`, so a new rejection path cannot be added without
reporting it — all nine paths (oversized value, unrecognized `origin_chain`,
unattributable under require-signed, unattributable seed override, unrecognized
signer under require-signed, malformed value JSON, decimals out of range, symbol
length, symbol outside printable ASCII) are covered by one table-driven test.
A mapping over the entry cap is refused as a unit and reports one rejection
covering the whole map, rather than one per entry.

**Rendering.** `rejected_metadata_diagnostics` emits a new
`rejected-token-metadata` rule per refused entry, alongside the intents rather
than replacing them — the refusal protects the render, it doesn't invalidate
it. Both halves of the message pass through `charset_safe`: the asset id is a
caller-controlled map key and the reason can quote it back (a JSON parse error,
a length), so an embedded newline would otherwise render as extra apparent
fields on the signing screen. That is the same class already fixed for
`memo`/`msg`/`method_name`, reached through a different field, and it has its
own regression test.

**One registry per conversion.** It was built inside `decode_intents`, once per
action. Since the metadata is request-scoped, a two-action transaction would
have repeated every rejection. Built once in `render_on_chain` now, with a test
pinning the count.

**Feature propagation and CI.** `parser_cli`, `parser_cli_core` and
`parser_app` propagate `diagnostics` to `visualsign-near`. `make test` and
`make lint` gained `-p visualsign-near --features diagnostics --lib` and the
`parser_cli_core` equivalent, matching the convention in
`docs/contributor-guides/lint-diagnostics.mdx`. A `near-json` CLI fixture pair
(`.display.expected` + `.diagnostics.expected`, generated from real output)
covers the rendered form, and the fixture harness's chain skip-list gained the
`near` entry it was missing, so a `--no-default-features` build skips it
correctly.

**The CLI human view flattens diagnostics.** `format_field`'s accessor arm
renders any variant it has no case for from `label()`/`fallback_text()`, which
for a diagnostic drops the level, rule and instruction index — most of what it
carries. A `Diagnostic` arm now prints all of them, which improves Solana's
diagnostics in that view too:

```
   ├─ [warn] rejected-token-metadata: token metadata supplied for nep141:my-token.near was rejected and not used: invalid value JSON: missing field `symbol`
   ├─ [warn] deadline: deadline has passed; the intents would be rejected
```

The arm's body lives in `write_diagnostic` so `format_field` stays within
clippy's line limit — its match already covers every field variant. The arm and
its two tests are gated on a new `parser_cli_core/diagnostics` feature, because
`visualsign` only defines the `Diagnostic` variant under its own.

## Payload shape

A refused entry adds one field where there was none. With `diagnostics` off —
`parser_app`, the production shape HSMs and wallets derive a metadata digest
from — that is a `Warning` text field, matching how NEAR's other soft findings
already degrade. Requests that supply no metadata, or whose metadata is
accepted, are unchanged.

## Docs

`lint-diagnostics.mdx` gains NEAR in the feature list and CI-invocation
description, plus a per-chain rule table. The NEAR section explains the two
ways NEAR differs from Solana's convention: bare rule names rather than
`domain::rule_name`, and a `Warning`-text fallback instead of dropping the
finding when the feature is off.

`make lint` and `make test` are green across the workspace in both feature
states.

### DIFF
--- a/src/chain_parsers/visualsign-near/src/convert.rs
+++ b/src/chain_parsers/visualsign-near/src/convert.rs
@@ -36,11 +36,20 @@ use crate::tx::NearTransaction;
 /// present signature is checked against -- its own under the strict posture,
 /// the deployment's `authorized_token_metadata_signers` under the permissive
 /// one. A present signature is checked against that list under either.
+///
+/// Returns the registry plus a diagnostic field for every entry the extraction
+/// refused. Callers render those once per payload, not once per action.
 fn token_registry_for(
     options: &VisualSignOptions,
     network: NearNetwork,
     trust_policy: &MetadataTrustPolicy,
-) -> LayeredRegistry<NearTokenRegistry> {
+) -> Result<
+    (
+        LayeredRegistry<NearTokenRegistry>,
+        Vec<SignablePayloadField>,
+    ),
+    VisualSignError,
+> {
     // Identity decides whether an entry renders as verified and whether it may
     // override a curated seed, neither of which the posture itself answers, so a
     // list is needed under both postures. The strict posture carries its own;
@@ -59,18 +68,22 @@ fn token_registry_for(
         MetadataTrustPolicy::AcceptUnsigned => authorized_token_metadata_signers(),
         _ => &no_signers,
     };
-    let request = try_extract_token_metadata_from_chain_metadata(
+    let extraction = try_extract_token_metadata_from_chain_metadata(
         options.metadata.as_ref(),
         network,
         allowlist,
         trust_policy,
     );
-    match request {
+    let registry = match extraction.registry {
         Some(request) => {
             LayeredRegistry::with_request(Arc::new(NearTokenRegistry::default()), request)
         }
         None => LayeredRegistry::new(Arc::new(NearTokenRegistry::default())),
-    }
+    };
+    Ok((
+        registry,
+        crate::presets::intents::rejected_metadata_diagnostics(&extraction.rejected)?,
+    ))
 }
 
 /// Resolve the network for one request: the `network_id` the request supplied,
@@ -215,6 +228,14 @@ impl NearVisualSignConverter {
             create_address_field("To", tx.receiver_id().as_str(), None, None, None, None)?
                 .signable_payload_field,
         );
+        // Built once for the whole transaction: the metadata is request-scoped,
+        // so a rejection is a property of the request, not of each action that
+        // consults the registry. Building it per action would repeat every
+        // rejection diagnostic for a multi-action transaction.
+        let (registry, rejection_diagnostics) =
+            token_registry_for(options, network, &self.trust_policy)?;
+        fields.extend(rejection_diagnostics);
+
         let total_actions = tx.actions().len();
         for action in tx.actions() {
             fields.extend(render_action(action, total_actions)?);
@@ -222,8 +243,8 @@ impl NearVisualSignConverter {
                 tx.receiver_id().as_str(),
                 action,
                 options,
+                &registry,
                 network,
-                &self.trust_policy,
             )?);
         }
 
@@ -260,8 +281,8 @@ fn decode_intents(
     receiver_id: &str,
     action: &Action,
     options: &VisualSignOptions,
+    registry: &LayeredRegistry<NearTokenRegistry>,
     network: NearNetwork,
-    trust_policy: &MetadataTrustPolicy,
 ) -> Result<Vec<SignablePayloadField>, VisualSignError> {
     if receiver_id != "intents.near" {
         return Ok(vec![]);
@@ -272,8 +293,7 @@ fn decode_intents(
     if fc.method_name != "execute_intents" {
         return Ok(vec![]);
     }
-    let registry = token_registry_for(options, network, trust_policy);
-    crate::presets::intents::try_decode_execute_intents(&fc.args, &registry, options, network)
+    crate::presets::intents::try_decode_execute_intents(&fc.args, registry, options, network)
         .map_err(intents_error)
 }
 
@@ -297,12 +317,13 @@ fn render_intent_envelope(
     network: NearNetwork,
     trust_policy: &MetadataTrustPolicy,
 ) -> Result<ConversionResult, VisualSignError> {
-    let registry = token_registry_for(options, network, trust_policy);
+    let (registry, rejection_diagnostics) = token_registry_for(options, network, trust_policy)?;
     // The resolved network is part of every token-metadata signed scope, so the
     // payload has to show which network that scope was checked against -- the
     // same field, for the same reason, as the on-chain path renders.
     let mut fields =
         vec![create_text_field("Network", network.display_name())?.signable_payload_field];
+    fields.extend(rejection_diagnostics);
     fields.extend(
         crate::presets::intents::try_render_single_intent(
             json.as_bytes(),
@@ -829,5 +850,172 @@ mod tests {
             assert!(json.contains(expected), "missing {expected}: {json}");
         }
         assert!(!json.contains("Standard"), "unexpected signature section");
+    }
+
+    /// Metadata the parser refuses must be visible in the payload, not just in
+    /// an operator log.
+    ///
+    /// The refusal here is the `require-signed` posture dropping an unsigned
+    /// entry. Without the diagnostic, the signer sees an amount in raw base
+    /// units against an `unresolved` asset id and has no way to tell that
+    /// metadata was supplied at all.
+    #[test]
+    fn rejected_metadata_surfaces_a_diagnostic_end_to_end() {
+        let asset_id = "nep141:rejected-token.near";
+        let options = VisualSignOptions {
+            metadata: Some(ChainMetadata {
+                metadata: Some(chain_metadata::Metadata::Near(NearMetadata {
+                    network_id: Some("NEAR_MAINNET".to_string()),
+                    token_mappings: [(
+                        asset_id.to_string(),
+                        generated::parser::TokenMetadataEntry {
+                            value: r#"{"symbol":"DROPPED","decimals":6}"#.to_string(),
+                            signature: None,
+                            origin_chain: None,
+                        },
+                    )]
+                    .into_iter()
+                    .collect(),
+                })),
+            }),
+            ..Default::default()
+        };
+
+        let envelope = r#"{"signer_id":"alice.near","verifying_contract":"intents.near","deadline":"2999-01-01T00:00:00Z","nonce":"XVoKfmScb3G+XqH9ke/fSlJ/3xO59sNhCxhpG821BH8=","intents":[{"intent":"ft_withdraw","token":"rejected-token.near","receiver_id":"bob.near","amount":"1000000"}]}"#.to_string();
+        let payload = NearVisualSignConverter::with_trust_policy(
+            MetadataTrustPolicy::RequireAllowlistedSigner(SignerAllowlist::new()),
+        )
+        .to_visual_sign_payload(NearTransaction::Intent(envelope), options)
+        .expect("convert");
+        let fields = &payload.payload.fields;
+
+        assert!(
+            fields.iter().any(
+                |f| crate::presets::intents::test_support::is_warning_diagnostic(
+                    f,
+                    "rejected-token-metadata"
+                )
+            ),
+            "a refused entry must report itself in the payload: {fields:?}"
+        );
+        let json = payload.payload.to_json().expect("json");
+        assert!(
+            !json.contains("DROPPED"),
+            "the refused entry must not resolve the symbol: {json}"
+        );
+        assert!(
+            json.contains(asset_id),
+            "the diagnostic must name the asset id it refused: {json}"
+        );
+    }
+
+    /// The rejection is a property of the request's metadata, so it reports
+    /// once even when several actions consult the registry.
+    #[test]
+    fn rejected_metadata_reports_once_for_a_multi_action_transaction() {
+        let asset_id = "nep141:rejected-token.near";
+        let inner = r#"{"signer_id":"alice.near","verifying_contract":"intents.near","deadline":"2999-01-01T00:00:00Z","nonce":"XVoKfmScb3G+XqH9ke/fSlJ/3xO59sNhCxhpG821BH8=","intents":[{"intent":"ft_withdraw","token":"rejected-token.near","receiver_id":"bob.near","amount":"1000000"}]}"#;
+        let args = serde_json::json!({"signed":[{
+            "standard": "raw_ed25519",
+            "payload": inner,
+            "public_key": "ed25519:8rVvtHWFr8hasdQGGD5WiQBTyr4iH2ruEPPVfj491RPN",
+            "signature": "ed25519:3vtbNQJHZfuV1s5DykzyjkbNLc583hnkrhTz57eDhd966iqzkor6Twgr4Loh2C195SCSEsiGfrd6KcxpjNq9ZbVj"
+        }]});
+        let call = || {
+            Action::FunctionCall(Box::new(FunctionCallAction {
+                method_name: "execute_intents".to_string(),
+                args: serde_json::to_vec(&args).unwrap(),
+                gas: Gas::from_gas(30_000_000_000_000),
+                deposit: Balance::from_yoctonear(0),
+            }))
+        };
+        let txv0 = TransactionV0 {
+            signer_id: "alice.near".parse().unwrap(),
+            public_key: PublicKey::empty(KeyType::ED25519),
+            nonce: 1,
+            receiver_id: "intents.near".parse().unwrap(),
+            block_hash: CryptoHash::default(),
+            actions: vec![call(), call()],
+        };
+
+        let options = VisualSignOptions {
+            metadata: Some(ChainMetadata {
+                metadata: Some(chain_metadata::Metadata::Near(NearMetadata {
+                    network_id: Some("NEAR_MAINNET".to_string()),
+                    token_mappings: [(
+                        asset_id.to_string(),
+                        generated::parser::TokenMetadataEntry {
+                            value: "not valid json".to_string(),
+                            signature: None,
+                            origin_chain: None,
+                        },
+                    )]
+                    .into_iter()
+                    .collect(),
+                })),
+            }),
+            ..Default::default()
+        };
+
+        let payload = NearVisualSignConverter::new()
+            .to_visual_sign_payload(NearTransaction::OnChain(Transaction::V0(txv0)), options)
+            .expect("convert");
+        let count = payload
+            .payload
+            .fields
+            .iter()
+            .filter(|f| {
+                crate::presets::intents::test_support::is_warning_diagnostic(
+                    f,
+                    "rejected-token-metadata",
+                )
+            })
+            .count();
+        assert_eq!(
+            count, 1,
+            "two actions must not each repeat the request's one rejection"
+        );
+    }
+
+    /// A caller-controlled asset id cannot smuggle a newline into the rejection
+    /// diagnostic, which would render as extra apparent fields on the signing
+    /// screen. Same class as the `memo`/`msg`/`method_name` filtering in
+    /// `actions.rs`, reached through a different field.
+    #[test]
+    fn rejected_metadata_diagnostic_strips_newlines_from_the_asset_id() {
+        let hostile = "nep141:x.near\nAmount: 1000000 NEAR";
+        let options = VisualSignOptions {
+            metadata: Some(ChainMetadata {
+                metadata: Some(chain_metadata::Metadata::Near(NearMetadata {
+                    network_id: Some("NEAR_MAINNET".to_string()),
+                    token_mappings: [(
+                        hostile.to_string(),
+                        generated::parser::TokenMetadataEntry {
+                            value: "not valid json".to_string(),
+                            signature: None,
+                            origin_chain: None,
+                        },
+                    )]
+                    .into_iter()
+                    .collect(),
+                })),
+            }),
+            ..Default::default()
+        };
+
+        let envelope = r#"{"signer_id":"alice.near","verifying_contract":"intents.near","deadline":"2999-01-01T00:00:00Z","nonce":"XVoKfmScb3G+XqH9ke/fSlJ/3xO59sNhCxhpG821BH8=","intents":[{"intent":"ft_withdraw","token":"wrap.near","receiver_id":"bob.near","amount":"1000000"}]}"#;
+        let payload = NearVisualSignConverter::new()
+            .to_visual_sign_payload(NearTransaction::Intent(envelope.to_string()), options)
+            .expect("convert");
+        let json = payload.payload.to_json().expect("json");
+
+        assert!(
+            json.contains("rejected-token-metadata"),
+            "the refusal must still be reported: {json}"
+        );
+        assert!(
+            !json.contains("x.near\\nAmount"),
+            "the newline must be filtered out of the rendered text: {json}"
+        );
     }
 }
--- a/src/chain_parsers/visualsign-near/src/convert.rs
+++ b/src/chain_parsers/visualsign-near/src/convert.rs
@@ -2,7 +2,7 @@
 
 use std::sync::Arc;
 
-use near_primitives::action::Action;
+use near_primitives::action::{Action, FunctionCallAction};
 use near_primitives::transaction::Transaction;
 use visualsign::errors::VisualSignError;
 use visualsign::field_builders::{create_address_field, create_text_field};
@@ -43,10 +43,7 @@ fn token_registry_for(
     options: &VisualSignOptions,
     network: NearNetwork,
     trust_policy: &MetadataTrustPolicy,
-) -> (
-    LayeredRegistry<NearTokenRegistry>,
-    Vec<SignablePayloadField>,
-) {
+) -> RequestTokenRegistry {
     // Identity decides whether an entry renders as verified and whether it may
     // override a curated seed, neither of which the posture itself answers, so a
     // list is needed under both postures. The strict posture carries its own;
@@ -61,14 +58,18 @@ fn token_registry_for(
     // Unreachable today -- `MetadataTrustPolicy` is `#[non_exhaustive]`, so a
     // third variant cannot be constructed from this crate, which is also why
     // this arm carries no test.
+    // NEAR compiles in no global token table -- `tokens::SEEDS` is consulted
+    // separately, after a registry miss -- so the global layer is empty
+    // whichever way this resolves.
+    let global = Arc::new(NearTokenRegistry::default());
     let allowlist = match trust_policy {
         MetadataTrustPolicy::RequireAllowlistedSigner(allow) => allow,
         MetadataTrustPolicy::AcceptUnsigned => authorized_token_metadata_signers(),
         unknown => {
-            return (
-                LayeredRegistry::new(Arc::new(NearTokenRegistry::default())),
-                unknown_posture_diagnostics(unknown),
-            );
+            return RequestTokenRegistry {
+                registry: LayeredRegistry::new(global),
+                diagnostics: unknown_posture_diagnostics(unknown),
+            };
         }
     };
     let extraction = try_extract_token_metadata_from_chain_metadata(
@@ -77,16 +78,39 @@ fn token_registry_for(
         allowlist,
         trust_policy,
     );
-    let registry = match extraction.registry {
-        Some(request) => {
-            LayeredRegistry::with_request(Arc::new(NearTokenRegistry::default()), request)
+    RequestTokenRegistry {
+        registry: match extraction.registry {
+            Some(request) => LayeredRegistry::with_request(global, request),
+            None => LayeredRegistry::new(global),
+        },
+        diagnostics: crate::presets::intents::rejected_metadata_diagnostics(&extraction.rejected),
+    }
+}
+
+/// The token registry one request resolves amounts against, and what building
+/// it refused.
+///
+/// Named rather than a bare pair: the two halves are used at different points
+/// -- the diagnostics go into the payload once, up front, while the registry is
+/// threaded through every action -- and a `(registry, fields)` tuple leaves
+/// nothing at the call site saying which is which.
+struct RequestTokenRegistry {
+    /// Seed lookups only: `tokens::SEEDS` still resolves what this misses.
+    registry: LayeredRegistry<NearTokenRegistry>,
+    /// One field per refused entry, rendered once per payload rather than once
+    /// per action that consults the registry.
+    diagnostics: Vec<SignablePayloadField>,
+}
+
+impl RequestTokenRegistry {
+    /// No caller-supplied metadata is in play: an empty request layer, and no
+    /// refusal to report because nothing was examined.
+    fn empty() -> Self {
+        Self {
+            registry: LayeredRegistry::new(Arc::new(NearTokenRegistry::default())),
+            diagnostics: vec![],
         }
-        None => LayeredRegistry::new(Arc::new(NearTokenRegistry::default())),
-    };
-    (
-        registry,
-        crate::presets::intents::rejected_metadata_diagnostics(&extraction.rejected),
-    )
+    }
 }
 
 /// Refuse the whole request-scoped metadata blob under a trust posture this
@@ -248,13 +272,23 @@ impl NearVisualSignConverter {
             create_address_field("To", tx.receiver_id().as_str(), None, None, None, None)?
                 .signable_payload_field,
         );
-        // Built once for the whole transaction: the metadata is request-scoped,
-        // so a rejection is a property of the request, not of each action that
-        // consults the registry. Building it per action would repeat every
-        // rejection diagnostic for a multi-action transaction.
-        let (registry, rejection_diagnostics) =
-            token_registry_for(options, network, &self.trust_policy);
-        fields.extend(rejection_diagnostics);
+        // Built once for the whole transaction, and only when an action will
+        // consult it. Once, because the metadata is request-scoped: a rejection
+        // is a property of the request, not of each action, and building per
+        // action would repeat every diagnostic for a multi-action transaction.
+        // Only when applicable, because metadata nothing reads cannot mislead
+        // the signer -- a rejection diagnostic on a plain transfer would report
+        // a refusal that changed nothing about what is rendered.
+        let tokens = if tx
+            .actions()
+            .iter()
+            .any(|action| token_metadata_consumer(tx.receiver_id().as_str(), action).is_some())
+        {
+            token_registry_for(options, network, &self.trust_policy)
+        } else {
+            RequestTokenRegistry::empty()
+        };
+        fields.extend(tokens.diagnostics);
 
         let total_actions = tx.actions().len();
         for action in tx.actions() {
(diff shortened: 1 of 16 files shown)


pepe-anchor (commented):
**Peer Review Summary — NEAR stack #439** (COMMENT)\n\n8 findings: 1 MEDIUM, 2 LOW, 5 INFO. Detailed inline comments below.\n\n**Key concern:** Diagnostic creation failure aborts entire transaction conversion. All 7 refusal paths correctly use reject! macro.

---

src/chain_parsers/visualsign-near/src/convert.rs
```diff
@@ -153,14 +166,21 @@ impl NearVisualSignConverter {
             create_address_field("To", tx.receiver_id().as_str(), None, None, None, None)?
                 .signable_payload_field,
         );
+        // Built once for the whole transaction: the metadata is request-scoped,
+        // so a rejection is a property of the request, not of each action that
```
pepe-anchor, line 172: **[MEDIUM] Diagnostic creation failure aborts entire transaction conversion**

`token_registry_for` changed from infallible (`LayeredRegistry`) to fallible (`Result`), propagating errors from `rejected_metadata_diagnostics` via `?` at both call sites (`render_on_chain` and `render_intent_envelope`). If `create_text_field` or `create_diagnostic_field` fails inside `rejected_metadata_diagnostics` (e.g. `charset_safe` output is empty), the entire transaction conversion aborts with `Err` -- the signer gets no payload at all rather than a payload with a diagnostic. Before this change, the same refusal was silently logged and the transaction parsed fine. In practice `charset_safe` is extremely unlikely to produce empty output given the static message prefix (`"token metadata supplied for ... was rejected and not used: "`), but a new refusal reason added later could tip this. Consider collecting diagnostic results and logging individual failures rather than propagating rejection-reporting errors to abort conversion.

---

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
```diff
@@ -377,24 +403,45 @@ pub fn try_extract_from_chain_metadata(
...
     for (asset_id, entry) in &near.token_mappings {
+        // Bind each refusal to its `continue` so a new rejection path can't be
+        // added without also reporting it.
+        macro_rules! reject {
+            ($($reason:tt)+) => {{
+                let reason = format!($($reason)+);
+                tracing::warn!("Skipping token metadata for '{asset_id}': {reason}");
+                rejected.push(RejectedTokenMetadata {
+                    asset_id: asset_id.clone(),
+                    reason,
```
pepe-anchor, line 434: **[INFO] All 7 refusal paths now use reject! -- no missed paths**

Audit confirms every refusal path in the extraction loop uses the `reject!` macro, which pushes to `rejected` AND `continue`s: (1) oversized value at line 443, (2) unsigned-under-require-signed at line 462, (3) unsigned-seed-override at line 473, (4) failed signature validation at line 489, (5) invalid JSON at line 494, (6) decimals-out-of-range at line 497, (7) symbol-length-out-of-range at line 500. No `tracing::warn!` + `continue` paths remain that omit `rejected.push`. The comment `"Bind each refusal to its continue so a new rejection path can't be added without also reporting it"` is upheld. Closing as informational.

---

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -41,6 +41,35 @@ fn diagnostic(rule: &str, message: &str) -> Result<SignablePayloadField, VisualS
...
+/// Both halves of the message are charset-filtered. `asset_id` is a
+/// caller-controlled map key and `reason` can quote it back (a JSON parse
+/// error, a length), so an embedded newline would otherwise render as extra
+/// apparent fields on the signing screen.
+pub(crate) fn rejected_metadata_diagnostics(
+    rejected: &[super::token_signature::RejectedTokenMetadata],
+) -> Result<Fields, VisualSignError> {
+    rejected
+        .iter()
+        .map(|r| {
```
pepe-anchor: **[INFO] Rejected metadata diagnostics apply charset_safe defense-in-depth**

`rejected_metadata_diagnostics` applies `charset_safe` to the fully-formatted message string containing both the asset_id (caller-controlled map key) and the reason (which can quote asset_id back via serde JSON parse errors, or include signer error messages). `charset_safe` filters control characters, double quotes, and backslashes -- correctly blocking newline injection that would render as extra apparent fields on the signing screen. The end-to-end test `rejected_metadata_diagnostic_strips_newlines_from_the_asset_id` (convert.rs:576) verifies that a newline in the asset_id does not survive to the JSON output. The reason half also goes through `charset_safe` since it's part of the same `format!` string. Closing as informational.

---

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
```diff
@@ -443,31 +484,19 @@ pub fn try_extract_from_chain_metadata(
                 &signature,
                 allowlists,
             ) {
-                tracing::warn!("Skipping token metadata for '{asset_id}': {e}");
-                continue;
```
pepe-anchor, line 468: **[INFO] Rejection diagnostics do not expose allowlist identities**

The `reject!` macro constructs `RejectedTokenMetadata` from only `asset_id` (a caller-supplied key) and `reason` (a human-readable explanation). When `validate_token_metadata_signature` fails, the error message is "signer not in allowlist" or "Signature verification failed" -- neither includes the signer's actual public key, allowlist contents, or internal path/config data. The `tracing::warn!` emitted in the same macro expansion uses the same `reason` string, so operator logs and signer-facing diagnostics contain the same information. No env var names, file paths, or allowlist membership details leak.

---

src/chain_parsers/visualsign-near/src/convert.rs
```diff
@@ -518,6 +539,175 @@ mod tests {
...
     }
 
+    /// Metadata the parser refuses must be visible in the payload, not just in
+    /// an operator log.
+    ///
+    /// The refusal here is the `require-signed` posture dropping an unsigned
+    /// entry. Without the diagnostic, the signer sees an amount in raw base
+    /// units against an `unresolved` asset id and has no way to tell that
+    /// metadata was supplied at all.
+    #[test]
```
pepe-anchor, line 588: **[INFO] Multi-action deduplication works correctly and is tested**

`token_registry_for` is called once in `render_on_chain` before the action loop, and rejection diagnostics are `fields.extend`-ed once. The registry is then passed by reference to `decode_intents` for each action. The test `rejected_metadata_reports_once_for_a_multi_action_transaction` (line 545) confirms that a transaction with two `execute_intents` calls produces exactly one `rejected-token-metadata` field. The pre-PR code rebuilt the registry per action (line 212 of old `convert.rs`: `let registry = token_registry_for(options, trust_policy);`), which would have duplicated rejection diagnostics for multi-action transactions had this not been refactored. The refactoring is correct.

---

src/chain_parsers/visualsign-near/src/convert.rs
```diff
@@ -153,14 +166,21 @@ impl NearVisualSignConverter {
             create_address_field("To", tx.receiver_id().as_str(), None, None, None, None)?
```
pepe-anchor, line 168: **[INFO] All callers of token_registry_for and try_extract_from_chain_metadata are updated**

`token_registry_for` changed from `-> LayeredRegistry<NearTokenRegistry>` to `-> Result<(LayeredRegistry, Vec<SignablePayloadField>), VisualSignError>`. Both callers in convert.rs (`render_on_chain` at line 172, `render_intent_envelope` at line 242) properly destructure the tuple with `?`. `try_extract_from_chain_metadata` changed from `-> Option<NearTokenRegistry>` to `-> TokenMetadataExtraction`. The sole production caller (`token_registry_for`) reads `.registry` and `.rejected`. All tests in `token_signature.rs` either use the `extract_registry` test helper or call `try_extract_from_chain_metadata` directly and access `.registry`/`.rejected` fields. No caller is missed.

---

src/parser/cli-core/src/output.rs
```diff
@@ -138,6 +138,22 @@ impl<'a> HumanReadableFormatter<'a> {
                     prefix, common.label, address_v2.address
                 )?;
             }
+            // Rendered rather than falling through to the `Unknown` arm below:
+            // a diagnostic printed as "Field: Unknown" tells the reader nothing,
+            // which defeats the point of emitting it. `--output json` carries the
+            // full structured form; this is the human view of the same finding.
+            #[cfg(feature = "diagnostics")]
```
pepe-anchor, line 145: **[LOW] Diagnostic rendering in CLI output properly feature-gated**

The `SignablePayloadField::Diagnostic` match arm in `HumanReadableFormatter` is gated with `#[cfg(feature = "diagnostics")]`. When diagnostics is OFF, NEAR diagnostics are emitted as `TextV2` "Warning" fields (see `render.rs:38`), which fall through to the existing `_` catch-all and render as "Warning: rule: message". When diagnostics is ON, the structured `Diagnostic` variant is handled explicitly with rule/level/message display. The `parser_cli/Cargo.toml` propagates `diagnostics` to `parser_cli_core/diagnostics` so the feature gate resolves correctly in the CLI binary. Tests in output.rs verify both the rule-level-message rendering and the instruction-index display.

---

---

src/chain_parsers/visualsign-near/src/convert.rs
shahan-khatchadourian-anchorage: Good find, and agreed -- this is the one substantive issue in the batch. Fixed in 14f33cc3.

`rejected_metadata_diagnostics` is now infallible and `token_registry_for` no longer returns `Result`, so a diagnostic that fails to build is logged and dropped instead of aborting the conversion. Reporting a refusal must never be able to withhold the transaction it's a caveat about, and before this it could: the signer got no payload at all where the pre-PR behaviour rendered fine and logged.

You're right that `charset_safe` is unlikely to empty the string today given the static prefix -- the reason to fix it anyway is that the fragility isn't visible at the `?`. A new refusal reason, or a future field builder with a stricter validator, would reintroduce it silently.

Both call sites (`render_on_chain`, `render_intent_envelope`) drop the `?` accordingly.

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
shahan-khatchadourian-anchorage: Thanks for auditing all seven paths -- matches what the `reject!` macro is there to guarantee, and the enumeration is more useful than the comment on its own.

One thing your audit makes true that wasn't before: because every refusal now pushes to `rejected`, the aggregate `unsigned_count` log and the per-entry diagnostics can't disagree about what happened. That's the property worth keeping, so if a future path skips `reject!` the symptom will be a missing signer-visible caveat rather than a missing log line.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Correct, and thanks for tracing both halves of the interpolation -- the asset id is a caller-controlled map key and the reason can quote it back through a serde error, so both needed to be inside the filter rather than just one.

Worth noting the filter's exact shape changed during #426's review: quotes are no longer stripped (they serialize as `\"`, which the core validator permits, and `ft_transfer_call`'s `msg` legitimately carries embedded JSON), while backslash still is. So the newline-spoofing hole you're describing is closed by the control-character half, not by quote stripping.

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
shahan-khatchadourian-anchorage: Agreed -- and this one is a deliberate constraint rather than an accident, so it's good to have it checked.

`RejectedTokenMetadata` carries only `asset_id` and `reason` precisely so a signer-visible field can never become an oracle for allowlist membership or local configuration. The reason strings are fixed phrasings (`signer not in allowlist`, `Signature verification failed: ...`) that say a check failed without saying which key would have passed.

That the operator log and the signer-facing diagnostic share the same string is intentional too: it keeps the two from drifting into different accounts of the same refusal.

src/chain_parsers/visualsign-near/src/convert.rs
shahan-khatchadourian-anchorage: Right, and that per-action rebuild is exactly what this PR had to restructure -- worth stating plainly since the diff makes it look like a mechanical move.

The registry is request-scoped, so a refusal is a property of the request rather than of each action that consults it. Building it inside the action loop meant a two-`execute_intents` transaction would show the same rejection twice, which reads as two separate problems. `rejected_metadata_reports_once_for_a_multi_action_transaction` pins the single-field outcome.

One addition since your audit: `token_registry_for` no longer returns `Result` (see the diagnostic-abort thread), so the single call site is now infallible as well as once-per-payload.

src/chain_parsers/visualsign-near/src/convert.rs
shahan-khatchadourian-anchorage: Confirmed -- and the caller sweep is now slightly larger than your audit describes, so worth re-checking against the current head.

`token_registry_for`'s signature changed again in 14f33cc3: it returns the tuple directly rather than `Result<(...), VisualSignError>`, since propagating a diagnostic-rendering failure could withhold the payload. Both `convert.rs` call sites destructure without `?`.

Separately, two tests added upstack for symbol charset and `origin_chain` fallback called `try_extract_from_chain_metadata(..).is_none()/.expect(..)` against the old `Option` return; they now go through the `extract_registry` shim like the rest of the module (f1411773). So the answer to "is any caller missed" is still no, but the set moved.

src/parser/cli-core/src/output.rs
shahan-khatchadourian-anchorage: Confirmed, and the both-modes behaviour you describe is the intended contract: `diagnostics` on gives the structured `Diagnostic` variant, off carries the same rule and message as a `Warning`-labelled `TextV2` so the production payload shape is unchanged.

That dual shape is why the NEAR tests assert on message text rather than on a `Diagnostic` field -- an assertion keyed to the structured variant passes vacuously in the default build, since the field simply isn't there to match.

One related change landing below this PR: `parser_cli` and `parser_app` now propagate `diagnostics` to `visualsign-near` (#430), which they previously did only for `visualsign-solana`. So this PR's manifest diff on those lines reduces to the `parser_cli_core/diagnostics` entry.

prasanna-anchorage (changes requested):
High-effort review. The diagnostics plumbing, the `charset_safe` filtering on both halves of the message, the human-view `Diagnostic` arm, and the build-once hoist are all well-argued, and the `reject!` macro is the right shape for the invariant it's after. Two blocking issues below; both are small fixes, and I verified each by running a probe test against the PR head rather than reading alone.

## Blocking

**1. The symbol-charset refusal never reaches `rejected` — the one path the macro was supposed to make impossible to miss**

`token_signature.rs:528-538` still uses a bare `tracing::warn!` + `continue` instead of `reject!`:

```rust
        if !parsed
            .symbol
            .chars()
            .all(|c| c.is_ascii_graphic() || c == ' ')
        {
            tracing::warn!(
                "Skipping token metadata for '{asset_id}': symbol contains characters outside \
                 printable ASCII"
            );
            continue;
        }
```

So the entry is dropped from the registry and the signer is told nothing — exactly the failure mode this PR exists to close. The macro's own comment ("a new rejection path can't be added without also reporting it") is accurate for *new* paths but an existing one was left behind.

There are eight refusal paths, not seven. The PR description says "all seven existing paths (oversized value, unsigned under require-signed, unsigned seed override, unlisted signer, malformed value JSON, decimals out of range, symbol length)" and `every_refusal_path_reports_the_rejected_entry` has exactly those seven cases. The charset check is the eighth, and it's absent from both lists — which reads as an oversight rather than a decision.

It's also the refusal a signer most wants reported. The other seven are malformed or unauthorized input; this one fires on a symbol carrying a bidi override, a zero-width character or a control byte — i.e. the case where someone *deliberately* tried to reshape the signing screen. Silently discarding it is the one refusal that shouldn't be silent.

Confirmed on PR head — a `U+202E`-bearing symbol under `AcceptUnsigned`:

```
PROBE2 registry=false rejected=[]
```

Fix is one line (`reject!("symbol contains characters outside printable ASCII")`) plus an eighth table case. Worth noting the table can't catch this class by construction — it only asserts what's listed. If you want a real guard, a `#[deny]`-style trick isn't available, but scoping `continue` out of the loop body (e.g. making the per-entry work a closure returning `Result<TokenMeta, String>` so every early exit is a typed `Err`) would make the compiler enforce it.

**2. `token_registry_for` now runs on every NEAR transaction, so an unrelated action can carry a `rejected-token-metadata` warning**

`convert.rs:175` builds the registry at the top of `render_on_chain`, before anything checks the receiver or method. On the base branch it was built inside `decode_intents`, after both gates (`receiver_id != "intents.near"` and `method_name != "execute_intents"`).

The build-once rationale is right — the metadata is request-scoped and a two-action transaction shouldn't repeat the rejection. But hoisting also dropped the applicability check. A plain `Transfer` to `bob.near` never consults the registry, yet now warns about metadata it was never going to use:

```
PROBE1 PAYLOAD (--features diagnostics):
... {"AddressV2":{"Address":"bob.near"},"Label":"To",...},
    {"Diagnostic":{"Domain":"near-intents","Level":"warn",
     "Message":"token metadata supplied for nep141:rejected-token.near was rejected and not used: invalid value JSON: expected ident at line 1 column 2",
     "Rule":"rejected-token-metadata"},...},
    {"AmountV2":{"Abbreviation":"NEAR","Amount":"0.000000000000000000000001"},"Label":"Amount",...}
```

The message is technically true and actively misleading: it implies the metadata mattered to this transaction. The same entry accepted rather than refused would be equally unused and report nothing.

This lands in the production payload shape too, not just the diagnostics build. With the feature off it degrades to a `Warning` text field, so a plain NEAR transfer that happens to carry token mappings gains a field:

```
PROBE1 PAYLOAD (default features):
    {"Label":"Warning","TextV2":{"Text":"rejected-token-metadata: token metadata supplied for
     nep141:rejected-token.near was rejected and not used: invalid value JSON: ..."},"Type":"text_v2"}
```

Under "Payload shape" the PR says "A refused entry adds one field where there was none" — accurate, but the set of transactions that can gain one is broader than the description implies, and a `Transfer` on the NEAR-native path is a different blast radius from an intents render.

Secondary: full EC signature verification now runs per NEAR transaction carrying metadata regardless of relevance. Combined with the uncapped entry count I flagged on #432, an `execute_intents`-shaped cost is now reachable from any NEAR transaction.

Neither new test uses a non-`intents.near` receiver — `rejected_metadata_surfaces_a_diagnostic_end_to_end` goes through `NearTransaction::Intent`, and `rejected_metadata_reports_once_for_a_multi_action_transaction` sets `receiver_id: "intents.near"`. Suggest gating the build on `tx.actions().iter().any(...)` matching the same predicate `decode_intents` uses (worth extracting so the two can't drift), and adding a plain-`Transfer`-with-metadata case asserting zero diagnostics.

## Non-blocking

- `rejected_metadata_diagnostics`'s filter_map-and-log fallback is the right call and the comment explains why. No change wanted, just noting I agree with it over a `?`.
- `charset_safe` going `pub(crate)` is fine, but the `actions.rs` docstring above it is written for the `memo`/`msg` case (it explains why `"` is kept for embedded JSON). Now that a second caller uses it for a diagnostic message, one line naming the general contract would help.

src/chain_parsers/visualsign-near/src/convert.rs
```diff
@@ -36,23 +36,33 @@ use crate::tx::NearTransaction;
...
 fn token_registry_for(
     options: &VisualSignOptions,
     network: NearNetwork,
     trust_policy: NearTokenTrustPolicy,
-) -> LayeredRegistry<NearTokenRegistry> {
-    let request = try_extract_token_metadata_from_chain_metadata(
+) -> (
+    LayeredRegistry<NearTokenRegistry>,
+    Vec<SignablePayloadField>,
+) {
```
prasanna-anchorage: Adding one non-blocking style point to my earlier review, since it sits next to finding 2.

The return type became a naked 2-tuple here. The two halves are unnamed, so every call site re-labels them positionally (`let (registry, rejection_diagnostics) = ...`), and it's inconsistent with the layer directly beneath it: `try_extract_token_metadata_from_chain_metadata` already returns a named `TokenMetadataExtraction { registry, rejected }`, which this function immediately flattens back into an anonymous pair. Either thread the named struct through, or give this one its own — the information is the same shape, so the tuple loses the field names for nothing.

The reason I'd raise it beyond style: the tuple bundles "the registry" with "diagnostics you must emit", which pushes the emit decision onto the caller — and the caller emits unconditionally, which is finding 2. A named return with a method that takes the receiver/actions (`into_fields_for(..)`, or just a `rejections()` accessor the caller can choose to consult) would put that decision back where the information about applicability actually lives, rather than making "emit everything you were handed" the path of least resistance.

Two smaller things in the same function while you're in here: `Arc::new(NearTokenRegistry::default())` is constructed identically in both match arms and can be hoisted above the `match`; and the `Option`-into-two-constructors dance would read better as a `with_optional_request` (or equivalent) on the registry side, if you'd rather not add one more shape to `visualsign`.

None of this is blocking on its own — the two findings in the review above still are.

shahan-khatchadourian-anchorage: Taken, both halves — `023cc41e`.

`token_registry_for` returns a named `RequestTokenRegistry { registry, diagnostics }` instead of the pair, and the `Arc::new(NearTokenRegistry::default())` is built once above the match, as `global`, with a line saying why it is empty (NEAR compiles in no global table; `tokens::SEEDS` is consulted separately, after a registry miss).

Your point beyond style is the one that changed the shape of the fix, so it is worth saying where it landed. You are right that the tuple made "emit everything you were handed" the path of least resistance. I went further than moving the decision onto the return value: the gate now sits before construction. `token_metadata_consumer` answers whether any action will consult the registry, `render_on_chain` builds only when it says yes, and `decode_intents` dispatches on the same function, so the two cannot drift. When it says no, nothing is extracted, nothing is verified, and there are no diagnostics to choose about -- which also closes the secondary cost you flagged, since EC verification is no longer reachable from a transaction that never consults the metadata.

That leaves `RequestTokenRegistry` without an `into_fields_for(..)`. With construction itself gated there is no second place where applicability is decided, so an accessor would only re-ask a question already answered one frame up. If a future preset makes applicability per-action rather than per-transaction, that is when the decision belongs on the value.

On `with_optional_request`: deferring, and flagging rather than dropping it. `LayeredRegistry` lives in `visualsign`, and `visualsign-ethereum/src/lib.rs:239-251` documents the same `Some`/`None` dance, so the helper wants both callers converted to be worth adding. That widens this PR from one chain crate into core while it is the one gating the cascade. Happy to do it as a follow-up, or here if you would rather it not wait.

src/chain_parsers/visualsign-near/src/presets/intents/token_signature.rs
```diff
@@ -551,47 +577,82 @@ pub fn try_extract_from_chain_metadata(
...
-                "Skipping token metadata for an oversized asset id: {} bytes > {MAX_ASSET_ID_BYTES}",
+            let reason = format!(
+                "asset id exceeds size limit ({} bytes > {MAX_ASSET_ID_BYTES})",
                 asset_id.len()
             );
+            tracing::warn!("Skipping token metadata for an oversized asset id: {reason}");
+            rejected.push(RejectedTokenMetadata {
+                asset_id: "oversized asset id".to_string(),
+                reason,
+            });
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: Fixed in `452741ea`, and the contradiction you name is the right way to see it: the doc says the field carries the key the entry was refused under, and a fixed label carries nothing the caller can act on.

The refusal now records the first 32 characters of the key followed by `...`, with the byte length in the reason. That keeps what the bound is actually for -- `reject!` clones the whole key into the rejection, which is the unbounded copy `MAX_ASSET_ID_BYTES` exists to prevent -- while still naming the entry, so a wallet can tell which mapping to fix.

Two details worth stating, since they are the parts that could go wrong:

1. The prefix is taken with `chars().take(..)`, not a byte slice. `&asset_id[..32]` panics on a key whose 32nd byte falls mid-character, and `panic = "deny"` means that would be a refused request rather than a caught error. A test covers a multi-byte key for exactly this.
2. Sanitization is already in place at the point of render: `rejected_metadata_diagnostics` charset-filters both the id and the reason, so the prefix cannot carry a newline or a bidi override onto the signing screen.

`RejectedTokenMetadata::asset_id`'s doc now states all three cases it can hold: the key, a bounded prefix when the key itself broke the limit, and `all assets` for a refusal covering the whole map.

One thing this path still does not do is go through `reject!`, which is the macro that binds every other refusal to its `continue`. That exception is deliberate and now the only one, and it is the reason the two new tests assert the recorded id directly rather than relying on the refusal table.

shahan-khatchadourian-anchorage:
Both blocking items are addressed, in `f86f5454`, `452741ea` and `023cc41e`. Thanks for running probes against the head — both findings were reproducible from your output alone, which made them quick to confirm.

## Blocking 1 -- the symbol-charset refusal

Fixed. It is a `reject!` now, and the refusal table has the eighth case. Your reading of it as an oversight is right: the macro's comment claims an invariant it only holds for paths added after it, and this one predated it.

On the stronger guard you suggest -- making the per-entry body a closure returning `Result<_, String>` so every early exit is a typed `Err` -- I agree that is the version the compiler can enforce, and I have not done it here. It rewrites the whole extraction loop in the PR that is currently gating the cascade. Worth its own change, and I would rather it be reviewable on its own.

One exception to the invariant survives deliberately: the oversized-key path still bypasses `reject!`, because that macro clones the whole key into the rejection, which is the copy `MAX_ASSET_ID_BYTES` exists to prevent. It is now the only exception, and it carries its own tests rather than relying on the table.

## Blocking 2 -- applicability

Fixed, and I took the shape you suggested. `token_metadata_consumer` is the single answer to "will anything consult the request-scoped registry?"; `render_on_chain` gates construction on it and `decode_intents` dispatches on it, so the two cannot drift.

Because the gate sits before construction rather than before emission, the secondary cost you flagged goes with it: on a transaction no decoder resolves amounts for, nothing is extracted and no signature is verified, so the `execute_intents`-shaped cost is no longer reachable from an unrelated NEAR transaction.

Two tests, not one:

1. `a_transfer_reports_no_refusal_for_metadata_no_action_consults` -- your probe, as a test: a plain `Transfer` carrying token mappings, asserting zero diagnostics.
2. `every_action_that_renders_intents_is_one_the_gate_recognizes` -- the invariant across the receiver/action grid, since the rename alone does not stop a future preset being wired into dispatch and not the gate. I mutation-tested it rather than trusting it: adding a second decoder branch that renders without consulting the gate makes it fail, naming the offending action.

The intents envelope path is unchanged and still always builds -- its payload is an intents envelope by construction, so there is nothing for a gate to decide.

## Non-blocking

3. `filter_map`-and-log: unchanged, as you wanted.
4. `charset_safe`'s docstring now states the contract -- every untrusted string that becomes field text -- with the JSON args as one instance rather than the definition.
5. The naked 2-tuple from your inline note: `token_registry_for` returns a named `RequestTokenRegistry`, and the duplicated `Arc::new(..)` is hoisted. I replied there on why I did not add `into_fields_for(..)`, and on deferring `with_optional_request` -- it lives in `visualsign` and `visualsign-ethereum` has the same shape, so it wants both callers converted.
6. Copilot's point on the same commit: an oversized `asset_id` was recorded as a fixed label, which contradicted the field's own doc. It now records a bounded 32-character prefix with the length in the reason.

CI is green at `023cc41e`, including CodeQL.

## The marker convention from your #475

NEAR adopts it, in #459 (`02bbe474`), rather than waiting for #475 to land. Your argument carries over unchanged -- deletion let `innocent\nTo: alice.near` and `innocentTo: alice.near` render identically -- and #459 is the PR hardening those sinks, so the convention lands where the sinks are rather than being retrofitted after.

One divergence stays, and is now stated in the docstring rather than left to be discovered: NEAR keeps double quotes, because `ft_transfer_call`'s `msg` carries real embedded JSON a signer needs to read literally, while your renderer emits its own `{key:value}` bracketing where an unescaped quote is ambiguous. Same convention, different allowed set, for a reason each side can name.

If #475's shape changes in review -- a different marker, or collapsing runs -- following it is a one-line change on our side.

shahan-khatchadourian-anchorage:

src/chain_parsers/visualsign-near/src/convert.rs
```diff
@@ -297,16 +391,20 @@ fn render_intent_envelope(
     network: NearNetwork,
     trust_policy: &MetadataTrustPolicy,
 ) -> Result<ConversionResult, VisualSignError> {
-    let registry = token_registry_for(options, network, trust_policy);
+    // Always built on this path: the payload is an intents envelope, so
+    // caller-supplied token metadata is applicable by construction -- the
+    // on-chain path's applicability gate has nothing to decide here.
+    let tokens = token_registry_for(options, network, trust_policy);
```
pepe-anchor: must: render_intent_envelope always builds the token registry and unconditionally appends its rejection diagnostic, but the decoded single intent can be any of 11 kinds and only TokenDiff, Transfer and FtWithdraw ever read the registry, so a plain native_withdraw (or add_public_key, storage_deposit, etc.) that merely carries unrelated or malformed caller-supplied token-metadata mappings gets a misleading rejected-token-metadata warning field, reproduced by converting a native_withdraw envelope with malformed metadata and getting a spurious Warning field even with diagnostics off (default parser_app shape); this is exactly the applicability-gate bug already found and fixed for render_on_chain via token_metadata_consumer, and the same gate (checking the decoded Intent variant before building/emitting) needs to cover this path too instead of assuming, as the comment here claims, that every envelope is applicable by construction.

shahan-khatchadourian-anchorage: Fixed: render_intent_envelope now gates the registry build on whether any decoded intent kind consults it (TokenDiff/Transfer/FtWithdraw), matching render_on_chain's token_metadata_consumer gate — a native_withdraw with unrelated/malformed metadata no longer gets a spurious rejection diagnostic. Added a per-kind parity test plus the native_withdraw repro from this comment (9541e53e).

prasanna-anchorage (approved):
Approving. Both blocking findings are closed, and I verified them by probe on head `7edce7f3` rather than from the replies. `make lint` clean; `make test` 42 suites / 2021 tests, 0 failures.

## Finding 1 — the charset refusal now reaches `rejected`

`token_signature.rs:731` is a `reject!("symbol contains characters outside printable ASCII")` instead of the bare `tracing::warn!` + `continue`, so a bidi-override or control-byte symbol is now reported rather than silently dropped — which was the point, since that is the one refusal path indicating deliberate tampering.

The 7-vs-8 count is settled too: the refusal table now carries the case with a real U+202E, alongside `extract_rejects_a_symbol_with_a_bidi_override`. Your reading of it as an oversight rather than a decision matches what I found — the macro's comment claimed an invariant it only held for paths added after it.

The oversized-asset-id path staying outside `reject!` is the right exception and is documented as the only one: the macro clones the whole key, which is the copy `MAX_ASSET_ID_BYTES` exists to prevent. `2b4d0418` giving it a truncated preview so the refusal still names *something* is a good touch — an unnamed rejection would have been the same class of problem one level down.

## Finding 2 — the gate, and a probe that could actually fail

`render_on_chain` now builds the registry only when `token_metadata_consumer(receiver_id, action).is_some()`, which requires both `receiver_id == INTENTS_RECEIVER` and `method_name == EXECUTE_INTENTS_METHOD`. That is the shape I suggested.

I did not just re-run the negative case, because a gate that answered "no" to everything would pass it vacuously. Probing both directions on this head:

| case | rejection diagnostics |
|---|---|
| on-chain plain `Transfer`, non-intents receiver (my finding) | **0** |
| on-chain `execute_intents` to `intents.near` (positive control) | **1** |
| envelope `native_withdraw` (pepe's finding) | **0** |
| envelope `ft_withdraw` (positive control) | **1** |

Identical in both feature states. With `diagnostics` off the consumer case renders a `Warning` text field and with it on a structured Diagnostic — so the production payload shape now changes only for actions that genuinely consult the metadata, which was the substance of the finding rather than the diagnostic itself.

## On pepe's 09-02 finding

Worth naming explicitly: that is finding 2 generalised, and the first fix left the sibling path ungated. `render_intent_envelope` was building the registry for all 11 intent kinds while only `TokenDiff`/`Transfer`/`FtWithdraw` read it, so a `native_withdraw` carrying malformed metadata got a warning for metadata nothing in that envelope consults.

`7edce7f3`'s fix is the right one, and `intent_consumes_token_registry` is the right shape for it — an exhaustive 11-variant match with no wildcard arm, so a new intent kind is a compile error that forces a decision rather than defaulting into either behaviour. That is also what makes the per-kind parity test meaningful instead of a snapshot of today's list.

## Non-blocking points, all addressed

- **The naked 2-tuple is gone.** `RequestTokenRegistry { registry, diagnostics }` with an `empty()` constructor and a doc on each field. This is more than cosmetic: it is what lets the two gates read symmetrically, and "return `empty()`" is now an obvious, nameable thing to do — which is exactly what made "emit every rejection you were handed" the path of least resistance before.
- **`charset_safe`'s docstring** now names the token-metadata refusal as a caller rather than being written for `memo`/`msg` alone.
- **The `Result`-returning closure** for the extraction loop: declined with reasoning I accept. It rewrites the whole loop in the PR currently gating the cascade, and it is genuinely separable. I would still like it eventually — the macro-comment invariant that finding 1 broke is the kind of thing only the compiler holds reliably — but not here.

## Also read

`output.rs`'s new `Diagnostic` arm renders `[level] rule (instruction N): message` instead of falling through to the accessor arm, which would have reduced a diagnostic to its label and fallback text and dropped the level, rule and index. Correct call, and the comment says why. Splitting `write_diagnostic` out to keep the match arm one line is fine.

**One nit:** the diff drops the blank line between `parse_and_display`'s closing brace and `#[cfg(test)]` in `output.rs`. Looks accidental rather than intended.

PR #440, opened 2026-08-06
feat(prs-556): make ABI trust posture an explicit policy

pepe-anchor:
## Why am I making this PR?

The parser currently makes ABI trust decisions per-request, meaning a caller can opt into leniency by omitting a signature. The trust posture should be a deploy-time choice, fixed at startup and auditable in the TVC manifest.

## What am I changing?

Add `MetadataTrustPolicy` enum (AcceptUnsigned | RequireAllowlistedSigner) and `SignerAllowlist` to `visualsign::signing`. No Default impl — a deployment must explicitly state which posture it runs. Marked `#[non_exhaustive]` so future posture variants are safe.

## What is the Linear ticket?

PRS-556

## What are the rollback steps?

Revert the commit. No wire-format or API changes at this layer.

## Is this change backwards compatible?

Yes. No existing callers are broken — `MetadataTrustPolicy` has no Default impl so every construction site is explicit.

## Does this require cross-team/service coordination?

No.

## How do I know it works as designed? Which tests exercise this code?

Existing signing.rs unit tests cover SignerAllowlist construction and policy accessors.

### DIFF
--- a/src/visualsign/src/signing.rs
+++ b/src/visualsign/src/signing.rs
@@ -127,12 +127,10 @@ impl FromIterator<Vec<u8>> for SignerAllowlist {
 /// life of the process. Nothing in a request can move the parser between the two
 /// variants: a caller cannot opt itself into leniency by omitting a signature,
 /// and cannot opt itself into strictness either. That is the point of the type.
-/// The intent is for the choice to live on the parser's cmdline (for the
-/// enclave binary, in the `pivotArgs` of the signed TVC manifest), so a signer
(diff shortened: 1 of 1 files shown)


pepe-anchor:
Closing in favour of a replacement PR. This one is still the last live member of GitHub stack #445 (the other four entries are the closed/merged #441-#444), which locks its base against retargeting and keeps showing a stack topology that no longer exists. Same branch, same commit, no content change.

PR #441, opened 2026-08-06
feat(prs-556): wire deploy-time ABI trust posture into parser_app

pepe-anchor:
## Why am I making this PR?

Wire the MetadataTrustPolicy from PR #440 into parser_app so deployments express their posture via cmdline flags instead of inheriting an implicit default.

## What am I changing?

Add `--accept-unsigned-abis` and `--accept-signatures-from-pubkey` flags to parser_app CLI. Mutually exclusive (clap `forbids`), at least one required at startup. Parse into `ParserConfig.abi_trust` and thread through `create_registry` to the converter.

## What is the Linear ticket?

PRS-556

## What are the rollback steps?

Redeploy with a manifest whose pivotArgs omit the new flags. The old parser_app binary ignores unknown flags.

## Is this change backwards compatible?

No — parser_app now refuses to start without an explicit posture. Existing launch paths (Makefile, integration harness, CI) pass `--accept-unsigned-abis` explicitly.

## Does this require cross-team/service coordination?

Yes — live deployments need a redeploy to pick up a posture.

## How do I know it works as designed? Which tests exercise this code?

CLI arg parsing tests cover both/neither rejection. Registry test pins both postures reach the converter.


pepe-anchor:
Closing in favour of a replacement with the correct base. GitHub refuses to retarget this PR ('Cannot change the base branch because the pull request is part of a stack'), and it needs to sit on prs-556-ethereum-converter rather than prs-556-core-policy: parser_app consumes `abi_metadata::signer_allowlist_from_hex`, which the eth-converter commit introduces. No comments or reviews here to lose.

Issue #337, opened 2026-05-26
parser_gateway: emit bootProof in response (gated env, mock for non-TEE)


prasanna-anchorage:
## Problem

The local parser_gateway returns a `TurnkeyResponseWrapper` (`src/parser/gateway/src/main.rs:65-69`) that contains only `response` and an optional `error`. The production Turnkey visualsign API — the shape wallet integrators code against — returns a top-level `bootProof` field alongside `response` carrying AWS Nitro / QoS attestation data.

As a result, anything talking to the local container gets a response that doesn't match the production contract. Downstream consumers that strictly follow the wallet-integration spec (reference Go client at [visualsign-turnkeyclient/api/types.go::TurnkeyVisualSignResponse](https://github.com/anchorageoss/visualsign-turnkeyclient/blob/main/api/types.go#L109)) drop the field on the floor and any further pipeline step that asserts `bootProof` is present fails.

Concretely: we just chased exactly this regression downstream — a parser image bump moved the response shape and only a 50-minute e2e suite hitting **preprod Turnkey** caught it, because the locally-built image never emitted bootProof to begin with. Self-contained CI tests against the local container were blind by design.

The upstream CLAUDE.md is candid about the gap: *"Same API as production TDX deployment, only difference is no attestation."* That "only difference" turns out to matter quite a lot for contract conformance.

## Proposal

Add a stable, deterministic mock bootProof to every gateway response:

- Extend `TurnkeyResponseWrapper` with `boot_proof: TurnkeyBootProof` at the top level (so the wire shape matches production).
- Populate it with a fixed placeholder matching the production field set:
  - `awsAttestationDocB64` — fixed base64 sentinel
  - `qosManifestEnvelopeB64` / `qosManifestB64` — fixed base64 sentinels
  - `ephemeralPublicKeyHex` — fixed sentinel hex
  - `enclaveApp` — `"visualsign-parser"`
  - `deploymentLabel` — `"local-mock"`
- The gateway runs only in non-TEE local dev/CI (production deployments wrap parser_app directly inside the Nitro enclave), so emitting a mock is always safe — there's no path by which these values reach a production consumer.

The values do not need to verify against AWS Nitro — they only need to be present and well-shaped. Downstream consumers that perform real verification will reject them, which is correct: the mock is for shape conformance / contract testing, not for letting signing actually succeed.

## Why not just hit preprod from CI

Two reasons:

[...]

PR #446, opened 2026-08-06
refactor(host_primitives): single home for the Turnkey wire envelope

pepe-anchor:
## Why

Two host binaries have to serve the same Turnkey JSON envelope to the same client.

- `parser_gateway`: REST in front, gRPC to `parser_grpc_server` behind. Non-TEE local dev and CI, so it emits a mock `bootProof`.
- `parser_http_server` (#450): HTTP+JSON inside the enclave, calling `parser_app` in process, no gRPC hop. This is what gets deployed to the TVC, because Cloudflare in front of `app-<uuid>.turnkey.cloud` rejects gRPC with 403. It emits a real attested `bootProof`.

Different transports, different trust levels, one wire contract. The Go [`visualsign-turnkeyclient`](https://github.com/anchorageoss/visualsign-turnkeyclient) and the wallet integrators behind it cannot tell the two apart, and must not be able to.

Two features also land in that envelope during PRS-581: `bootProof` (#337) and `intermediateOutput` (#414). If the definition stays inline in `parser_gateway`, the pivot copies it, and every envelope change from here on gets made twice and kept in sync by hand. The unmerged x402 branch had already grown its own third copy.

So: one home, two importers, and the `bootProof` difference becomes a parameter instead of a fork.

Downstream, #450, #451 and #452 build on `parser_http_server`; #449 builds on `parser_gateway`. Both sides import the envelope from here, which is also what shrinks the PR #304 rebase from "reconcile two envelope definitions" to "add a module".

## What

- `host_primitives::turnkey` becomes the single home for the Turnkey request/response envelope, as the union of both existing definitions.
- `bootProof` is now an injected value rather than a hardcoded mock: `error_response(msg, boot_proof)`. `parser_gateway` keeps its stable local-dev mock through a local shim; the enclave pivot supplies a real attested one.
- Construction moved with the types: `success_response` sits next to `error_response`, so `parse_handler` no longer assembles the success envelope field by field in the gateway.
- `parser_gateway` imports the shared types and keeps the four `MOCK_BOOT_PROOF_*` constants and the tests that are about its own mock and top-level wire shape. Three tests that would have duplicated coverage now living in `host_primitives::turnkey` moved there instead of being maintained in both places.
- `intermediate_output` gained `serde(default)` alongside its existing `skip_serializing_if`. Without it the omitted key serialized fine but failed to deserialize, so a response could be written and not read back. Nothing in tree deserializes a response today (both binaries only emit one), which is why no test caught it. The client-direction derives are there so the envelope is symmetric for the out-of-tree clients that do read it, and the new round-trip test is what keeps it that way. Happy to drop those derives and the `default` until something in tree needs them, if you'd rather not carry them.

No wire change. The gateway's JSON output is byte-identical.

## Test evidence

```
cargo test -p host_primitives -p parser_gateway
  host_primitives: 5 passed
  parser_gateway: 11 passed
cargo fmt --all -- --check: clean
make -C src lint: clean
```

The gateway test count moves from 13 to 11 because three tests relocated to `host_primitives::turnkey`, where the types they cover now live: the six-key `bootProof` assertion (`boot_proof_wire_shape_is_exactly_six_camel_case_keys`), the empty `intermediateOutput` omission check, and the Solana chain-metadata discriminator test. `host_primitives` goes 4 to 5 with the new round-trip test. No coverage was dropped.

The two regression tests that matter both still pass with unchanged constant values: `mock_boot_proof_matches_production_wire_shape`, `error_response_carries_mock_boot_proof`.

The "no wire change" claim was checked rather than assumed: a throwaway test reconstructed the pre-refactor struct definitions verbatim from base and byte-compared `serde_json` output against the new `host_primitives::turnkey` types across success, error, signature-present, and intermediate-output-present cases. All identical.

One trap worth recording: the x402 branch's `TurnkeyResponseWrapper` had no `rename_all = "camelCase"`, which the gateway's inline struct did have. Adding `boot_proof` without it would have serialized as `boot_proof` instead of `bootProof` and silently broken the wallet contract. The six-key wire-shape test catches it.

## Rollback

Revert the commits. No deploy, no migration, no wire change, so a revert restores the previous state exactly.

## Linear

PRS-581

Stack position: base of the PRS-581 stack. #337, #414.

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -4229,8 +4229,11 @@ name = "host_primitives"
 version = "0.1.0"
 dependencies = [
  "borsh 1.6.0",
+ "generated",
  "prost 0.12.6",
  "qos_core",
+ "serde",
+ "serde_json",
  "tonic 0.10.2",
 ]
 
--- a/src/host_primitives/Cargo.toml
+++ b/src/host_primitives/Cargo.toml
@@ -16,6 +16,11 @@ borsh = { version = "1.0", features = [
   "std",
   "derive",
 ], default-features = false }
+generated = { path = "../generated", features = ["serde_derive"] }
+serde = { workspace = true }
+
+[dev-dependencies]
+serde_json = { workspace = true }
 
 [lints]
 workspace = true
--- a/src/host_primitives/src/lib.rs
+++ b/src/host_primitives/src/lib.rs
@@ -9,6 +9,8 @@ use prost::Message;
 use qos_core::protocol::{ProtocolError, msg::ProtocolMsg};
 use tonic::Status;
 
+pub mod turnkey;
+
 /// Buffer size for socket message queue.
 pub static ENCLAVE_QUEUE_CAPACITY: usize = 12;
 /// Maximum gRPC message size. Set to 25MB (25*1024*1024)
--- a/src/host_primitives/src/turnkey.rs
+++ b/src/host_primitives/src/turnkey.rs
@@ -24,8 +24,9 @@ pub struct TurnkeyRequest {
     pub include_intermediate_output: bool,
 }
 
-// ChainMetadataInput needs Serialize too so parser_gateway can build the
-// outbound HTTP body symmetrically with what parser_http_server parses.
+// Serialize is derived on the request types (not just Deserialize) for a future
+// caller that builds this envelope symmetrically to what parses it; no in-tree
+// consumer serializes a request today.
 
 /// Tagged representation of chain metadata for unambiguous JSON deserialization.
 ///
@@ -138,6 +139,22 @@ pub fn error_response(msg: String, boot_proof: TurnkeyBootProof) -> TurnkeyRespo
             },
         },
         error: Some(msg),
+    }
+}
+
+/// Success envelope. `boot_proof` is injected by the caller for the same reason as
+/// in [`error_response`].
+pub fn success_response(
+    boot_proof: TurnkeyBootProof,
+    payload: TurnkeyPayload,
+    signature: Option<TurnkeySignature>,
+) -> TurnkeyResponseWrapper {
+    TurnkeyResponseWrapper {
+        boot_proof,
+        response: TurnkeyResponse {
+            parsed_transaction: TurnkeyParsedTransaction { payload, signature },
+        },
+        error: None,
     }
 }
 
--- a/src/host_primitives/src/turnkey.rs
+++ b/src/host_primitives/src/turnkey.rs
@@ -1,0 +1,230 @@
+//! Turnkey-compatible request/response envelope for parse endpoints.
+
+use generated::parser::{ChainMetadata, EthereumMetadata, SolanaMetadata, chain_metadata};
+use serde::{Deserialize, Serialize};
+
+/// SHA-256 of empty input: used as the canonical "no data" sentinel for digest fields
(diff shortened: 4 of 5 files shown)


shahan-khatchadourian-anchorage (approved):
Right call: one home, `boot_proof` as a parameter instead of a fork. Verified locally — 5 + 11 tests pass, clippy and `fmt --check` clean, and no-wire-change holds.

**Good:** mock stays local to the gateway; parity was byte-compared rather than asserted; test relocation accounted for (13 -> 11, 4 -> 5); the `untagged` hazard docstring and its test moved with the type; the x402 `rename_all` trap is covered by the six-key test.

**Notes**, all non-blocking:

1. **Take the offer on the client-direction derives.** Drop `Serialize` on request types, `Deserialize` on response types. Worth noting none of this trips `dead_code` — `pub` items in a lib crate never do, so clippy passing says nothing about reachability here.
   - Serializing a request emits `"signature":null,"abiType":null,...` and mixes casing (`unsigned_payload` + `networkId`). Not a regression — the read path has always accepted that shape (`README.md:57`) — but `Serialize` is now what emits it, and nothing pins the request wire shape the way the six-key test pins the response.
   - No inverse of `From<ChainMetadataInput> for ChainMetadata`, so a caller holding the proto type can't reach the serialize path.
   - Deserialize is stricter than serialize — a response missing one `bootProof` field fails on `missing field deploymentLabel`. The round-trip test only reads back what the same type wrote.

   I measured the trim (compiles, 4 + 11 pass): 266 -> 242 source lines, but expanded 2251 -> 883 and rlib 364KB -> 165KB. Side effect: commit 3 becomes moot, since `default` on `intermediate_output` goes inert without `Deserialize`. Keep the attribute anyway; PR drops to two commits. The comment at `turnkey.rs:26-28` is aspirational per `CLAUDE.md` — belongs in the PR body.

   If you'd rather keep the derives, the three things that would make them real: a request wire-shape test, a `TryFrom<ChainMetadata> for ChainMetadataInput`, and a call on null-vs-omit.

2. **`default` on `chain_metadata` is a no-op** — serde already defaults `Option<T>` to `None` (checked). Inconsistent with `error`/`signature`, which carry only `skip_serializing_if`. `rename_all` on `ChainMetadataInput` is also dead (both variants explicitly renamed), but that's pre-existing.

3. **The enclave picked up serde derives it doesn't use.** `parser_app` links `host_primitives`, so `generated` goes from `tonic_types` to `serde, serde_derive, tonic_types`. `images/parser_app/Containerfile:16-18` reasons about attestation surface and the vsock image has no HTTP surface; #450 enables the feature itself. A non-default `turnkey` feature covers it. Survives the trim in (1). Minor: `health_check`'s `default-features = false` on `host_primitives` is a no-op — no features declared.

4. **The duplication that hurts didn't move.** #449:1314 and #450:970 hand-build the success envelope field by field, so `success_response` has one caller and the two PRs motivating it bypass it. The triplicated part is proto -> wire: `Chain::from_str_name`, the `SignatureScheme as i32` match, the base64 encode, the `parsed_transaction`/`payload` unwrapping. `success_response_from_proto(...)` would collapse all three.

5. **#304 is the third envelope copy**, still open on `spec/x402-gated-http-api` — worth confirming it rebases onto this rather than reconciling definitions.

PR #447, opened 2026-08-06
ci(stagex): build parser_http_server and parser_grpc_server images

pepe-anchor:
## Why

PRS-581 deploys `parser_http_server` as the TVC pivot. The stagex "TVC deployment details" step hardcodes `parser_app`, so it is the only binary that ever gets a reproducible executable digest and a paste-ready deploy block. Without this, the pivot cannot be deployed or probed at all, which blocks the `/dev/nsm` reachability probe and every later step in the stack.

Landing it early so image builds are never the thing holding the stack up.

## What

- Adds `parser_grpc_server` and `parser_http_server` to the stagex build matrix.
- Generalizes the deployment-details step over the matrix target: `TARGET_NAME` env, per-target container URL, per-target extract container and binary path, per-target deploy-md and release-note sentinels so parallel matrix legs cannot clobber each other's release body edits.
- Adds `images/parser_http_server/Containerfile`, mirroring `parser_app`'s. It keeps `--features "vsock ${CHAIN_FEATURES}"`, which is load-bearing: `vsock` switches `qos_core::EPHEMERAL_KEY_FILE` to the in-enclave path, and without it the deployed pivot panics at startup and TVC reports 0 healthy replicas.

Two changes from the x402 branch were deliberately NOT taken, because they belong to other PRs:
- the PR-label rename `ci:stagex` to `stagex`. Main's label is kept.
- `qos_version="v2026.2.6"`. Main's `0.12.0` is kept; the QOS bump is its own PR in this stack.

## Test evidence

The stagex path only runs on a labelled PR, so this needs the `ci:stagex` label to be validated. What was verified locally:

```
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/stagex.yml'))"  -> parses
grep 'label.name' stagex.yml  -> 'ci:stagex'  (unchanged from main)
grep -c 'v2026.2.6' stagex.yml -> 0
```

I have NOT run the stagex build itself. Please add `ci:stagex` and confirm all matrix legs go green, including the two new ones, before merging.

## Rollback

Revert the commit. CI-only change plus one new Containerfile; nothing is deployed by merging it.

## Linear

PRS-581

Stack position: independent of the rest of the stack, based on `main`. Can merge in any order.

### DIFF
--- a/.github/workflows/stagex.yml
+++ b/.github/workflows/stagex.yml
@@ -151,33 +151,34 @@ jobs:
           tvc_tag="${SEMVER_TAG:-$SHA_TAG}"
           container_url="ghcr.io/anchorageoss/${TARGET_NAME}:${tvc_tag}"
           container_digest=$(docker inspect --format='{{index .RepoDigests 0}}' "${container_url}" | cut -d '@' -f 2)
-          docker create --name tmp-extract-${TARGET_NAME} "${container_url}" /bin/true
-          docker cp "tmp-extract-${TARGET_NAME}:/${TARGET_NAME}" "/tmp/${TARGET_NAME}"
-          docker rm "tmp-extract-${TARGET_NAME}"
-          digest=$(sha256sum "/tmp/${TARGET_NAME}" | awk '{print $1}')
+          docker create --name tmp-extract "${container_url}" /bin/true
+          docker cp "tmp-extract:/${TARGET_NAME}" /tmp/binary
+          docker rm tmp-extract
+          digest=$(sha256sum /tmp/binary | awk '{print $1}')
           pinned_url="${container_url}@${container_digest}"
           # Hardcoded: TVC currently only supports this QOS version.
           qos_version="0.12.0"
 
-          # parser_app is the enclave-side binary (signs responses).
-          # parser_gateway is the host-side binary (verifies the signature
-          # against a pinned pubkey + handles x402 settlement). Operators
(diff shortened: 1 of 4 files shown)


pepe-anchor:
Closing unmerged. The features moved into #450.

The split did not work: this PR added `images/parser_http_server/Containerfile` and the stagex matrix entry, but the `parser/http-server` crate they build lands in #450. Neither half built on its own, and PR 03 carried no Containerfile, Makefile rule, or stagex entry of its own.

Two other things were broken here and are fixed in #450:

- The root Makefile had no rule for `out/parser_grpc_server/index.json` or `out/parser_http_server/index.json`, so both new legs died at the Build step before Docker ran. `images/parser_grpc_server/Containerfile` was missing entirely.
- Generalizing the deployment-details step over the matrix took it from one leg to three running concurrently, each doing `gh release view` then `gh release edit` with a full-body overwrite. Per-target sentinels keep a leg idempotent across re-runs but do not serialize legs, so whichever edit landed last silently dropped the others sections. #450 retries with a fresh read and verifies the write stuck.

Verified on the merged branch: all five `make -n out/<target>/index.json` rules resolve, the workflow parses and every `run:` block passes `bash -n`, and both new targets compile as their Containerfiles build them (`cargo check -p parser_http_server --no-default-features --features "vsock ethereum solana sui tron unspecified"` and `cargo check -p parser_grpc_server --no-default-features`).

PR #285, opened 2026-05-07
test(surfpool): native cargo runner via build.rs + per-IDL macro

shahan-khatchadourian-anchorage:
## Summary

Stacked on #202. Replaces the bash runner (`scripts/surfpool_fuzz_all_idls.sh`) with a native Rust setup so `cargo test` itself enumerates the per-IDL roundtrip tests.

- `chain_parsers/visualsign-solana/build.rs` runs `cargo metadata` at build time, locates `solana_parser`'s IDL directory (the one matching the rev in `Cargo.lock`), and emits it as `cargo:rustc-env=SOLANA_IDL_DIR`. Tests pick it up via `env!()` — no runtime cargo invocation.
- A `macro_rules!` macro in `tests/surfpool_fuzz.rs` generates one `#[tokio::test]` per IDL (`surfpool_idl_<name>`). Cargo's harness handles enumeration, parallelism, pass/fail counts, and per-test failure diagnostics — the things the bash script was reimplementing.
- The surfpool test path now distinguishes the three failure modes (decode rejected, no instructions, missing discriminator) with messages that name the IDL file. Previously a malformed IDL produced a misleading "no discriminator" panic.
- `collision.json` and `cyclic.json` are excluded: they're solana_parser's negative test fixtures (duplicate type names / cyclic type refs) intentionally rejected by `decode_idl_data`.

## CI change

The workflow drops the bash invocation and runs:
```
cargo test -p visualsign-solana --test surfpool_fuzz -- --ignored --test-threads=1
```
`--test-threads=1` serialises surfpool spawns so each test owns its mainnet fork.

## Documentation

Adds a Testing Patterns bullet in `CLAUDE.md` covering the macro-based pattern and how to add a new IDL.

## Test plan

- [x] Local: `HELIUS_API_KEY=<key> cargo test -p visualsign-solana --test surfpool_fuzz -- --ignored --test-threads=1` — 14 passed in ~9s.
- [ ] Add the `surfpool` label on this PR (after #202 merges) to exercise CI.

## Why this PR

Addresses the third-party reviewer comment from #202 ("you might be able to do all this in Rust itself"). The narrow interpretation was already done there via the `idl-meta` Rust tool; this PR closes the loop by removing the bash orchestrator entirely.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #458, opened 2026-08-06
fix(solana_test_utils): reject confirmed-but-failed airdrops, redact datasource credentials

### LINKED ISSUE #291: fix(solana_test_utils): SurfpoolManager::airdrop returns Ok on confirmed-but-failed transactions
Follow-up from #202 (and unaddressed by stacked #285, #289).

## What

`SurfpoolManager::airdrop` returns `Ok(signature)` whenever the RPC returns a status, regardless of whether the inner `Result` indicates success or `TransactionError`.

## Where

`src/solana_test_utils/src/surfpool/manager.rs`, around line 142:

```rust
for _ in 0..max_attempts {
    if let Ok(Some(_status)) = client.get_signature_status(&signature) {
        return Ok(signature);
    }
    tokio::time::sleep(Duration::from_millis(100)).await;
}
```

`get_signature_status` returns `Result<Option<Result<(), TransactionError>>>`:

- `Ok(Some(Ok(())))` — confirmed success
- `Ok(Some(Err(tx_err)))` — confirmed **failure** ← currently treated as success
- `Ok(None)` — not yet finalized
- `Err(rpc_err)` — RPC-level error

The match `Ok(Some(_status))` discards the inner result. A confirmed airdrop failure is reported as success.

## Why it matters

Latent bug — nothing in-tree currently calls `airdrop()`, but the moment a test relies on it for funding (e.g. to land a real on-chain transaction), confirmed failures would silently pass and downstream assertions would fail in misleading ways.

## Suggested fix

```rust
match client.get_signature_status(&signature) {
    Ok(Some(Ok(()))) => return Ok(signature),
    Ok(Some(Err(tx_err))) => {
        return Err(anyhow::anyhow!(
            "airdrop {} confirmed with transaction error: {tx_err}",
            signature
        ));
    }
    Ok(None) | Err(_) => {
        tokio::time::sleep(Duration::from_millis(100)).await;
    }
}
```

(Or `client.confirm_transaction(&signature)?` if a higher-level helper exists in the `solana-client` rev we depend on.)

shahan-khatchadourian-anchorage:
Closes #291.

## What

`SurfpoolManager::airdrop` reports a confirmed-but-failed airdrop as success.
`get_signature_status` returns `Result<Option<Result<(), TransactionError>>>`,
so a confirmed failure is still *a status*, and matching `Ok(Some(_status))`
discards the inner result.

Three commits:

**1. Reject confirmed-but-failed airdrops.** All four cases are matched
explicitly: only `Ok(Some(Ok(())))` returns the signature, a confirmed
`TransactionError` returns `Err` naming it, `Ok(None)` retries, and an
RPC-level error retries while retaining the error for the timeout message.
Both blocking RPC calls run under `spawn_blocking` behind an `Arc<RpcClient>` —
the blocking client calls `block_in_place` internally, which panics on a
`current_thread` runtime, so the previous direct call made `airdrop` unusable
from a plain `#[tokio::test]`.

**2. Keep datasource credentials out of logs.** `SurfpoolConfig::default()`
builds `rpc_url` from `HELIUS_API_KEY`, and two log sites render it: the
config's `Debug` output and the spawn argument list. `redact_url_credentials`
reduces a URL to scheme and host, covering credentials in the query string,
the path, and the userinfo. A hand-written `Debug` for `SurfpoolConfig` routes
`rpc_url` through it, so every `{:?}` is covered rather than one call site.

**3. Cover every confirmation outcome offline.** `airdrop_with` takes the
client and retry budget as arguments, so `RpcClient::new_mock` can drive each
outcome without a validator. Four tests pin all four arms and run in normal CI
in milliseconds.

## Verification

- `cargo test -p solana_test_utils --lib` — 11 passed (4 confirmation arms,
  7 redaction).
- **Mutation-checked:** reintroducing `if let Ok(Some(_status))` fails the two
  confirmed-failure tests and leaves success and timeout passing, so the
  regression cannot return silently.
- `cargo test -p solana_test_utils --test airdrop -- --ignored` passes against
  a live surfpool mainnet fork.
- `cargo clippy -p solana_test_utils --all-targets -- -D warnings` and
  `cargo fmt --check` clean under the pinned 1.88.0 toolchain.

## Notes for review

- The credential redaction is a separate concern from the airdrop fix, kept as
  its own commit so it can be read independently. Neither log site currently
  emits anywhere — nothing in these tests installs a `tracing` subscriber — so
  this is defense-in-depth for whenever one is present, not a live leak.
- `tests/airdrop.rs` is `#[ignore]` and no CI job runs it: the only `--ignored`
  runner is scoped to `-p visualsign-solana --test surfpool_fuzz`. The four
  offline tests cover the logic; wiring the live test into
  `surfpool-solana.yml` is left for after #285 rewrites that workflow, to avoid
  authoring a conflict.
- The live test compares a balance delta rather than an exact balance:
  `Pubkey::new_unique` is a counter, not a random source, so the target address
  is identical on every run.
- A follow-up worth considering, out of scope here:
  `solana_client::nonblocking::rpc_client::RpcClient` would delete the `Arc`,
  both `spawn_blocking` hops, and the nested per-client runtime — but
  `rpc_client()` must keep returning the blocking client for existing callers,
  so adopting it means two client types in one file.


prasanna-anchorage (approved):
Approving. Sorry this sat for five weeks — it was in neither of my tracking docs, which is on me. The airdrop half is right, and the PR body is unusually good: the mutation check, the `#[ignore]`/no-runner note, and the `Pubkey::new_unique`-is-a-counter detail are all things a reviewer would otherwise have to discover.

`cargo test -p solana_test_utils --lib` 11 passed; `make lint` clean; `make test` 39 suites / 1547 tests, 0 failures.

**The airdrop fix is correct.** All four arms of `Result<Option<Result<(), TransactionError>>>` are matched explicitly with no wildcard, so a change in that shape is a compile error rather than a silent re-classification. A confirmed `TransactionError` returns immediately instead of burning the retry budget on something already final, `Ok(None)` retries, and the RPC error is retained for the timeout message rather than discarded. The `spawn_blocking` reasoning checks out too — `block_in_place` panicking on a `current_thread` runtime made `airdrop` unusable from a plain `#[tokio::test]`, so that's a real second bug fixed in passing.

I also confirmed the framing in your notes: `solana_test_utils` is only in `visualsign-solana`'s `[dev-dependencies]`, and nothing in its reach installs a subscriber — so "defense-in-depth, not a live leak" is accurate, and that's why the following is non-blocking rather than a hold.

## The redaction should use `url`, not hand-rolled string splitting

`url v2.5.7` is **already in `Cargo.lock`** (transitive through the Solana deps), so this isn't a new-dependency argument. Two concrete things the hand-rolled version gets wrong, both of which are the characteristic failure of parsing URLs by hand — a component you didn't think of:

**1. The fragment is not redacted.** `authority_end = rest.find(['/', '?'])` doesn't include `#`:

```
redact_url_credentials("https://rpc.example.com#SECRET") -> "https://rpc.example.com#SECRET"
```

Narrow in practice (nobody puts an RPC key in a fragment) but it's a whole URL component the function doesn't know exists.

**2. The arg-list caller sniffs, so it bypasses the function on exactly the input the function was written to handle.** `manager.rs` gates on `if arg.contains("://")`, while `Debug` redacts unconditionally. So the two sites disagree:

```
  https://mainnet.helius-rpc.com/?api-key=SECRET | contains("://") = true  -> redacted
          mainnet.helius-rpc.com/?api-key=SECRET | contains("://") = false -> logged verbatim
```

That's reachable, not theoretical: `SurfpoolConfig::default()` takes `SOLANA_RPC_URL` from the environment verbatim (`config.rs:74`), with no requirement that it carry a scheme, and `with_rpc_url` accepts any string. And the scheme-less case is one you clearly *intended* to handle — `redacts_schemeless_input` tests it at the function, which is what makes the caller excluding it worth fixing.

The fix for (2) is worth doing regardless of `url`: redact by **identity** rather than by content — the arg after `--rpc-url` is the credential-bearing one, and you already know which index that is when you build `args`. Sniffing for `://` asks the string what it is; the arg list already knows.

With `Url::parse` both collapse into `scheme` + `host_str` + `port` and everything else dropped by construction, with no component list to keep in sync. One thing to decide explicitly if you go that way: `Url::parse` **fails** on scheme-less input (`relative URL without a base`), so the parse-error arm must emit something opaque — never the raw string. That's the classic leak in library-based redaction, and it's the mirror of the gap in (2).

## Smaller notes

1. **`Debug` is hand-written, which is the right call, and worth a line saying so.** It means adding `#[derive(Debug)]` later is a duplicate-impl compile error rather than a silent un-redaction — a real protective property that reads like an accident. If a second credential-bearing field ever lands (a second datasource, an explicit key field), nothing forces it through the redactor, so a one-line comment on the struct is cheap insurance.
2. **A host-embedded credential still prints.** Some providers key by subdomain (`https://<key>.provider.com`) rather than query or path. Helius doesn't, so this is a note rather than a gap — but the docstring says "Datasource URLs carry credentials in three places", and it's four.
3. **It's based on `main` @ 08-06.** Still `MERGEABLE`, and my `make test` run here shows 39 suites against 43 on current `main`, so it's ~5 weeks behind. Worth a rebase before merge so CI runs against what it will actually land on.
4. Agreed on the `nonblocking::RpcClient` follow-up being out of scope, and the reason given (`rpc_client()` must keep returning the blocking client) is the right one.

PR #459, opened 2026-08-06
fix(near): sanitize untrusted strings on the intents render path

shahan-khatchadourian-anchorage:
Stacked on #439.

Every caller-controlled string the intents renderer puts on the signing screen goes through `charset_safe`. The core charset validator permits `\n` as the wallet's documented multi-line separator (`validate_charset`, and its own comment says attacker-controlled strings must be sanitized at the insertion site), so an unfiltered `memo` or `msg` renders as extra apparent confirmed fields the parser never authored:

```
   ├─ To: 608fe1e7...
   ├─ Amount: 1 (unresolved nep141:sol.omft.near)
   └─ Message: innocent
To: alice.near
Amount: 0.001 SOL
```

The sibling borsh transaction path already filters at each insertion site. `charset_safe` moves to `fmt.rs` so both NEAR paths share one definition rather than the intents path reaching across into `actions.rs`.

## Sinks covered

| Field | Where |
|---|---|
| `memo` | `token_diff`, `transfer`, and all three withdraws |
| `msg` | `transfer`'s notification, the withdraws, `auth_call` |
| `token_id` | `nft_withdraw` (`non_fungible_token::TokenId` = `String`) |
| `token_ids` | `mt_withdraw` (`defuse_nep245::TokenId` = `String`) |
| asset id | the unresolved-amount fallback and the `unverified-token-metadata` diagnostic |
| every diagnostic message | filtered inside `diagnostic()` itself |

`referral` is exempt and stays unfiltered: it is an `AccountId`, whose own charset rules already exclude everything the filter strips, so filtering it would be dead code.

`TokenId` is **not** exempt despite its account-id-shaped prefix — its `FromStr` parses only the contract half as an `AccountId` and takes the remainder verbatim into a plain `String`, which `Display` round-trips. Verified against the pinned `defuse-core` rev `6dad94c`.

## Two details worth a reviewer's eye

**Resolution runs on the raw asset id; only the rendered form is filtered.** Filtering first would let a crafted id collapse onto a seeded one — `nep141:wrap\u{7f}.near` -> `nep141:wrap.near` — and borrow that token's symbol and decimals. Pinned by `asset_id_resolves_on_its_raw_form_not_its_sanitized_form`.

**Filtering inside `diagnostic()` rather than at each call site.** Every rule on this path quotes untrusted input — an asset id, a `serde_json::Error` that interpolates the offending value with `{}` — so the choke point is what makes a newly added rule safe by construction.

## Behavior changes

- `mt_withdraw` renders its `memo`. It carries the field and silently dropped it, while `ft_withdraw` and `nft_withdraw` both rendered theirs. Folding `memo` into `push_withdraw_call_details` alongside `msg` and `storage_deposit` closes it by construction.
- Filtering happens before the emptiness test, so an all-non-ASCII memo drops out instead of rendering as a blank `Memo` line. Applied at all three memo sites.
- Filtering strips rather than rejects: a legitimate memo with an accented character or emoji loses those characters instead of failing the whole parse. This matches the borsh path and keeps a non-ASCII memo from denying the signer their transaction.

## Testing

Tests assert exact output rather than the absence of a newline, so a filter that dropped only `\n` while passing `\t`, `\r`, control bytes or backslashes would still fail. `charset_safe` gains direct coverage in `fmt.rs` including the bidi-override and all-non-ASCII cases.

205 tests pass with default features, 189 with `--no-default-features` (the `Warning`-text build), clippy clean, full `make test` green.

### DIFF
--- a/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
+++ b/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
@@ -12,6 +12,8 @@ use visualsign::errors::VisualSignError;
 use visualsign::field_builders::{create_amount_field, create_text_field};
 use visualsign::registry::LayeredRegistry;
 
+use crate::fmt::charset_safe;
+
 use super::tokens;
 use super::verify::SignatureCheck;
 use super::{NEAR_DECIMALS, NEAR_SYMBOL, NearTokenRegistry};
@@ -64,7 +66,7 @@ pub(crate) fn rejected_metadata_diagnostics(
     rejected
         .iter()
         .filter_map(|r| {
-            let message = crate::actions::charset_safe(&format!(
+            let message = charset_safe(&format!(
                 "token metadata supplied for {} was rejected and not used: {}",
                 r.asset_id, r.reason
             ));
@@ -250,8 +252,10 @@ fn render_token_diff(td: &TokenDiff, registry: &Reg) -> Result<Fields, VisualSig
         )?);
     }
     if let Some(memo) = &td.memo {
-        fields.push(create_text_field("Memo", memo)?.signable_payload_field);
+        fields.push(create_text_field("Memo", &charset_safe(memo))?.signable_payload_field);
     }
+    // `referral` is an `AccountId`: its own charset rules already exclude
+    // everything `charset_safe` would strip, so filtering it would be dead code.
     if let Some(referral) = &td.referral {
         fields.push(create_text_field("Referral", referral.as_str())?.signable_payload_field);
     }
@@ -269,7 +273,7 @@ fn render_transfer(t: &Transfer, registry: &Reg) -> Result<Fields, VisualSignErr
         )?);
     }
     if let Some(memo) = &t.memo {
-        fields.push(create_text_field("Memo", memo)?.signable_payload_field);
+        fields.push(create_text_field("Memo", &charset_safe(memo))?.signable_payload_field);
     }
     // A `Transfer`'s optional notification changes what the transfer does
     // beyond moving the named tokens: `msg` calls `mt_on_transfer` on
@@ -277,7 +281,9 @@ fn render_transfer(t: &Transfer, registry: &Reg) -> Result<Fields, VisualSignErr
     // `_transfer_call` form), and `state_init` initializes the receiver's
     // contract in the same receipt.
     if let Some(notification) = &t.notification {
-        fields.push(create_text_field("Message", &notification.msg)?.signable_payload_field);
+        fields.push(
+            create_text_field("Message", &charset_safe(&notification.msg))?.signable_payload_field,
+        );
         if notification.state_init.is_some() {
             fields.push(state_init_field()?);
         }
@@ -285,18 +291,23 @@ fn render_transfer(t: &Transfer, registry: &Reg) -> Result<Fields, VisualSignErr
     Ok(fields)
 }
 
-/// A withdraw's optional `msg` (switches the call into its `_transfer_call`
-/// form, passing this to the receiver) and `storage_deposit` (a separate,
-/// unconditional wNEAR debit for the receiver's storage on `token`, never
-/// refunded on failure) -- both must render, since either changes what the
-/// withdraw actually does beyond moving the named token/amount.
+/// A withdraw's trailing fields, shared by all three token standards so none
+/// can silently drop one: the optional `memo`, the optional `msg` (switches
+/// the call into its `_transfer_call` form, passing this to the receiver), and
+/// `storage_deposit` (a separate, unconditional wNEAR debit for the receiver's
+/// storage on `token`, never refunded on failure). Each changes what the
+/// withdraw does beyond moving the named token/amount, so each must render.
 fn push_withdraw_call_details(
     fields: &mut Fields,
+    memo: &Option<String>,
     msg: &Option<String>,
     storage_deposit: Option<near_sdk::NearToken>,
 ) -> Result<(), VisualSignError> {
+    if let Some(memo) = memo.as_deref().filter(|m| !m.is_empty()) {
+        fields.push(create_text_field("Memo", &charset_safe(memo))?.signable_payload_field);
+    }
     if let Some(msg) = msg.as_deref().filter(|m| !m.is_empty()) {
-        fields.push(create_text_field("Message", msg)?.signable_payload_field);
+        fields.push(create_text_field("Message", &charset_safe(msg))?.signable_payload_field);
     }
     if let Some(deposit) = storage_deposit {
         fields.push(near_amount_field(
@@ -318,10 +329,7 @@ fn render_ft_withdraw(w: &FtWithdraw, registry: &Reg) -> Result<Fields, VisualSi
         w.amount.0,
         registry,
     )?);
-    if let Some(memo) = &w.memo {
-        fields.push(create_text_field("Memo", memo)?.signable_payload_field);
-    }
-    push_withdraw_call_details(&mut fields, &w.msg, w.storage_deposit)?;
+    push_withdraw_call_details(&mut fields, &w.memo, &w.msg, w.storage_deposit)?;
     Ok(fields)
 }
 
@@ -329,12 +337,12 @@ fn render_nft_withdraw(w: &NftWithdraw) -> Result<Fields, VisualSignError> {
     let mut fields = vec![
         create_text_field("Token", w.token.as_str())?.signable_payload_field,
         create_text_field("To", w.receiver_id.as_str())?.signable_payload_field,
-        create_text_field("NFT Token Id", w.token_id.as_str())?.signable_payload_field,
+        // `token_id` is a plain `String` (`non_fungible_token::TokenId`), not
+        // an `AccountId`, so it carries whatever bytes the sender chose.
+        create_text_field("NFT Token Id", &charset_safe(w.token_id.as_str()))?
+            .signable_payload_field,
     ];
-    if let Some(memo) = &w.memo {
-        fields.push(create_text_field("Memo", memo)?.signable_payload_field);
-    }
-    push_withdraw_call_details(&mut fields, &w.msg, w.storage_deposit)?;
+    push_withdraw_call_details(&mut fields, &w.memo, &w.msg, w.storage_deposit)?;
     Ok(fields)
 }
 
@@ -351,11 +359,15 @@ fn render_mt_withdraw(w: &MtWithdraw) -> Result<Fields, VisualSignError> {
         create_text_field("To", w.receiver_id.as_str())?.signable_payload_field,
     ];
     for (id, amount) in w.token_ids.iter().zip(w.amounts.iter()) {
+        // As with `nft_withdraw`'s `token_id`, an MT token id is a plain
+        // `String` (`defuse_nep245::TokenId`); filter it before it joins the
+        // composite, so the amount half cannot be pushed onto its own line.
         fields.push(
-            create_text_field("MT Token", &format!("{} x{}", id, amount.0))?.signable_payload_field,
+            create_text_field("MT Token", &format!("{} x{}", charset_safe(id), amount.0))?
+                .signable_payload_field,
         );
     }
-    push_withdraw_call_details(&mut fields, &w.msg, w.storage_deposit)?;
+    push_withdraw_call_details(&mut fields, &w.memo, &w.msg, w.storage_deposit)?;
     Ok(fields)
 }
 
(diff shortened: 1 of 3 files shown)


prasanna-anchorage (changes requested):
Verified against the real on-chain contract source (`defuse` crate, not just doc comments) — both premises hold:

- `NotifyOnTransfer.msg` is a required `String`, not `Option<String>` (`core/src/intents/tokens.rs:31-32`), and `notify_on_transfer` builds the `ext_mt_receiver` promise unconditionally whenever `notification` is `Some(_)` (`tokens.rs:74-80, 122-132`) — the *content* of `msg` never gates the callback.
- `ft_withdraw`/`nft_withdraw`/`mt_withdraw`'s `msg: Option<String>` really is the classic `Option` switch: `Some(_)` (including `Some("")`) takes the `_transfer_call` branch that invokes a receiver callback; `None` takes the plain non-callback branch (`nep141/withdraw.rs:141-145`, mirrored in `nep171`/`nep245`).

Two spots collapse "present but empty-after-sanitization" to the same render output as "absent," hiding that distinction from the signer:

1. **`render_transfer`** (`src/chain_parsers/visualsign-near/src/presets/intents/render.rs:314-320`) — this is a regression newly introduced by this PR. The base branch (`near-f-diagnostics`) rendered `notification.msg` unconditionally. This PR routes it through `nonempty_filtered`, so a `notification: Some(NotifyOnTransfer { msg: "", .. })` (or an all-non-ASCII `msg`) now renders zero fields for the notification — byte-identical to the no-notification case pinned by the existing `transfer_omits_notification_fields_when_absent` test (`labels == ["To", "Amount"]`). But `mt_on_transfer` still fires on-chain regardless of `msg` content, so the signer approves what looks like a plain transfer while a receiver-contract callback actually executes.

2. **`push_withdraw_call_details`** (`render.rs:337-341`) — this collapse already existed on the base branch for the empty-string case (`near-f-diagnostics`'s `render_base.rs:242` already filtered `msg.as_deref().filter(|m| !m.is_empty())`), so it's not new to this PR, but this PR entrenches it by extending the same filtering to also catch an all-non-ASCII `msg`. Either way, `Some("")` or `Some(<all-filtered-string>)` still switches the withdraw into the callback-invoking `_transfer_call` variant on-chain, with no visible "Message" field to distinguish it from a plain, non-callback withdraw.

Neither case is covered by a test with `msg` set to `""` or an all-non-ASCII string, and nothing upstream bounds `msg` to be non-empty.

Suggested fix: gate visibility on the field's *presence* (`Option::is_some()`), not on the sanitized text being non-empty — e.g. still push a "Message" field (with an explicit "(empty)" or similar fallback text) whenever `notification`/`msg` is `Some(_)`, and reserve `nonempty_filtered` for fields where presence genuinely carries no independent on-chain meaning.

Requesting changes on both spots; #1 should block, #2 is worth fixing in the same pass since this PR already touches the surrounding code.

shahan-khatchadourian-anchorage:
Rebased onto #439's new head (`023cc41e`) and pushed as `02bbe474`. Two things in it.

## The rebase

`git range-diff` reports the two reviewed commits replay unchanged except for one conflict resolution: #439 had edited the `charset_safe` docstring in `actions.rs`, and this branch moves the function to `fmt.rs`. Resolved as the move -- the docstring's content is superseded by the change below rather than carried across.

## `charset_safe` now marks elided characters instead of deleting them

Adopting the convention from #475 (@prasanna-anchorage's Solana renderer), on the same argument, before the sinks in this PR settle rather than after.

Deletion let distinct values collapse onto one rendering: `innocent\nTo: alice.near` and `innocentTo: alice.near` both reached the signer as `innocentTo: alice.near`. On a surface whose whole job is showing what is about to be signed, that is the wrong direction to fail. Deletion also presents what survived as though it were the whole value, so a truncated memo reads as complete.

Three notes on what changed beyond the substitution itself:

1. **An all-non-ASCII memo now renders as markers rather than being dropped.** `nonempty_filtered` used to discard it, on the reasoning that a blank labelled field reads as a deliberately empty memo. With markers the field carries information -- the sender attached text the signer cannot read -- and a signer who sees nothing cannot distinguish that from a memo never sent. A genuinely empty string is still dropped.
2. **Double quotes stay allowed**, which is where NEAR's set diverges from Solana's. `ft_transfer_call`'s `msg` carries real embedded JSON a signer needs to read literally; the Solana renderer emits its own `{key:value}` bracketing, where an unescaped quote is ambiguous. The docstring states the divergence and the reason.
3. **Twelve existing tests asserted the deletion semantics** and now assert the marked form. The tests themselves did not change shape -- only the expected strings -- so what they cover is unchanged. Two new cases: an all-non-ASCII string renders as markers, and two values differing only by an unrenderable character do not render identically.

`make test` and `make lint` are clean. If #475 changes shape in review -- a different marker, or collapsing runs of them -- matching it here is a one-line change.

## What this means for the two findings above

It closes half of each, and not the half that blocks.

Both findings are about `nonempty_filtered` collapsing "present but empty after sanitization" onto "absent", in a place where presence carries on-chain meaning: `mt_on_transfer` fires regardless of `msg` content, and `Some("")` still selects the `_transfer_call` branch. You named two inputs that reach the collapsed state -- an all-non-ASCII `msg`, and `Some("")`.

1. **The all-non-ASCII input no longer reaches it.** `charset_safe` marks rather than deletes, so that `msg` renders as markers and the field is emitted. That path is closed by construction rather than by a guard.
2. **`Some("")` still collapses.** An empty string sanitizes to an empty string, `nonempty_filtered` drops it, and `render_transfer` still renders byte-identically to the no-notification case. Finding 1 stands as written for that input, and finding 2 with it.

So the fix you suggest -- gate on `Option::is_some()` and give a present-but-empty value explicit fallback text, reserving `nonempty_filtered` for fields whose presence carries no independent meaning -- is still the right one, now needed only for the empty-string case. Doing that next unless you would rather it take a different shape, along with the tests you note are missing for `""` and for an all-non-ASCII `msg`.

shahan-khatchadourian-anchorage:
Both findings are addressed now, in `24d5c87c`. Taking your fix as suggested: gate on presence, give a present-but-empty value explicit fallback text, and reserve `nonempty_filtered` for fields whose presence carries no independent meaning.

`present_value` charset-filters a value whose presence changes what executes and substitutes `(empty)` when nothing renders. `render_transfer` pushes `Message` inside `if let Some(notification)` rather than on content, and `push_withdraw_call_details` matches `msg.as_deref()` directly. `memo` keeps `nonempty_filtered` on your reasoning -- it annotates the transfer and changes nothing the transfer does -- and both docstrings now state which applies where, so the next field added lands on the right side of the distinction rather than by whichever was copied.

Tests, all three from your review:

1. `transfer_with_an_empty_notification_message_still_renders_it` -- labels are `["To", "Amount", "Message"]`, against the `["To", "Amount"]` that `transfer_omits_notification_fields_when_absent` pins. The two now hold each other apart, which is the property that was missing rather than either test alone.
2. `transfer_with_an_unrenderable_notification_message_renders_markers` -- the other input you named, closed by the marker change and locked here.
3. `ft_withdraw_with_an_empty_message_still_renders_it` -- asserts both directions: `msg: ""` renders the field, and a withdraw with no `msg` still renders none, so the fix cannot drift into rendering `Message` for the plain non-callback form.

Worth confirming from the tests rather than the types: `"msg":""` does deserialize as a present notification rather than `None`, so the flattened-`Option` premise holds against the defuse types themselves.

`make test` and `make lint` are clean.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -121,6 +144,46 @@ fn token_amount_field(
...
+/// An empty `msg` still selects a contract-calling form on-chain: a
+/// `NotifyOnTransfer` invokes `mt_on_transfer` on the receiver whatever its
+/// `msg` holds, and a withdraw's `Some("")` takes the `_transfer_call` branch
+/// rather than the plain one. Dropping the field for want of text would render
+/// those byte-identically to the transfer that calls nothing, so the signer
+/// would approve a receiver callback with nothing on screen distinguishing it.
+fn present_value(text: &str) -> String {
+    let filtered = charset_safe(text);
+    if filtered.is_empty() {
+        EMPTY_VALUE.to_string()
```
pepe-anchor, line 186: should: present_value returns the "(empty)" marker both for a genuinely empty string and for the literal msg "(empty)", so those two distinct attacker-supplied values render identically, contradicting the distinctness rationale the rest of this file is built on.

pepe-anchor (dismissed):

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
pepe-anchor, line 1: should: The zero-delta guard in render_token_diff interpolates token_id into a ValidationError message without the charset_safe filtering every other echo of a TokenId gets in this same file, so a crafted NEP-171/NEP-245 asset id with an embedded newline (delta "0") reaches that message raw, the exact bug class this PR otherwise closes.

```suggestion
            return Err(VisualSignError::ValidationError(format!(
                "token_diff entry for {} has a zero delta",
                charset_safe(&token_id.to_string())
            )));
```

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Confirmed, and a bare deletion wouldn't have fixed it either -- both the genuinely-empty msg and the literal text "(empty)" pass through unchanged as plain ASCII, so any printable-ASCII marker is spoofable by a fully attacker-controlled string.

Fixed in 7a739915: EMPTY_VALUE now ends in a literal backslash. charset_safe elides every backslash a caller's text contains (it's the one character it never lets through unchanged), so no filtered msg can ever equal the marker -- the two cases are distinct by construction, not by luck of wording. Added a test asserting a msg that spells out "(empty)" renders as that literal text while a genuinely empty msg renders as the backslash-suffixed marker.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
shahan-khatchadourian-anchorage: Fixed in 7a739915, taking your suggestion as-is: the zero-delta ValidationError now charset_safes token_id before interpolating it, closing the one echo this PR's sanitization pass missed. Added a regression test with an embedded-newline asset id asserting the error message carries the sanitized form and no raw newline.

pepe-anchor (approved):

prasanna-anchorage (approved):
Approving on head `0e606a4d`. `make lint` clean; `make test` 42 suites / 2073 tests, 0 failures.

## Both findings fixed, verified by probe

| case | labels | Message text |
|---|---|---|
| transfer, `msg: ""` | To, Amount, **Message** | `(empty)\` |
| transfer, `msg` all-non-ASCII | To, Amount, **Message** | `??` |
| `ft_withdraw`, `msg: Some("")` | Token, To, Amount, **Message** | `(empty)\` |
| `ft_withdraw`, `msg: None` | Token, To, Amount | — |

Both sites gate on presence with the reasoning in a comment rather than left implicit: `notify_on_transfer` builds the `mt_on_transfer` promise whenever the notification exists whatever `msg` holds, and `Some("")` selects the `_transfer_call` form that invokes a receiver callback. `memo` correctly keeps `nonempty_filtered` — it annotates the transfer and changes nothing the transfer does, which is the distinction the two docstrings now state so the next field added lands on the right side of it rather than by whichever was copied.

**A correction on my own process:** my first probe passed `"notification":{"msg":""}` and showed no `Message` field at all, which read as the finding being unfixed. The notification is flattened into the intent, so the correct form is a top-level `"msg"` — my probe was constructing a transfer with no notification and asserting on the wrong thing. Worth saying because a probe that silently tests nothing is exactly how a false finding gets reported.

## pepe caught two things I missed

Both worth naming rather than glossing:

1. **`present_value` was spoofable**, returning `(empty)` for both a genuinely empty string and a `msg` literally spelling `(empty)`. That is a direct consequence of the fallback *I* suggested, and I should have seen it. The fix is better than a wording change: `EMPTY_VALUE` ends in a literal backslash, and `charset_safe` marks every backslash — the one character it never passes through — so no filtered value can equal the marker. Distinct by construction rather than by luck of wording. Verified: a `msg` of `(empty)\` renders as `(empty)?`.
2. **The zero-delta `ValidationError` echoed `token_id` raw** — the one unsanitized sink left in the file this PR exists to sanitize. Now `charset_safe`d, with a regression test asserting an embedded-newline asset id surfaces as `nep245:mt.near:x?To: attacker.near?Amount: 1000 USDC`.

## The elide-marker change, reviewed on its own merits

I flagged this as an unrequested scope expansion that alters every sanitized field, so I read it independently rather than as part of the fix. It earns its place. The collapse argument is real and it is the same one I made on the Solana side: `innocent\nTo: alice.near` and `innocentTo: alice.near` used to render identically, and now differ. On a surface whose whole job is showing what is about to be signed, silently dropping bytes is the wrong direction to fail.

`charset_safe`'s docstring is the part I would point someone at. It explains the `AccountId` exemption and why filtering those would be dead code; why `TokenId` is *not* such a value despite its account-id-shaped prefix (its `FromStr` parses only the contract half and takes the remainder verbatim); why substituting beats rejecting; and why NEAR keeps double quotes where the Solana renderer marks them, because `ft_transfer_call`'s `msg` legitimately carries embedded JSON while Solana's `{key:value}` bracketing makes an unescaped quote ambiguous. That last one is a real difference between the two chains reasoned out rather than papered over.

## Non-blocking

**1. `ELIDED` does not have the property `EMPTY_VALUE` was just given.** `?` is printable ASCII, so a caller's literal `?` passes through unchanged:

```
charset_safe("é") == "?"
charset_safe("?") == "?"     // equal
```

A marked character is indistinguishable from one the caller typed. This is genuinely weaker than the `(empty)` case rather than equivalent — nothing is *hidden*, since the smuggled newline still surfaces as a visible `?` — but distinctness-by-construction is the rationale pepe invoked and that `EMPTY_VALUE` now satisfies, and this is the same file. I do not think there is a clean fix: there is no un-typeable printable-ASCII character to reach for. So my ask is only that the docstring say so, since as written a reader could reasonably infer that a `?` on screen means "something was elided here".

**2. The marker's charset safety is real but unpinned.** `EMPTY_VALUE`'s trailing backslash is safe *because it is always the entire field value*: serialized it becomes `\\"`, and neither `\\` nor `\"` is in `FORBIDDEN_JSON_ESCAPES`. Probed both directions:

```
payload carrying the marker      -> validate_charset() Ok
field value EMPTY_VALUE + "until" -> validate_charset() Err
```

The second fails because a literal backslash serializes to `\\`, and `\\u` contains the forbidden `\u`. Both `present_value` call sites pass the result whole to `create_text_field`, so there is no bug today — but a later `format!("{} (truncated)", present_value(x))` landing before a `u`/`t`/`r`/`b`/`f`/`/` would reject the entire transaction, and nothing in the suite would catch it. One assertion that `validate_charset()` is `Ok` on a payload carrying the marker would pin the invariant the constant now depends on.

PR #462, opened 2026-08-07
fix(cli): render every field variant instead of "Field: Unknown"

shahan-khatchadourian-anchorage:
`format_field` handles `TextV2`, `PreviewLayout`, `AmountV2` and `AddressV2`. Every other `SignablePayloadField` variant — seven of the eleven in a default build, eight of twelve with the `diagnostics` feature — fell through to a catch-all printing `Field: Unknown`, and `common_label` mirrored the gap by returning `"Unknown"`, losing the label too.

## What this hides today

`Number` is the variant that matters in practice, and it is not rare. `visualsign-solana` emits it for:

- `jupiter_swap` — slippage and platform fee
- `compute_budget` — all four limits
- `system` — account space

All already on main. Those sit inside a `PreviewLayout`'s expanded list, and the nested render path goes through the same catch-all, so it hid them too. Someone triaging a swap saw every routing detail except the slippage they were agreeing to.

## The fix

The catch-all reads `label()` and `fallback_text()`, which are exhaustive over the enum. This closes the class rather than the one variant, including any variant added later. `common_label` had exactly one caller and is removed.

Display only — the wire payload was always correct. `Number`'s custom `Serialize` maps it to `amount_v2`, recovering the unit from `fallback_text`, so backends and wallets were unaffected. It still matters because the CLI is the tool reached for to triage a production payload.

## Testing

`output.rs` had no test module at all, which is why this went unnoticed — the Jupiter tests assert on decoded values, never on rendered output. This adds one covering `Number` at the top level, `Number` nested inside a `PreviewLayout` (the path the Solana fields actually take), and an assertion that no variant renders as `Unknown`. All three were verified to fail without the fix.

## Coordination

#439 modifies the same match block, adding a `Diagnostic` arm above the catch-all plus its own `mod tests` at the end of this file. When that stack rebases onto a main containing this PR, expect:

- a conflict in the `_` arm — keep this PR's accessor version, keep #439's `Diagnostic` arm above it
- #439's test module is renamed to `diagnostic_tests` so the two modules coexist; without that rename, keeping both hunks is `E0428`
- #439's arm comment ("a diagnostic printed as `Field: Unknown` tells the reader nothing") needs rewording, since diagnostics no longer print that way

### DIFF
--- a/src/parser/cli-core/src/output.rs
+++ b/src/parser/cli-core/src/output.rs
@@ -138,8 +138,17 @@ impl<'a> HumanReadableFormatter<'a> {
                     prefix, common.label, address_v2.address
                 )?;
             }
+            // Every remaining variant renders from the accessors, which are
+            // exhaustive over the enum. A variant added later prints its
+            // label and value rather than silently reading as "Unknown".
             _ => {
-                writeln!(writer, "{} Field: {}", prefix, common_label(field))?;
+                writeln!(
+                    writer,
+                    "{} {}: {}",
+                    prefix,
+                    field.label(),
+                    field.fallback_text()
+                )?;
             }
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


prasanna-anchorage (approved):
This closes a real gap — good catch that `Field: Unknown` was hiding 7-8 of 11-12 field variants including `Number`, and that `common_label` mirrored the same gap. Verified the new match arms cover the missing variants correctly.

One nit, non-blocking: the new catch-all `writeln!(writer, "{} {}: {}", prefix, field.label(), field.fallback_text())?` in `src/parser/cli-core/src/output.rs:145` uses positional format arguments where `prefix` is already an in-scope variable that could be captured inline (`"{prefix} ..."`), per this repo's "use inline format strings" convention. Small style fix, doesn't need to block.

Approving.

PR #128, opened 2025-12-16
[Feat] Solana Generic IDL Parser

prasanna-anchorage:
This builds on the work I did during/after ETHGlobal to add a Generic IDL parser that uses the same LayeredRegistry Abstraction. This is stacked on https://github.com/anchorageoss/visualsign-parser/pull/120 because it replicates the same functionality as I added for Ethereum but for Solana programs.

Now you can decode an arbitrary Anchor IDL using something like the following flow (Using JupiterLending as an example here)

This builds on the work I did in https://github.com/prasincs/solana-parser/tree/custom-idl-support and here's the [Open PR](https://github.com/tkhq/solana-parser/pull/20)

You can now download any arbitrary IDL with 
```
wget https://raw.githubusercontent.com/jup-ag/jupiter-lend/9eb4c372c6555c06e113b46e0a4beac47b5a3b81/target/idl/lending.json
```

and then use it in following way:

```
[...]

PR #465, opened 2026-08-10
Flag unregistered programs in Solana intermediate output

rituparna-mandal-anchor:
**Summary**

  - The Solana intermediate output now includes every inner/CPI instruction observed in a caller-supplied transaction simulation
  - Each simulated instruction and top-level instruction is tagged with a registeredSource (native/SPL, built-in, in-crate preset, caller-supplied, or unregistered)
  - Where our parser can decode a simulated instruction's arguments via a known IDL, those decoded args are included; where the RPC itself already decoded it (e.g. common SPL transfers), that decode is passed through.
  -  metadata_digest shifts for every Solana request because simulated_transaction_result isn't borsh-skipped, mirror decoders using solana chain metadata, should be updated.
  - This is purely additive and opt-in: existing consumers who don't supply a simulation result see no change

 **Why**

  Policy engines need to catch unrecognized programs, including ones that only become visible through pre-signing simulation. 
  
  **Test plan**

  - Full test suite passes
  - Real mainnet simulation fixture (Jupiter swap) verified to correctly classify all trust levels our decoder can produce, and to round-trip through our
    binary encoding intact


src/chain_parsers/visualsign-solana/src/cli_plugin.rs
```diff
@@ -101,6 +101,7 @@ pub(crate) fn create_chain_metadata(idl_json_mappings: &[String]) -> Option<Chai
             network_id: None,
             idl: None,
             idl_mappings: idl_mappings.into_iter().collect(),
+            simulate_transaction_result: None,
```
prasincs, line 104: Nit: I think it should be simulated_ because it maps with rest of code better

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -267,6 +287,19 @@ impl From<&parser::SolanaInstruction> for SolanaIntermediateInstruction {
...
 
+/// Built from a caller-supplied simulation result -- one call (top-level or
+/// inner/CPI) that a `simulateTransaction` call observed. `is_unregistered`
+/// is computed the same way as the static-decode path.
+impl From<&generated::parser::SimulatedInstruction> for SolanaSimulatedInstruction {
+    fn from(value: &generated::parser::SimulatedInstruction) -> Self {
+        Self {
+            program_key: value.program_key.clone(),
+            instruction_data_hex: value.instruction_data_hex.clone(),
+            is_unregistered: !crate::idl::builtin_programs::is_trusted_program(&value.program_key),
```
prasincs: We have to expand the concept of registered here because registered in our case doesn't just mean being in built in crates but also whether it's part of the signed IDLs in the request. I can't think of a better term to use here but essentially there's a materialized registry by this point to lookup in wherever that program came from.

nathan-margosian-anchor: > signed IDLs in the request
I thought IDLs were still being shipped as part of the OpenVSP binary for the time being. Has that already changed?

prasincs: We have had support for it for at least 8 months 

https://github.com/anchorageoss/visualsign-parser/pull/128

shahan-khatchadourian-anchorage (changes requested):
Solid refactor — flattening in `f2187da0` drops the positional-alignment assumption and reads much better than the nested version. Good test coverage for the size: native trust independent of IDL match, preset-only coverage via `preset_program_ids()`, a real on-chain router-CPI-ing-Jupiter case, borsh round-trip.

Ran a security pass too; nothing met the bar for a vulnerability. The provenance note below is legibility, not an exploit.

**CI fails on `f2187da0`** — `abi_metadata.rs:1029` missing `simulate_transaction_result`. It's in `visualsign-ethereum`'s test target, so `cargo build` is clean and a Solana-crate test run never compiles it; `--all-targets` catches it.

**`is_unregistered` only on the simulated list.** `c6ed0e8b` had it on `SolanaIntermediateInstruction` ("Downstream policy engines (e.g. the HSM) use this to decide whether to reject a transaction containing it"); `f2187da0` removed it. Deliberate, or collateral of the flattening? The static computation reads that instruction's own `program_key` and never depended on the positional alignment being removed.

As it stands the blob has two lists with opposite provenance — `instructions[]` from `message_hex`, `simulated_instructions[]` from caller metadata — and the only verdict is on the second. Consumers can't fill the gap: `is_trusted_program`'s set (`NATIVE_PROGRAM_NAMES` + `ProgramType::from_program_id` + `preset_program_ids()`) lives in the binary and grows with every preset. Restoring looks like a revert of the `c6ed0e8b` field hunk plus the static-side variants of the three tests renamed to `_via_simulation`.

**Schema version.** `c6ed0e8b` bumped to 2, `7036a85f` reverted. The appended length prefix ships on every emission, not just opted-in ones. A Rust mirror using `borsh::from_slice` or `from_reader` fails loudly — both reject trailing bytes — but a decoder calling `deserialize` directly, or a non-Rust mirror without a leftover-input check, reads a stale shape while its `schema_version == 1` assert still passes. That second case is what the version field exists to prevent, and it's the one we can't verify from here. Cheaper to bump now than after anything adopts 1.

**What "registered" resolves to.** Not reopening the thread, just one constraint that isn't visible from the diff: `extract_idl_mappings_with_signers` validates a signature only when one is present and accepts unsigned IDLs (`core/visualsign.rs:320`, flagged in-tree as a known gap). So the materialized registry holds signed and unsigned alike, and a literal registry lookup would admit unsigned caller IDLs — not the ask upthread, which was scoped to signed. Matching formulation: built-ins + presets + caller IDLs with a verified, allowlisted signature. `IdlRegistry::get_idl(&str)` takes the base58 string directly, and the registry is already built at the call site (`core/visualsign.rs:438`) two lines above the conversion — so the gate is the decision, not the plumbing. Ideally one shared helper for both paths.

**`metadata_digest`** changes for every Solana request, `include_intermediate_output` or not — borsh writes a tag byte for the `None`. Does anything downstream recompute it for Solana, and does proto regeneration ship in the same rollout? Fails closed, and the Ethereum `abi_mappings` change set the precedent, so this is deploy ordering. Upside: the digest now covers `simulate_transaction_result`, so a consumer can bind what it supplied.

**Smaller:**

- Caller-asserted data has the same shape and apparent authority as enclave-derived fields, under a boolean that reads as an enclave verdict. Provenance is in prose at both layers but not in the encoding; `idl_source` (`intermediate.rs:121`) is the existing idiom for tagging it.
- `program_key`/`instruction_data_hex` copied verbatim. The static sibling is `hex::encode` output from `solana_parser`, so one decode routine across both breaks on `0xDEADBEEF` or a non-base58 key. `Pubkey::from_str` / `visualsign::encodings::decode_hex` are the house pattern.
- `intermediate.rs:292` ("computed the same way as the static-decode path") contradicts `:65` ("the only place `is_unregistered` is computed").
- The inline `serde(default)` point still applies post-flattening: `SimulateTransactionResult.instructions` has no default, so `"simulateTransactionResult": {}` fails through the gateway.

src/chain_parsers/visualsign-solana/src/intermediate.rs
nathan-margosian-anchor: @prasincs @prasanna-anchorage Ok, I'm seeing this now -- thanks for linking the PR.

Basically the BE can send a map of IDLs (either signed or unsigned -- in this case we would want to always sign them, right?) and those get added to a registry. Then, when the parser is actually decoding, it first checks the IDLs stored as JSON files in the binary and if it doesn't find the program/instruction there, then it looks in the registry for the program/instruction.

It seems there is actually a third IDL source: `solana_parser`, which is included in the registry's implementation of `has_idl()` but is not a source in `get_idl()`, which seems to be a bug.

Based on this I have some questions:
1. How is it safe to accept IDLs from the backend? Even if they are signed and we can verify in the TEE that the payload has not been tampered with (via the signed value), how is it safe to trust that the BE is sending the right IDL? This probably didn't matter when we were just using these for visualization -- but now that we are enforcing SCX policies on the output, this doesn't feel safe.
    - Also, signing of the BE IDLs doesn't seem to be required currently so we don't even have that guarantee.
3. Should the solana_parser IDLs be included in `get_idl()`?
4. What is the north star for these sources? Will we want to maintain all three of these sources indefinitely or is there one source we are trying to converge on?

prasanna-anchorage: replied out of band

src/chain_parsers/visualsign-solana/src/cli_plugin.rs
rituparna-mandal-anchor: addressed, rename this var

shahan-khatchadourian-anchorage (changes requested):
All six from last round addressed: CI green, `registered_source` on the static instructions, schema version 2 with the original wording restored, provenance in the encoding instead of a bool. Taking the raw RPC response instead of a caller-decoded list is a better shape — the caller's decode is out of the trust path. `SolanaIdlParseError` and the mainnet Jupiter fixture are both good.

**Static decode output changed.** `extract_solana_intermediate_output` now merges 19 preset IDLs into `parse_transaction_with_idls` with `override_builtin: true` (`builtin_programs.rs:373`); on `origin/main` that argument is caller configs only. For any transaction touching a preset program, simulation or not: `idl_source` flips `"BuiltIn"` -> `"Custom"` and `idl_hash` changes for the four IDs that are also `solana_parser` builtins (Drift, Meteora, Orca, Jupiter v6), and `parsed_instruction_data` goes `None` -> populated for the rest. The schema bump covers shape, not these values; the PR body's "byte-for-byte unchanged" no longer holds.

`idl_source: "Custom"` now covers both our presets and wallet-supplied IDLs — the distinction `RegisteredSource` draws two fields over. Add a `Preset` source string, or document that `registered_source` is the field to read.

**Four silent paths to an empty simulation:** base64 `.ok()?`, JSON `.ok()?`, `construct_idl_records_map` err -> `Vec::new()`, `UiInstruction::Compiled` -> `continue`. All byte-identical to "no simulation sent", no log. A caller sending the JSON-RPC envelope instead of the bare `result`, or non-`jsonParsed` encoding, gets nothing — and a consumer gating on "no unregistered programs" sees nothing to object to. The `construct_idl_records_map` case drops the `Unregistered` classifications too, which need no IDL at all. Warn at each; better, reject a present-but-unparseable field.

**`response.value.err` dropped.** A reverted simulation's partial trace attaches as complete; an unregistered call never reached looks like one that isn't there.

**No bound on the payload.** Caller IDL JSON is capped at 1 MiB in the same file; this isn't. It compounds: per-instruction IDL resolution means N inner instructions against a 438 KB Drift parse, and `merge_preset_idl_configs` clones ~2 MB on every intermediate request (was `None` in the common case). Memoize per program, cache the parsed map.

**`ENTRIES` drifts from the derived set.** `preset_program_ids()` enumerates `available_visualizers()`; `ENTRIES` is hand-maintained, and `meteora_damm_v2.json` is already missing from it — reports `Preset` with nothing to decode it. Worth a test that every preset dir with an IDL JSON appears in `ENTRIES`, plus the same step in `solana-add-idl`.

**Smaller:**

- `index` never checked against the instruction count, though correlating back into `instructions[]` is its purpose.
- `stack_height` defaults to `1` (top-level) for entries that are all CPIs. `0` or `Option<u32>` reads as unknown.
- Proto comment names `parse_transaction_with_idls`; simulated instructions go through `parse_partially_decoded_instruction_idl`. It's also silent on envelope-vs-`result` and on `jsonParsed`, both of which fail silently when wrong.
- `merge_preset_idl_configs`'s doc describes a caller override winning over a preset; `from_idl_mappings` drops trusted IDs, so it can't happen.

**Carried over:** `metadata_digest` still shifts for every Solana request.

prasanna-anchorage:
@shahan-khatchadourian-anchorage re:

> `ENTRIES` drifts from the derived set. `preset_program_ids()` enumerates `available_visualizers()`; `ENTRIES` is hand-maintained, and `meteora_damm_v2.json` is already missing from it — reports `Preset` with nothing to decode it.

This one is stale as of a9e28a32. `meteora_damm_v2.json` is in `ENTRIES`, and the two lists match exactly today: 19 entries in `ENTRIES`, 19 preset directories shipping an IDL JSON, one-to-one.

The structural point holds, with a different example. `preset_program_ids()` derives from `available_visualizers()`, which includes presets that register a program ID but ship no IDL JSON. Today that is `swig_wallet` (`swigypWHEksbC64pWKwah1WTeh9JXwx8H1rJHLdbQMB`): not in `NATIVE_PROGRAM_NAMES`, not in `ENTRIES`, so `registered_source` returns `Preset` and there is nothing to decode it with. That is not drift that a rebase fixes — it is permanent, because the preset renders from code rather than an IDL.

So the question the test would be pinning down isn't only "did someone forget an `ENTRIES` line", it's what `Preset` is asserting. Right now it conflates "we vouch for this program ID" with "we can decode this instruction", and those are different sets. If a policy engine reads `Preset` as the latter, `swig_wallet` is already a counterexample. Worth either splitting the two or documenting which one `registered_source` answers — then the test enforces whichever you pick.

prasanna-anchorage:
Reviewed a9e28a32. Direction is right — taking the raw RPC response instead of a caller-decoded list keeps the caller's decode out of the trust path, and `RegisteredSource` over a bool is a real improvement. Two things I'd hold on, then some smaller items. I re-checked the round-2 review against the code and everything there still applies except the `ENTRIES` example, which I replied to separately.

## Blocking

**1. Static decode output changed, and the PR body says it didn't.**

`extract_solana_intermediate_output` (`intermediate.rs:660`) now calls `merge_preset_idl_configs(configs)`, layering 19 preset IDLs with `override_builtin: true` into `parse_transaction_with_idls`. On `main` that argument is caller configs only. For any transaction touching a preset program — simulation or not, opted in or not — `idl_source` flips `BuiltIn` -> `Custom`, `idl_hash` changes for the IDs that are also `solana_parser` builtins, and `parsed_instruction_data` goes `None` -> populated for the rest.

This is content that gets signed. The schema bump covers shape, not values, so a consumer pinned to v2 sees no signal. "static decode's output is completely unaffected either way" in the PR body needs to go, and this needs to be a deliberate, called-out change rather than a side effect of wiring up the simulation path.

Related: `idl_source: "Custom"` now covers both our presets and wallet-supplied IDLs, which is exactly the distinction `RegisteredSource` spends two fields drawing. Either add a `Preset` source string or document that `registered_source` is the field to read.

**2. Four silent paths to an empty simulation.**

All four produce output byte-identical to "no simulation was sent", none of them log:

- `base64 ... .ok()?` — `core/visualsign.rs:539`
- `serde_json::from_slice(...).ok()?` and `response.value.inner_instructions?` — `intermediate.rs:394-397`
- `construct_idl_records_map(...)` err -> `Vec::new()` — `intermediate.rs:414`
- `UiInstruction::Compiled` -> `continue` — `intermediate.rs:~424`

A caller who sends the full JSON-RPC envelope instead of the bare `result`, or who simulates without `jsonParsed` encoding, gets nothing back and no indication why. A policy engine gating on "no unregistered programs in the trace" sees an empty list and passes. Given the whole point of the feature is catching unrecognized programs, empty-because-we-couldn't-parse and empty-because-there-were-none must not be the same value.

The `construct_idl_records_map` case is the worst of the four: it drops the `Unregistered` classifications too, and those need no IDL at all to compute.

I'd reject a present-but-unparseable `simulated_transaction_result` outright rather than warn. The caller asked for something we couldn't deliver; failing the request is honest and failing closed is the right default on a signing path.

## Also worth fixing

**`solana_rpc_parsed_data.parsed_json` is not canonicalized.** `rpc_parsed.parsed.to_string()` (`intermediate.rs:471`) serializes a `serde_json::Value` straight into a borsh field that lands in the signed digest. This is the same `preserve_order` hole #414 fixed for `program_call_args_json`, and `canonicalize_value` / `canonicalize_map` are already in the same module. As written the module doc's determinism claim (lines 17-23) is no longer true for the new field, and a consumer that re-serializes this JSON computes a different string than we signed.

**The fixture doesn't show the feature working.** In `registered_source_classifications_from_jupiter_route_simulation`, the Jupiter v6 CPI classifies as `Preset` but comes back `parsed_instruction_data: None` with `DiscriminatorNotFound`. On the one real mainnet case in the PR, per-instruction IDL decode of a preset CPI does not produce args. That's the headline capability. Worth understanding whether it's the fixture, the discriminator lookup, or the CPI instruction shape before this ships.

**`instruction_data_hex: bs58::decode(...).unwrap_or_default()`** (`intermediate.rs:~433`) turns a malformed base58 payload into `""`, indistinguishable from an instruction that genuinely has no data. Everywhere else in this file that condition surfaces as `DiscriminatorNotFound`.

**The `UiParsedInstruction::Parsed` arm sets `accounts: Vec::new()` and `instruction_data_hex: String::new()`** (`intermediate.rs:473-476`). The account list is inside `parsed` for that variant, so a consumer reading the structured `accounts` field gets an empty list for exactly the common System/SPL calls — the ones most likely to matter to a transfer policy.

**Numbers on the cost.** `preset_idl_configs()` is `OnceLock`-cached but `merge_preset_idl_configs` clones it, and the preset JSON is 2.0 MB total. That clone happens twice per intermediate request (`intermediate.rs:410` and `:660`), plus a full `construct_idl_records_map` parse per request. Previously this was `None` in the common case. Memoize the parsed record map per program.

**`metadata_digest`** — confirmed it shifts for every Solana request: `parse.rs:101` hashes `borsh::to_vec(&chain_metadata)` and the new `optional string` writes a `None` tag byte whether or not the caller opts in. Fails closed, and the Ethereum `abi_mappings` change set the precedent, so this is deploy ordering rather than a defect — but proto regeneration and the parser rollout have to ship together.

**CI**: `ubuntu` is green. Only `check-cla-file` fails, which looks like it just needs a rebase now that #470 merged the signers-list entry.

prasanna-anchorage:
Correcting my own comment above: I framed `swig_wallet` as a counterexample, which made it sound like an anomaly. It isn't. None of the native/core presets ship an IDL either — `system`, `spl_token`, `token_2022`, `compute_budget`, `associated_token_account`, `stakepool` all have zero JSON files, and every ID in `NATIVE_PROGRAM_NAMES` is in the same position.

So "registered but with no IDL to decode it" is not an edge case to close, it's the majority of the trusted set, and by design — System and SPL Token decode through `solana_parser`'s native paths or the RPC's `jsonParsed` output, not through an IDL. Both tests in this PR already show it: the System transfer asserts `parsed_instruction_data.is_none()` with the comment "native decode path, not IDL", and the two SPL Token CPIs in the Jupiter fixture carry `solana_rpc_parsed_data` with `parsed_instruction_data: None`.

That kills the "split `Preset` into vouched vs decodable" half of what I suggested — the split doesn't exist along preset lines, it runs through `Native` too. What's left is the documentation half, and it matters more than I gave it credit for: `registered_source` answers "do we vouch for this program ID", never "can we decode this instruction", and a policy engine that reads it as the latter is wrong about System and SPL Token before it ever reaches `swig_wallet`.

One concrete consequence for the simulated path. A `Native` program's args come only from the RPC's `jsonParsed` decode, since there's no IDL to fall back on. If the caller simulates with any other encoding, those instructions arrive as `PartiallyDecoded`, find no IDL record, and land as `(None, None)` — `registered_source: Native`, no args, no `idl_parse_error`, nothing to distinguish it from a program we decoded and found nothing in. That's the same silent-empty problem as the four paths in my other comment, but it survives even when everything upstream worked, and it hits the most common instructions in any transaction.

pepe-anchor (commented):
Reviewed with [Claude](https://claude.com/claude-code) on behalf of @pepe-anchor. 8 suggestion(s) of 10 drafted; 0 marked as blocking. _Treat findings as suggestions, not directives._

Round 4 at `b16cae66`. That commit is `a9e28a32` plus a rustfmt merge of main, so the code has not changed since round 3 and all eleven findings there are still open. I re-checked each against the tree rather than assuming. Not re-posting them inline, treat that review as live.

Two of them outrank most of what is below, so flagging them here: the `override_builtin: true` preset merge still changes `idl_source` and `idl_hash` on the static path for every transaction touching a preset (the PR body still says otherwise), and the four silent paths to an empty simulation still fail open on the exact signal this PR exists to provide.

New this round, mostly the rollout side, verified against the monorepo at `a2517f7ed112`:

- The schema bump has no tripwire. The HSM mirror is still at version 1, and both downstream tests are fixture- or pin-based, so neither fires on merge. Details on the constant.
- The `jsonParsed` arm is wholly caller-authored. Line 462, and the one I would fix first.
- The borsh round-trip test never encodes the new types, and two `RegisteredSource` variants have no coverage at all.

Shape suggestion, take it or leave it. The proto field and the schema bump have different blast radii and could ship separately. The proto field is safe alone: `ChainMetadata.metadata` is a oneof, so Ethereum's bytes are byte-identical before and after, the Solana variant grows seven bytes but nothing constructs it, and there is no copy of parser.proto in the monorepo. The schema bump wants firmware first, ideally replacing the single constant with a supported set so a decoder tolerates N and N+1.

pepe-anchor (commented):
Inline notes for the round-4 review above.

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,46 +323,309 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+
+        for ui_instruction in entry.instructions {
+            let UiInstruction::Parsed(parsed) = ui_instruction else {
+                continue;
+            };
+
+            match parsed {
+                UiParsedInstruction::PartiallyDecoded(decoded) => {
+                    let accounts = decoded.accounts;
+                    let instruction_data_hex = bs58::decode(&decoded.data)
```
pepe-anchor: `bs58::decode(...).unwrap_or_default()` turns a decode failure into an empty `instruction_data_hex`.

When the program has an `IdlRecord`, the sibling call records `DiscriminatorNotFound("instruction data is not valid base58")` at `:522`, so the failure is visible. When it does not, `parse_partially_decoded_instruction_idl` early-returns `(None, None)` at `:507` before the data is ever looked at, and the instruction lands with empty data and no error. That covers every `Native` and `Unregistered` program, since neither ships an IDL, so it is the common case rather than an edge one.

Extending Prasanna's bullet on this line rather than raising it fresh. Either record the failure here or keep the raw base58 string.


---

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,46 +323,309 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+                        program_key: decoded.program_id,
+                        accounts,
+                        instruction_data_hex,
+                        registered_source,
+                        parsed_instruction_data,
+                        solana_rpc_parsed_data: None,
+                        idl_parse_error,
+                    });
+                }
+                UiParsedInstruction::Parsed(rpc_parsed) => {
```
pepe-anchor, line 544: This arm keeps only the caller's `program` and `parsed` JSON. `accounts` is `Vec::new()`, `instruction_data_hex` is `String::new()`, `parsed_instruction_data` is `None`, so nothing survives that a consumer could check the claim against. The sibling `PartiallyDecoded` arm re-derives from the base58 data; this one derives nothing.

So any instruction the caller marks `jsonParsed` is wholly caller-authored, while `registered_source` still reports whatever `program_id` maps to. A caller can send `{"program":"system","programId":"1111...","parsed":{"type":"transfer","info":{"lamports":1}}}` and the blob shows a benign System transfer, `registered_source: Native`, no bytes to re-derive from. Keep `accounts` and the base58 data at minimum, or run this arm through the same IDL re-derivation and treat the RPC decode as a hint.

---

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -64,6 +70,40 @@ pub struct SolanaIntermediateInstruction {
...
+    ThirdParty,
+    /// Matched only via caller-provided `idl_mappings`.
+    CallerSupplied,
+    /// Matched none of the above; nothing was found at all.
+    Unregistered,
+}
+
+#[derive(BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
+pub struct SolanaSimulatedInstruction {
+    pub index: u32,
```
pepe-anchor, line 132: `index` is the outer instruction's index, so every entry in a CPI group shares it. `:424` sets `outer_index` once per `inner_instructions` entry and every instruction pushed under it copies that value. With no doc comment, a consumer keying a dedup or a lookup on it will silently collide entries. `outer_instruction_index`, plus a line saying it is a grouping key rather than a unique one, would fix it.


---

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -64,6 +70,40 @@ pub struct SolanaIntermediateInstruction {
...
+    pub stack_height: u32,
+    pub program_key: String,
+    pub accounts: Vec<String>,
+    pub instruction_data_hex: String,
+    pub registered_source: RegisteredSource,
+    pub parsed_instruction_data: Option<SolanaParsedInstructionDataIo>,
```
pepe-anchor, line 144: These two are always mutually exclusive in practice, but nothing in the type says so. A consumer that checks only `parsed_instruction_data`, the natural move since it matches the sibling `SolanaIntermediateInstruction`, silently misses every RPC-recognised instruction. One enum instead of two `Option`s makes that a compile error. Low priority though: out-of-tree mirrors have to track this shape, so a doc comment may be the better trade.

---

src/chain_parsers/visualsign-solana/src/idl/builtin_programs.rs
```diff
@@ -210,6 +217,33 @@ pub fn is_trusted_program(program_id_str: &str) -> bool {
     canonical_name(program_id_str).is_some() || preset_program_ids().contains(program_id_str)
 }
 
+/// Where `program_id_str` is registered, if at all -- see
+/// [`crate::intermediate::RegisteredSource`]. `caller_idl_program_ids` is the
+/// set of program IDs the caller supplied a custom IDL for (the keys of
+/// `IdlRegistry::get_all_configs()`), used only to detect `CallerSupplied`;
+/// pass an empty map if unavailable.
+pub fn registered_source(
```
pepe-anchor, line 225: `ThirdParty` and `CallerSupplied` have no test coverage. Each has exactly one construction site, `:239` and `:241`, and neither appears in any test module. The tests only ever assert `Native`, `Preset` and `Unregistered`.

Those two are the variants that decide whether policy is looking at a program we compiled in or one the caller described to us. A case for each, a `ProgramType`-only program and one present only in `idl_mappings`, would also pin down Prasanna's question about what `Preset` asserts.


---

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -54,6 +59,7 @@ pub struct SolanaIntermediateOutput {
     pub spl_transfers: Vec<SplTransfer>,
     pub recent_blockhash: String,
     pub address_table_lookups: Vec<SolanaAddressTableLookup>,
+    pub simulated_instructions: Vec<SolanaSimulatedInstruction>,
```
pepe-anchor, line 63: The bump to schema 2 exists to protect the new bytes, and two of the new types never see borsh in a test.

`SolanaRpcParsedInstructionDataIo` has no test occurrence at all, it appears only at `:105`, `:168` and `:478`. `SolanaIdlParseError` is asserted on in the Jupiter test, but that path never encodes, and the two populated round-trips in `core/visualsign.rs` reach `RegisteredSource` without ever reaching an error variant. `intermediate_output_round_trip_is_deterministic` builds `instructions: vec![]`, so it encodes neither.

`SolanaIdlParseError` is the half worth covering. It is a borsh enum with named-struct and tuple variants, materially more work for a hand-written mirror decoder than the flat structs around it, and `DiscriminatorNotFound` is the only variant any test touches. Worth extending the round-trip test to a populated instruction and simulated instruction carrying a non-trivial error variant.


---

src/chain_parsers/visualsign-solana/src/cli_plugin.rs
```diff
@@ -101,6 +101,7 @@ pub(crate) fn create_chain_metadata(idl_json_mappings: &[String]) -> Option<Chai
             network_id: None,
             idl: None,
             idl_mappings: idl_mappings.into_iter().collect(),
+            simulated_transaction_result: None,
```
pepe-anchor: `SolanaArgs` has only `idl_json_mappings` and `create_chain_metadata` hardcodes `simulated_transaction_result: None`, so the CLI cannot exercise the feature while the gRPC path can.

That bites twice: `parser_cli decode` is the documented way to inspect a transaction locally, and it is the practical way to produce a fixture for an out-of-tree decoder. As it stands there is no way to emit a v2 blob with populated simulated instructions without going through the service. A `--simulated-tx-result` flag threaded like `idl_json_mappings` would do it.


---

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -38,7 +43,7 @@ use crate::idl::IdlRegistry;
 /// Version of the `SolanaIntermediateOutput` Borsh schema. Bump on ANY change
 /// to the shape below. Mirrored decoders assert this value, so a bump makes a
 /// schema drift fail loudly instead of silently misparsing.
-pub const SOLANA_INTERMEDIATE_SCHEMA_VERSION: u16 = 1;
+pub const SOLANA_INTERMEDIATE_SCHEMA_VERSION: u16 = 2;
```
pepe-anchor, line 47: Two things on the bump itself.

The doc comment says "Mirrored decoders assert this value" without saying where any of them live, and CONTRIBUTING does not either. Naming them, or at least stating that out-of-tree hand-written decoders exist, is the cheapest fix here. Right now nothing in this repo tells a contributor bumping this constant who else has to move.

Second, the guarantee this comment claims does not hold for a decoder that reads the whole body before it looks at the version. Borsh is positional, so a real v2 blob misaligns on the extra per-instruction fields and dies on a short read or a trailing-bytes check well before any `schema_version` comparison runs. It fails closed either way, so this is diagnosability rather than safety, but the operator gets an alignment error instead of "schema_version 2 is not supported". A mirror only gets the clean error if it reads the `u16` first and dispatches on it, which is also what any decoder accepting more than one version needs. Worth stating that requirement next to the constant: the field being first is what makes it possible, and nothing currently says why that ordering matters.


---

prasanna-anchorage:
Drive-by perf note from benchmarking the intermediate path on #472 — not a blocker, but it lands on a hot path.

`merge_preset_idl_configs` does `preset_idl_configs().clone()` ([`builtin_programs.rs:383`](https://github.com/anchorageoss/visualsign-parser/blob/rituparnamandal/prof-354-is-unregistered-tag/src/chain_parsers/visualsign-solana/src/idl/builtin_programs.rs#L383)), which deep-clones all 19 bundled IDL JSON strings. Those total **~2.0 MB** on disk. It's called from two places — `intermediate.rs:410` (simulation) and `intermediate.rs:660` — and the second is the unconditional static path, so every intermediate request pays it whether or not a simulation was supplied.

For scale: on #472 I benchmarked `convert_with_intermediate` on real fixtures at 16–31 µs for token ops and 2.2–2.4 ms for Orca/Jupiter, after getting the intermediate's overhead down to ~0% over render-only. A 2 MB clone per request is large relative to that.

**`Arc` doesn't quite solve it**, which was my first instinct too. `parse_transaction_with_idls` takes `custom_idls: Option<HashMap<String, CustomIdlConfig>>` **by value** (solana-parser rev `2146929`, `src/solana/parser.rs:85-88`), so an owned map has to be materialised at the call site no matter how the cache is stored.

The bigger win is one variant over. `CustomIdl` is:

```rust
pub enum CustomIdl {
    /// A pre-parsed IDL struct (avoids re-parsing)
    Parsed(Idl),
    /// An IDL as a JSON string (will be parsed)
    Json(String),
}
```

`preset_idl_configs()` currently stores `CustomIdl::Json(...)`, so each request doesn't just clone 2 MB of text — it **re-parses** it downstream. Holding `CustomIdl::Parsed(Idl)` in the `OnceLock` instead parses the 19 IDLs once at first use, and the per-request clone then copies parsed structures rather than re-parsing JSON. That's the change I'd suggest here.

Killing the remaining clone entirely would need `parse_transaction_with_idls` to accept `&HashMap`. Since solana-parser is ours and pinned by rev, that seems worth a follow-up issue rather than anything blocking this PR.

Happy to send a patch for the `Parsed` change if useful — we're planning to stack on this branch, so I'd rather flag it now than land on top of it.

prasanna-anchorage:
One more from re-reading the simulation path — I think this one is worth holding on, separate from the perf note above.

## Reverted simulations are consumed as if they succeeded

`parse_and_decode_simulated_instructions` (`intermediate.rs:386`) reads the RPC response like this:

```rust
let response: Response<RpcSimulateTransactionResult> = serde_json::from_slice(raw_json).ok()?;
let inner_instructions = response.value.inner_instructions?;
Some(decode_inner_instructions(inner_instructions, idl_registry))
```

`response.value.err` is never read. It exists on the type — `RpcSimulateTransactionResult { err: Option<TransactionError>, .. }` (solana-rpc-client-types 2.3.13, `src/response.rs:400`) — and `simulateTransaction` populates it while *still* returning whatever `innerInstructions` executed before the failure.

So a simulation that reverted contributes its partial CPI trace to `intermediate_output`, indistinguishable from one that ran to completion.

That matters more here than it would elsewhere, because of what this field is for. `intermediate_output` is borsh-appended into the signed digest and consumed by a policy engine, and the stated goal of this PR is catching unrecognized programs. A reverted simulation's trace stops at the revert point, so the program set it reports is a **subset** of what real execution would touch — the policy engine can conclude "only known programs here" in exactly the case where it shouldn't. It's a false negative in the control itself, not just missing data.

Suggested shape: treat `err.is_some()` as a distinct outcome rather than a success. Either refuse to attach simulated instructions at all, or carry the error explicitly in the schema so the consumer can decide — the latter is probably more useful, and it's a natural fit next to `SolanaIdlParseError`.

## Two smaller cases with the same shape

Both turn "we couldn't read this" into "there was nothing here":

- `UiInstruction::Compiled => continue` (`intermediate.rs:~428`) silently drops instructions, so a simulation sent without `jsonParsed` encoding yields an empty-but-present list that looks identical to "no CPIs".
- The three `.ok()?` / `else { return Vec::new() }` paths — base64 decode in `extract_raw_simulated_instructions`, `serde_json::from_slice`, and `construct_idl_records_map` — make a malformed simulation indistinguishable from no simulation supplied.

For a field that feeds policy, "absent" and "unparseable" probably shouldn't be the same value. Worth at least a `tracing::warn!` on each, and ideally a schema-level signal for the encoding mismatch.

Happy to be wrong on the severity of the first one if the caller contract already guarantees a successful simulation — but I couldn't find that stated anywhere, and the proto comment on `simulated_transaction_result` doesn't mention it.

src/Cargo.lock
```diff
@@ -12350,7 +12350,7 @@ dependencies = [
 [[package]]
 name = "solana_parser"
 version = "0.1.0"
-source = "git+https://github.com/anchorageoss/solana-parser.git?rev=2146929#21469298de95c213c98e5483da6ebf37aa987066"
+source = "git+https://github.com/anchorageoss/solana-parser.git?rev=c827dc3#c827dc3fc7dfa97f40641ba13ffa5f52a7d9afc4"
```
prasanna-anchorage: This pins `solana_parser` to `c827dc3`, which is the tip of `anchorageoss/solana-parser`'s `rituparnamandal/idl-record-entry-point` — a branch with **no open PR** (the only open PRs on that repo are #8, #5 and #1).

Could we get one opened before this merges? I reviewed the two commits it adds and they're good — I'd approve:

- `76100ef` adds `parse_transaction_with_idl_records`, with `parse_transaction_with_idls` delegating to it, so existing callers are untouched. This is a better fix for the per-request IDL cost than the `CustomIdl::Parsed` swap I suggested earlier: it hoists the whole parse-and-rehash step into a cache the caller owns, not just the JSON parse.
- `c827dc3` adds `IdlSource::Preset` / `IdlRecord::is_preset` so application-shipped IDLs report distinctly instead of masquerading as `Custom`. `custom_idl_source()` is applied at both existing `Custom` return sites and doesn't change which IDL is selected, so it's reporting-only — which is what makes it a clean answer to the "static decode output changed" objection.

I also checked the two things that could have bitten and neither does: inserting `Preset` before `Custom` shifts the borsh discriminant, but we store `idl_source` as a `String` via `idl_source_string()` (with a test pinning the values), so nothing on the wire moves; and all three arms are handled in `idl_parser.rs`, `main.rs` and our `idl_source_string`.

The reason I'd rather not merge on the branch pin: `idl-record-entry-point` is itself stacked on `shahankhatch/add-fuzz-targets-tkhq-base`, which is also unmerged, and neither is on any `main` — `anchorageoss/main` is currently 0 ahead / 9 behind `tkhq/main`. So this would put us three levels deep off-trunk, pinned to a branch tip that can be force-pushed or renamed out from under us with no review record of what we depend on.

Two small things worth folding into that PR when it's opened:

- `#[allow(dead_code)] // Public API, superseded internally by new_with_idl_records` on `SolanaTransaction::new_with_idls` — a `pub` method in a lib crate shouldn't trip `dead_code`. If it genuinely does, something about the export path is off and the `allow` hides it rather than fixing it.
- In `idl_parser.rs`, the new `/// Which of the two non-built-in sources...` doc line landed between `resolve_idl_for_record`'s existing doc and its signature, so it now documents the wrong item.

Related: I have `all_accounts` (positional account view, for ALT-backed instructions) out for review upstream at tkhq/solana-parser#30 and anchorageoss/solana-parser#8, both based on `2146929`. `c827dc3` descends from `2146929` cleanly, so once this lands somewhere durable I'll rebase onto it — no conflict either way, just flagging so we don't end up with two divergent pins.

prasanna-anchorage (dismissed):
Approving. I re-read the code at `dec529e4` rather than taking the commit messages at their word, and both of the things I raised are properly addressed — one of them better than what I proposed.

## Verified: reverted simulations no longer consumed as successes

This was the one I'd have held on. `parse_and_decode_simulated_instructions` now checks `response.value.err` and returns `(Vec::new(), Some(SolanaSimulationError::SimulationFailed))`, dropping the partial trace rather than attaching it as if the simulation completed. Malformed JSON returns `InvalidJson` instead of an indistinguishable empty result.

Crucially the outcome is carried **in the schema** rather than only logged, so a policy engine can tell "no simulation supplied" from "simulation reverted" from "unparseable input". That was the option I flagged as more useful than refusing outright, and it's what closes the false-negative surface: a reverted trace stops at the revert point, so its program set is a subset of what real execution touches — precisely when a policy engine must not conclude "only known programs here".

## Verified: the per-request IDL cost is gone

`merge_preset_idl_configs` — which deep-cloned ~2.0 MB of IDL JSON on every intermediate request, on the unconditional static path — is deleted. `lookup_idl_record` now layers caller → preset → builtin against process-wide `OnceLock` caches, borrowing rather than cloning, and the doc comment records why.

The upstream half (`parse_transaction_with_idl_records`) is a better fix than the `CustomIdl::Parsed` swap I suggested: it hoists the entire parse-and-rehash step into a cache the caller owns, not just the JSON parse.

## Verified: static decode output no longer changes

`extract_solana_intermediate_output` passes `caller_idl_records(configs)` only — presets are not injected into the static decode as `Custom` any more. The `IdlSource::Preset` variant reports application-shipped IDLs distinctly instead, so `idl_source` no longer flips `BuiltIn -> Custom` and `idl_hash` no longer moves for the four ids that are also `solana_parser` builtins. That addresses @shahan-khatchadourian-anchorage's blocking objection without giving up the decoded-args coverage the PR is for.

I also checked the two things that could have broken silently and neither does: inserting `Preset` before `Custom` shifts the borsh discriminant, but `idl_source` is stored as a `String` via `idl_source_string()` with `idl_source_string_is_stable` pinning the values, so nothing on the wire moves; and all three arms are handled everywhere they're matched.

## Two notes, neither blocking this approval

**@shahan-khatchadourian-anchorage's `CHANGES_REQUESTED` from 26 Aug still stands**, and my approval does not clear it — the PR stays blocked until he re-reviews. I believe his objection is resolved (above), but that's his call to make, not mine.

**The `solana_parser` pin still needs somewhere durable to point.** See my comment on `src/Cargo.lock` — `c827dc3` is a branch tip with no PR, itself stacked on another unmerged branch. I'd treat opening that PR as a merge precondition rather than a code objection; the code in it is good and I'd approve it as-is.

prasanna-anchorage:
@shahan-khatchadourian-anchorage re-review request sent — your `CHANGES_REQUESTED` from 26 Aug is the only thing blocking this now.

Your objection was that `extract_solana_intermediate_output` merged 19 preset IDLs into `parse_transaction_with_idls` with `override_builtin: true`, so for any transaction touching a preset program `idl_source` flipped `BuiltIn -> Custom`, `idl_hash` moved for the four ids that are also `solana_parser` builtins, and `parsed_instruction_data` went `None -> populated` — despite the PR body claiming static decode was unaffected.

Three commits since (`7373a82d`, `f0d4c51e`, `dec529e4`). The relevant change: `merge_preset_idl_configs` is gone, and `extract_solana_intermediate_output` now passes `caller_idl_records(configs)` only. Presets are no longer injected into the static decode at all. The pinned `solana_parser` bump adds an `IdlSource::Preset` variant so application-shipped IDLs report distinctly rather than masquerading as `Custom` — reporting-only, `custom_idl_source()` doesn't change which IDL is selected.

So the flip you identified shouldn't happen any more, and the coverage the PR is for is retained via the new variant rather than by overriding builtins.

I've approved on the strength of that plus the simulation-error handling and the IDL-caching fix, and I checked the borsh-discriminant and exhaustive-match hazards from the new enum variant — both clean, details in my review. But you found the original problem and it's your call whether the fix actually settles it.

One open item that isn't yours to resolve: the `solana_parser` pin is a branch tip with no PR (comment on `src/Cargo.lock`). I'd treat that as a merge precondition.

pepe-anchor (commented):
[AI Review]

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,51 +367,441 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+fn build_idl_record_map(
+    caller_records: &BTreeMap<String, solana_parser::solana::structs::IdlRecord>,
+) -> std::collections::HashMap<String, solana_parser::solana::structs::IdlRecord> {
+    let mut records: std::collections::HashMap<_, _> = builtin_idl_records()
+        .iter()
+        .map(|(k, v)| (k.clone(), v.clone()))
+        .collect();
+    for (program_id, record) in crate::idl::builtin_programs::preset_idl_records() {
+        records.insert(program_id.clone(), record.clone());
+    }
```
pepe-anchor, line 640: must: This loop inserts every preset record **over** the `solana_parser` builtin record for the same program, and `preset_idl_records()` sets `program_type: None` (builtin_programs.rs:415), which removes the only fallback `resolve_idl_for_record` has (upstream idl_parser.rs:183-201 uses the builtin solely via `program_type`); the result is a decode regression on the top-level static path for programs that are both a preset and a builtin. I diffed the instruction sets case-insensitively: Meteora DLMM `LBUZKhRx…` loses **34** instructions (preset 31 vs upstream meteora.json 64, including `initializeLbPair`, `initializeBinArray`, `fundReward`, `goToABin`), Drift `dRiftyHA…` loses **20** (including `settleLp`, `addPerpLpShares`, `removePerpLpShares`, `transferProtocolIfShares`), Jupiter Agg V6 `JUP6Lkb…` loses **2** (`createOpenOrders`, `createProgramOpenOrders`). Every one of those decoded into `parsed_instruction_data` before this PR and now returns `DiscriminatorNotFound`. Two further wire-visible changes for these programs, unmentioned in the PR body: `idl_source` flips `BuiltIn` to `Preset`, and instruction/account/arg names flip camelCase to snake_case because the presets are Anchor 0.30 IDLs. Minimal fix: skip the preset record when `ProgramType::from_program_id(program_id).is_some()`, so preset records only cover programs that previously had no builtin at all.

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,51 +367,441 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+                    "unexpected compiled inner instruction in simulated_transaction_result; \
+                     simulateTransaction does not return this shape"
+                );
+                return (Vec::new(), Some(SolanaSimulationError::CompiledInstruction));
+            };
+
+            match parsed {
+                UiParsedInstruction::PartiallyDecoded(decoded) => {
+                    let accounts = decoded.accounts;
+                    let Ok(data) = bs58::decode(&decoded.data).into_vec() else {
```
pepe-anchor, line 510: must: `bs58` is an optional dependency gated behind `cli-plugin`, but this line and line 692 now use it in plain library code, so `cargo check -p visualsign-solana --no-default-features` fails with `use of unresolved module or unlinked crate bs58`; fix by moving `bs58` out of `optional`/`cli-plugin` into a plain `[dependencies]` entry. Verified by running it against HEAD (53114f67): three errors, at intermediate.rs:510 and intermediate.rs:692. `make test` does not catch this because its `cargo test -p visualsign-solana --no-default-features --lib` target pulls `bs58` in from `[dev-dependencies]` (Cargo.toml:49), which a non-test build does not get.

proto/parser/parser.proto
```diff
@@ -139,6 +139,12 @@ message SolanaMetadata {
   // Map of program_id (base58 string) to IDL definitions
   // Allows wallet to provide multiple IDLs, one per program
   map<string, Idl> idl_mappings = 3;
+  // Raw simulateTransaction RPC response bytes (innerInstructions section),
+  // base64-encoded (matching unsigned_payload's convention), sent with no
+  // caller-side decode/reshape. The parser runs this through the same
+  // static-decode path (parse_transaction_with_idls) used for
+  // unsigned_payload.
+  optional string simulated_transaction_result = 5;
```
pepe-anchor, line 147: must: `SolanaMetadata` carries `#[derive(borsh::BorshSerialize)]` (src/generated/src/generated/parser.rs:230) and `metadata_digest` is `sha_256(borsh::to_vec(chain_metadata))` (src/parser/app/src/routes/parse.rs:92-101), so appending this optional field adds a `0x00` tag byte and changes the digest for every Solana parse request even when no simulation is sent, contradicting the PR body's "existing consumers who don't supply a simulation result see no change"; either say so explicitly and coordinate with anchorage#241185, or keep the field off the borsh-hashed message the way `intermediate_output` is `#[borsh(skip)]`-ed out of the payload digest for exactly this reason.

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,51 +367,441 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+                        &rpc_parsed.program_id,
+                        configs,
+                    );
+
+                    simulated_instructions.push(SolanaSimulatedInstruction {
+                        index: outer_index,
+                        stack_height: rpc_parsed.stack_height.unwrap_or(0),
+                        program_key: rpc_parsed.program_id,
+                        accounts: Vec::new(),
+                        instruction_data_hex: String::new(),
```
pepe-anchor, line 557: must: For every program the RPC jsonParses (System, SPL Token, Stake, Vote, ATA, ComputeBudget) this emits `accounts: []` and `instruction_data_hex: ""`, which removes the input to the discriminator-level SCX allowlist that commit 70fb0bed named as the authoritative allow/reject check ("instruction_data_hex (preserved on every instruction regardless of source)"); the fixture test at intermediate.rs:1082-1090 exercises exactly this branch for two Token-program calls and asserts nothing about either field. Consequence for PROF-354: an inner `spl-token approve`, `setAuthority` or `closeAccount` arrives tagged `RegisteredSource::Native` with no bytes to check a discriminator against, and the only remaining content is `parsed_json`, which is caller-supplied and unverified. Keep the bytes: `UiPartiallyDecodedInstruction` and the parsed shape both come from the same RPC response, so at minimum carry `rpc_parsed.parsed["info"]`'s account keys, or reject the `Parsed` variant the way the compiled variant is rejected at line 499.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -482,22 +485,78 @@ impl VisualSignConverter<SolanaTransactionWrapper> for SolanaVisualSignConverter
...
-        Ok(output) => match borsh::to_vec(&output) {
-            Ok(bytes) => Some(bytes),
-            Err(err) => {
-                tracing::warn!("Failed to borsh-encode Solana intermediate output: {err}");
-                None
+        Ok(mut output) => {
+            // Simulation results are independent of static decode.
+            if let Some((instructions, simulation_error)) = raw_simulated_instructions {
+                output.simulated_instructions = instructions;
+                output.simulation_error = simulation_error;
```
pepe-anchor, line 498: must: `simulated_transaction_result` is caller-supplied and the parser never runs the simulation, yet nothing here checks that it belongs to this message: `entry.index` (intermediate.rs:496) is copied into `SolanaSimulatedInstruction.index` without ever being compared against `output.instructions.len()`, so a caller can attach a trace for a different transaction, or index 99 against a one-instruction message, and it is emitted as authoritative. Since PROF-354 wants the HSM to reject on unregistered inner calls, the same caller who controls the trace controls whether any unregistered program appears in it at all. Minimum fix here: reject the trace with a new `SolanaSimulationError` when any `entry.index >= output.instructions.len()`, and state in the proto comment that the field is untrusted input whose absence must itself be a hard failure downstream.

src/chain_parsers/visualsign-solana/src/idl/builtin_programs.rs
```diff
@@ -241,6 +275,157 @@ pub fn is_reserved_canonical_name(name: &str) -> bool {
...
+/// The IDL-backed in-crate presets' Anchor IDL JSON, keyed by program ID
+/// included here so that decode path can use these idls.
+fn preset_idl_configs()
+-> &'static std::collections::BTreeMap<String, solana_parser::CustomIdlConfig> {
+    use solana_parser::{CustomIdl, CustomIdlConfig};
+
+    static PRESET_IDL_CONFIGS: OnceLock<std::collections::BTreeMap<String, CustomIdlConfig>> =
+        OnceLock::new();
+    PRESET_IDL_CONFIGS.get_or_init(|| {
+        const ENTRIES: &[(&str, &str)] = &[
```
pepe-anchor, line 287: should: `preset_program_ids()` (line 190) is derived automatically from `available_visualizers()` via build.rs, but this `ENTRIES` list is hand-written and no test asserts the two agree, and they already disagree today: `swig_wallet` (`swigypWHEksbC64pWKwah1WTeh9JXwx8H1rJHLdbQMB`) is a registered preset that is neither native nor a `ProgramType`, so `registered_source()` tags it `RegisteredSource::Preset` while `preset_idl_records()` holds no record for it, making `parse_partially_decoded_instruction_idl` return `(None, None)`, which this PR's own docs define as "nothing was available to attempt against". The label promises a decode that cannot happen. It also gets worse over time: the repo's documented workflow, `.claude/skills/solana-add-idl/SKILL.md`, scaffolds `config.rs` and `mod.rs` and never mentions `builtin_programs.rs`, so the next preset lands the same way. Add a test asserting every `preset_program_ids()` entry whose preset directory ships a `.json` has an `ENTRIES` row.

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -122,6 +205,41 @@ pub struct SolanaParsedInstructionDataIo {
...
+}
+
+/// Why IDL decode failed for an instruction, when it was attempted at all.
+/// Mirrors `solana_parser::solana::structs::IdlParseError`, flattened for
+/// Borsh (that upstream type carries no Borsh derive). `None` on
+/// `parsed_instruction_data`/`solana_rpc_parsed_data`'s siblings means either decode
+/// succeeded or no IDL was available to attempt against in the first place --
+/// distinct from an attempt that ran and failed, which this type identifies.
+#[derive(BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
+pub enum SolanaIdlParseError {
```
pepe-anchor, line 223: should: `SolanaSimulationError` was given `#[borsh(use_discriminant = true)]` and discriminants starting at 1 (line 78) precisely so a decoder that renders `Option::None` as the inner type's zero value (borsh-go v0.3.1, per this PR's own comment) can tell absent from `InvalidBase64`, but `Option<SolanaIdlParseError>` on both `SolanaIntermediateInstruction` and `SolanaSimulatedInstruction` leaves `DataParseError` at index 0, so such a decoder renders every clean instruction as a `DataParseError` with empty strings; apply the same `use_discriminant` plus start-at-1 treatment here.

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,51 +367,441 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+    let mut records: std::collections::HashMap<_, _> = builtin_idl_records()
+        .iter()
+        .map(|(k, v)| (k.clone(), v.clone()))
+        .collect();
+    for (program_id, record) in crate::idl::builtin_programs::preset_idl_records() {
+        records.insert(program_id.clone(), record.clone());
+    }
+    for (program_id, record) in caller_records {
+        records.insert(program_id.clone(), record.clone());
+    }
```
pepe-anchor, line 643: should: `build_idl_record_map` clones all 21 builtin records plus all 19 preset records (1,989,106 bytes of preset JSON on disk, plus the parsed `Idl` trees) into a fresh `HashMap` on every `extract_solana_intermediate_output` call, which flatly contradicts three doc comments added in this same PR: `preset_idl_records` says "the map is never cloned", `lookup_idl_record` says the presets are "borrowed from their process-wide caches instead of being cloned into a fresh map on every request", and `decode_inner_instructions` says "neither the ~2.0 MB of preset IDLs nor the builtins are cloned per request". The layering only avoids the clone on the simulation path; the static path pays it every time. Either make `parse_transaction_with_idl_records` take the map by reference upstream, or correct the three comments so the next reader does not trust a property the primary path does not have.

proto/parser/parser.proto
```diff
@@ -139,6 +139,12 @@ message SolanaMetadata {
   // Map of program_id (base58 string) to IDL definitions
   // Allows wallet to provide multiple IDLs, one per program
   map<string, Idl> idl_mappings = 3;
+  // Raw simulateTransaction RPC response bytes (innerInstructions section),
+  // base64-encoded (matching unsigned_payload's convention), sent with no
+  // caller-side decode/reshape. The parser runs this through the same
+  // static-decode path (parse_transaction_with_idls) used for
+  // unsigned_payload.
```
pepe-anchor, line 146: should: The comment tells wallet integrators the parser "runs this through the same static-decode path (parse_transaction_with_idls) used for unsigned_payload", but the function was renamed to `parse_transaction_with_idl_records` in this very PR and simulated results never touch it: they go through `parse_and_decode_simulated_instructions` / `decode_inner_instructions`, and the `UiParsedInstruction::Parsed` branch does no decode at all, it passes the RPC's own JSON through verbatim. This comment is copied verbatim into the generated `src/generated/src/generated/parser.rs:249-253` and is the public contract, so rewrite it to say what actually happens, including that the field is untrusted caller input.

proto/parser/parser.proto
pepe-anchor, line 147: should: `SolanaMetadata` jumps from tag 3 to tag 5, leaving tag 4 (the `simulated_instructions` field added in 70fb0bed and removed in f2187da0) free for a future author to reuse against any build that shipped the old shape; `EthereumMetadata` in this same file reserves its removed tag (lines 124-125), so follow that pattern.

```suggestion
  reserved 4;       // was: repeated SimulatedInstruction simulated_instructions
  reserved "simulated_instructions";
  optional string simulated_transaction_result = 5;
```

src/chain_parsers/visualsign-solana/src/cli_plugin.rs
```diff
@@ -18,6 +18,17 @@ pub struct SolanaArgs {
...
+    /// Base64 of a `simulateTransaction` RPC result, used to decode the
+    /// transaction's inner (CPI) instructions. Same value the API's
+    /// `simulated_transaction_result` field carries.
+    ///
+    /// The encoded JSON must be the bare `result` object
+    /// (`{"context":..,"value":..}`), not the full JSON-RPC envelope, and the
+    /// simulation must have been run with `jsonParsed` encoding and
+    /// `innerInstructions: true`.
+    #[arg(long = "simulated-tx-result", value_name = "BASE64")]
+    pub simulated_tx_result: Option<String>,
```
pepe-anchor, line 31: should: `docs/changelog/_drafts/README.md:46-52` requires a fragment for "a new or changed CLI, gRPC, or HTTP surface" and for "a change to VisualSign payload shape, field names, or determinism guarantees", and `docs/changelog/_drafts/` still holds only README.md; separately `docs/api-reference.mdx:71-75` reproduces `message SolanaMetadata` for wallet integrators without `simulated_transaction_result = 5`, and `docs/parser-cli.mdx:28-36` does not list `--simulated-tx-result`.

src/chain_parsers/visualsign-solana/src/intermediate.rs
```diff
@@ -248,51 +367,441 @@ impl From<&SolanaParsedInstructionData> for SolanaParsedInstructionDataIo {
...
+    // A malformed base58 payload isn't an IDL-resolution problem (the IDL
+    // resolved fine); treat it as "no instruction could be matched" since
+    // there's no byte data to check a discriminator against.
+    let Ok(data) = bs58::decode(data_base58).into_vec() else {
+        return (
+            None,
+            Some(SolanaIdlParseError::DiscriminatorNotFound(
+                "instruction data is not valid base58".to_string(),
+            )),
+        );
```
pepe-anchor, line 698: nit: The caller already decoded `decoded.data` at line 510 and returned `InvalidInstructionData` on failure, so this decodes the same string a second time and its `DiscriminatorNotFound("instruction data is not valid base58")` branch can never be reached; pass the already-decoded `&data` down instead of `data_base58` and drop the branch.

src/chain_parsers/visualsign-solana/src/idl/builtin_programs.rs
```diff
@@ -241,6 +275,157 @@ pub fn is_reserved_canonical_name(name: &str) -> bool {
...
+                        program_id: program_id.clone(),
+                        program_name: format!(
+                            "Preset Program {}",
+                            &program_id[..8.min(program_id.len())]
+                        ),
+                        program_type: None,
+                        custom_idl: Some(idl),
+                        custom_idl_json: Some(json.clone()),
+                        override_builtin: true,
+                        is_preset: true,
```
pepe-anchor, line 421: nit: `program_name: format!("Preset Program {}", &program_id[..8])` throws away names this crate already knows (`builtin_idl_program_name` returns "Jupiter Aggregator V6", "Drift Protocol V2", "Orca Whirlpool" for four of these IDs, and `canonical_name` is right there); use it, falling back to the placeholder only for preset-only IDs.

shahan-khatchadourian-anchorage (changes requested):
Re-reviewed at `53114f67`. Keeping changes requested.

Good progress since the last round: `response.value.err` is now checked, so a reverted simulation no longer gets treated as a completed one; the per-request IDL clone (`merge_preset_idl_configs`) is gone in favor of memoized `OnceLock` caches; and `--simulated-tx-result` on the CLI means this path is finally testable outside the gRPC server. The new `SolanaIdlParseError` variants are a real improvement over silently returning an empty trace.

Still open — four of these are the same objections from my two earlier rounds, not new ground, just confirming they haven't moved and pepe independently landed on the same four:

1. **Preset records still override builtins** (`intermediate.rs:640`, see pepe's comment there). This is my 26 Aug objection ("static decode output changed") back under a different function after the `merge_preset_idl_configs` removal — `build_idl_record_map` still inserts every preset record over its builtin counterpart unconditionally, and those records carry `program_type: None`, so the builtin fallback is gone. Pepe's diff makes the cost concrete: Meteora DLMM loses 34 instructions, Drift 20, Jupiter Agg V6 2.

2. **`metadata_digest` still shifts for every Solana request** (`parser.proto:147`, pepe's comment). Flagged in my 12 Aug review, carried over in 26 Aug, still unfixed — `simulated_transaction_result` has no `#[borsh(skip)]` the way `intermediate_output` does, so this isn't gated behind opt-in the way the PR body claims.

3. **Simulated `index` isn't checked against the transaction** (`core/visualsign.rs:498`, pepe's comment). I filed this as a smaller nit on 26 Aug; pepe's framing is the sharper one — nothing stops a caller from attaching a trace for the wrong transaction, which matters specifically because PROF-354 wants the HSM to trust this field.

4. **`jsonParsed` arm drops `accounts`/`instruction_data_hex`** (`intermediate.rs:557`, pepe's comment). Same concern as my 12 Aug note that `instruction_data_hex` is what the SCX allowlist actually checks — for RPC-parsed native programs there's nothing left but caller-supplied `parsed_json`.

5. **`bs58` breaks `--no-default-features`** (`intermediate.rs:510`/`:692`) — new, not something either of us raised before. It's gated as optional behind `cli-plugin` but used unconditionally in library code. Nothing in CI exercises that build today, so it's dormant, but it's a one-line fix.

Not a code objection, but worth resolving before merge: prasanna's comment on `src/Cargo.lock` — the `solana_parser` pin is a branch tip three levels off any `main` with no open PR. She's reviewed the actual commits there and they're good; the ask is just to give it somewhere durable to point to.

shahan-khatchadourian-anchorage:
@prasanna-anchorage — thanks for the thorough re-verification on the simulation-error handling and the IDL caching, both of those read clean to me too.

On the static-decode point, though: I think the objection is still open, just moved. `extract_solana_intermediate_output` does only pass `caller_idl_records(configs)` as you said, but the third argument to `parse_transaction_with_idl_records` is `build_idl_record_map(&caller_records)`, and that function still inserts every preset `IdlRecord` over its builtin counterpart unconditionally (`intermediate.rs:640`). `preset_idl_records()` builds those with `program_type: None`, so the builtin fallback `resolve_idl_for_record` would otherwise use is gone for any program that's both a preset and a `solana_parser` builtin. Pepe caught this independently right after your review, with diffed instruction counts — Meteora DLMM loses 34, Drift 20, Jupiter Agg V6 2 — so it's not just a theoretical read of the code.

I've gone back to `CHANGES_REQUESTED` on that basis, plus a few other things pepe found (digest stability, an unvalidated `index` on caller-supplied simulation data, `bs58` breaking `--no-default-features`) — details in the review.

Agreed on the `Cargo.lock` pin — that's a fair ask and I'd treat it the same way, a precondition rather than a nit.

shahan-khatchadourian-anchorage (dismissed):
LGTM based on discussions. Thanks for this PR @rituparna-mandal-anchor

pepe-anchor (approved):
Approving based on the thread, thanks for walking through the preset/builtin tradeoff and the jsonParsed limitation there, that context settles most of what was still open.

Two things I'd still like to see before merge, both small:

1. RegisteredSource has Native at discriminant 0. Since we're already bumping the schema to v2 here, worth reordering so an unregistered/unknown value fails closed instead of defaulting to Native.
2. metadata_digest shifts for every Solana request now that simulated_transaction_result isn't borsh-skipped. Not asking for a code change here, just want it called out in the PR description and confirmed with whoever owns the mirror decoder so it doesn't surprise anyone on deploy.

Rest of it (preset freshness, docs) can follow up separately.

PR #466, opened 2026-08-24
feat(psc-3684): add SVM Governance visualizer preset for Solana

kyle-anchorage:
## 1. Why this PR exists

https://linear.app/anchorlabs/issue/PSC-3684

Transactions to the Solana Foundation governance program (`govYkyQ3ePtGULAtY6V75qjWE8UH4vCUVQ1W4HdCAZU`) now render as named governance actions — `SVM Governance: cast_vote` with the three vote weights and the named `signer` / `proposal` / `vote` accounts. Today they fall through to the `unknown_program` catch-all, which shows a program id and a hex blob.

PSC-3684 asks for the governance program to be added "so governance vote and vote-override operations are human-readable", from exactly the IDL file this PR bundles. A signer approving a governance vote cannot tell from the catch-all rendering which proposal they are voting on or how their stake is being split. Vote override is the sharper case: it claims a specific stake account's weight through a Merkle proof, and none of that is visible today.

The preset is IDL-driven and follows the existing `dflow_aggregator` shape, so the decode path is the one already exercised by 17 other presets. 11 new tests cover it, including an end-to-end test that drives a real bincode-encoded transaction through the converter and asserts both that dispatch lands on this preset and that the payload passes charset validation.

## 2. What changed

- **New preset** `chain_parsers/visualsign-solana/src/presets/svm_governance/` — bundled IDL (14 instructions), a config claiming the program with a `*`/`*` instruction match, and the renderer. `build.rs` discovers `SvmGovernanceVisualizer` from the directory name; there is no registry file to edit.
- **`VisualizerKind::Governance`** added to `solana/src/core/mod.rs`. Rejected alternatives: `Payments`, the bucket holding System / Squads / Swig, which mislabels governance in metadata that ships with the payload; and `StakingPools`, closer in spirit since votes are stake-weighted, but no instruction here stakes or unstakes. `kind()` is descriptive metadata — the one match on it (`swig_wallet`) has a `_` arm, so the variant is purely additive.
- **Renderer decisions.** One field per top-level IDL argument: `stake_merkle_proof` is a `Vec<[u8; 32]>`, so recursing into elements would bury the vote weights under per-byte fields. Byte arrays render as `0x`-hex. Every leaf string passes through `charset_safe`, because `create_proposal` carries a free-form `title` and `description` supplied by the proposer — the one place this program lets a transaction push arbitrary UTF-8 into a field value.
- **Uses `InstructionView::from_context`, not `resolve_accounts()?`.** The fallible accessors abort a v0 transaction whose accounts come from an address lookup table, since those never resolve against the static key slice. `InstructionView` renders `unresolved(N)` instead. This matches every other IDL preset.
- **Failure-mode shape change**: dispatch in `visualize_with_any` for program `govYkyQ...` — *`unknown_program` catch-all -> `SvmGovernanceVisualizer`*; affects the rendering of every instruction to this program and nothing else, since `can_handle` matches on the exact program id; mitigation: an undecodable instruction (unknown discriminator, truncated data) degrades to a labeled `SVM Governance: Unknown Instruction` fallback carrying the program id and the raw data, so no input that rendered before can now fail.
- **Carried review finding (P2).** `format_arg_value`, `charset_safe`, and `bytes_as_hex` are byte-identical copies of `dflow_aggregator`'s. Not extracted here, for two reasons. Three different `format_arg_value` implementations already exist across the 19 presets, so there is no canonical helper to reuse — choosing one is a design call over 19 files, not a mechanical extraction. And `dflow_aggregator`'s copy documents itself as a stopgap awaiting a DFlow-aware renderer, so extracting a shared helper now would fork again shortly. Filed as https://linear.app/anchorlabs/issue/PRS-632 rather than silently adding the copy.
- Nothing removed.

## 3. Test plan

**Agent-verified (automated)**

- `make -C src test` — exit 0, 39 test binaries, 1636 tests pass.
- `cargo clippy -p visualsign-solana --all-targets -- -D warnings` — clean both with and without `diagnostics`. Both configurations matter: `parser_app` builds without the feature, `parser_cli` with it.
- `cargo fmt --all -- --check` — clean.
- 13 new tests: the IDL loads and every instruction carries an 8-byte discriminator; an unknown discriminator and 3-byte data both error; `cast_vote` renders its three weights and names `signer` / `proposal`; `create_proposal` strips a non-ASCII title and description while keeping the printable ASCII; a `Vec<[u8; 32]>` Merkle proof renders as two `0x`-hex nodes rather than 64 per-byte fields; surplus accounts surface as `Remaining Account N`; and the unknown-instruction fallback renders.
- `test_cast_vote_override_renders_weights_proof_and_leaf` pins the ticket's own acceptance criterion for the harder of the two named operations: a full `cast_vote_override` payload decodes, and the weights, the two Merkle proof nodes, and the leaf's voting wallet, stake account, and stake amount all render as legible fields.
- Two more pin the contracts this PR takes on rather than the code's own shape. `test_undecodable_instruction_renders_fallback_instead_of_erroring` drives `visualize_tx_commands` with garbage and asserts it still returns a rendered field carrying the raw bytes — that is the failure-mode contract named in section 2, and it fails if the fallback branch ever becomes an error. `test_end_to_end_transaction_routes_here_and_passes_charset` builds a real bincode + base64 transaction with an emoji in the proposal title, then asserts dispatch reaches this preset, `validate_anchorage_wallet_renderable()` is clean, and `to_validated_json()` succeeds.
- Discriminators are looked up from the bundled IDL rather than hardcoded, so an IDL regeneration cannot leave the tests asserting stale bytes.
- The bundled IDL is byte-identical to the upstream file at `solana-foundation/solana-governance@main:svmgov/cli/idls/svmgov_program.json`.
- Review: `review-changes` (adversarial 4-lens) run on this HEAD. Two findings fixed — the missing failure-mode test above, and a decode-failure warning that did not name the instruction index. One finding carried, see below.

**Human QA (manual checklist)**

- [x] No UI surface — the deliverable is a decode path. The end-to-end test covers what a manual CLI decode would have covered, driving the full `SolanaVisualSignConverter` path rather than the visualizer in isolation.
- [ ] Not done: decode of a real mainnet transaction. Every test payload is synthetic, built from the bundled IDL's own discriminators. Real Cast Vote and Modify Vote samples are being gathered from the validator governance portal on the ticket's own Slack thread; when they land, they should become a `tests/fixtures/` snapshot like `orca_whirlpool`'s. That fixture is the one gap these tests leave open.

## 4. How this is maintainable by Claude

A fresh agent can work on this without broad exploration. The preset is one self-contained directory of ~530 lines plus the IDL JSON, which no agent needs to read in full. It is a near-copy of `dflow_aggregator`, so reading any one IDL preset transfers directly.

`CLAUDE.md` guards the parts easy to get wrong: field builders instead of hand-built structs, ASCII-only user-visible strings, no `unwrap` / `expect` / `panic` outside tests, `BTreeMap` for deterministic ordering. The tests catch the two regressions this file shape invites — field explosion on a nested argument, and non-ASCII reaching `validate_charset`.

One trap worth naming: the directory name is load-bearing. `build.rs` derives `SvmGovernanceVisualizer` from `svm_governance` by PascalCase, so renaming the directory without renaming the struct breaks the build.

## Manual steps needed (by human)

None — fully automated by CI. No feature flag, no migration, no backfill. Rollback is a revert of a single commit.

One product-visible note, **low** risk: the display name `SVM Governance` and the new `Governance` kind reach wallet-facing metadata. Worth a glance if that string has a convention I missed.

### DIFF
--- a/src/chain_parsers/visualsign-solana/src/core/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/core/mod.rs
@@ -32,6 +32,8 @@ pub enum VisualizerKind {
     StakingPools(&'static str),
     /// Payment and simple transfer-related operations
     Payments(&'static str),
+    /// On-chain governance: proposals, voting, and protocol administration
+    Governance(&'static str),
 }
 
 /// Resolution of a compiled instruction's program_id_index.
--- a/src/chain_parsers/visualsign-solana/src/presets/svm_governance/config.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/svm_governance/config.rs
@@ -1,0 +1,22 @@
+use super::SVM_GOVERNANCE_PROGRAM_ID;
+use crate::core::{SolanaIntegrationConfig, SolanaIntegrationConfigData};
+use std::collections::BTreeMap;
+
+pub struct SvmGovernanceConfig;
+
+impl SolanaIntegrationConfig for SvmGovernanceConfig {
+    fn new() -> Self {
+        Self
+    }
+
+    fn data(&self) -> &SolanaIntegrationConfigData {
+        static DATA: std::sync::OnceLock<SolanaIntegrationConfigData> = std::sync::OnceLock::new();
+        DATA.get_or_init(|| {
+            let mut programs = BTreeMap::new();
+            let mut instructions = BTreeMap::new();
+            instructions.insert("*", vec!["*"]);
+            programs.insert(SVM_GOVERNANCE_PROGRAM_ID, instructions);
+            SolanaIntegrationConfigData { programs }
+        })
+    }
+}
--- a/src/chain_parsers/visualsign-solana/src/presets/svm_governance/mod.rs
+++ b/src/chain_parsers/visualsign-solana/src/presets/svm_governance/mod.rs
@@ -1,0 +1,652 @@
+//! SVM Governance preset implementation for Solana
+
+mod config;
+
+use crate::core::{
+    InstructionView, InstructionVisualizer, SolanaIntegrationConfig, VisualizerContext,
+    VisualizerKind,
+};
+use config::SvmGovernanceConfig;
+use solana_parser::{
+    Idl, SolanaParsedInstructionData, decode_idl_data, parse_instruction_with_idl,
+};
+use std::collections::BTreeMap;
+use std::sync::OnceLock;
+use visualsign::errors::VisualSignError;
+use visualsign::field_builders::{create_raw_data_field, create_text_field};
+use visualsign::{
+    AnnotatedPayloadField, SignablePayloadField, SignablePayloadFieldCommon,
+    SignablePayloadFieldListLayout, SignablePayloadFieldPreviewLayout, SignablePayloadFieldTextV2,
+};
+
+pub(crate) const SVM_GOVERNANCE_PROGRAM_ID: &str = "govYkyQ3ePtGULAtY6V75qjWE8UH4vCUVQ1W4HdCAZU";
+
+const SVM_GOVERNANCE_DISPLAY_NAME: &str = "SVM Governance";
+
+const SVM_GOVERNANCE_IDL_JSON: &str = include_str!("svm_governance.json");
+
+static SVM_GOVERNANCE_CONFIG: SvmGovernanceConfig = SvmGovernanceConfig;
+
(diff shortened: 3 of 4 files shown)


kyle-anchorage:
@copilot review

PR #472, opened 2026-08-29
perf(solana): make intermediate_output a projection of one decode

prasanna-anchorage:
## Problem

`intermediate_output` re-parsed the transaction instead of being a genuinely intermediate representation.

Two independent decoders ran over the same bytes:

- the human-readable `SignablePayload` came from `instructions::decode_transfers` → `parse_transaction`
- the intermediate came from `extract_solana_intermediate_output` → `parse_transaction_with_idls`, over the re-serialized message

So opting into the intermediate cost a full extra parse, and the two decoders could disagree on the same input — the "user sees X, policy consumes Y" surface already flagged in `build_intermediate_bytes`'s doc comment.

## Change

Decode once, up front, when the caller opts in, and render the payload from the same `SolanaMetadata` the intermediate is projected from.

- **`intermediate.rs`** — split `parse_solana_metadata` (the single decode) from the projection. `extract_solana_intermediate_output` is now a thin wrapper.
- **`core/instructions.rs`** — add `decode_transfers_from_metadata`, the projection half of `decode_transfers`, so a caller holding decoded metadata renders from it rather than decoding again.
- **`core/visualsign.rs`** — thread `shared_metadata` through the legacy, versioned and v0 payload builders; project the intermediate from that same value. Adds `build_solana_intermediate_output`, a public entry point that stops at intermediate creation.

One subtlety worth flagging for review: a well-formed legacy transaction is wire-identical to `VersionedTransaction::Legacy`, so `from_string` routes essentially all real input through the **versioned** path. An earlier pass that wired only the legacy branch measured no improvement at all. The gains below only appear once the versioned and v0 paths consume the shared decode too.

## Compatibility

The default signing path (`include_intermediate_output: false`) performs no shared decode and is unchanged — same bytes, same signed digest. No fixture snapshots moved. 1603 tests pass; `clippy --all-targets -D warnings` clean.

## Benchmarks

Adds criterion per-stage benches, profilable via `--profile-time` with `perf` / `samply`:

- `parser/app/benches/parse_stages.rs` — decode, convert, charset validation, JSON serialization, digest, P-256 sign, registry construction, end-to-end `parse()`
- `chain_parsers/visualsign-solana/benches/solana_stages.rs` — decode, intermediate-only, convert, convert-with-intermediate over the real protocol fixtures

All figures below are criterion medians, `--release`, measured in **one session on one
machine** at branch HEAD `4be1aaf7`. The "before" column is the parent commit `e77273ea`
running the same benchmark harness, measured back-to-back with the "after" column — absolute
timings drift ~20% between sessions on this box, so cross-session comparison would not have
been sound.

### Solana — the intermediate's cost over render-only

| fixture | before: `convert` | before: `+intermediate` | before: overhead | after: `convert` | after: `+intermediate` | **after: overhead** | e2e change |
|---|---:|---:|---:|---:|---:|---:|---:|
| `spl_mint_to` | 20.73 µs | 32.05 µs | +54.6% | 20.95 µs | 21.01 µs | **+0.3%** | −34.4% |
| `t22_burn_checked` | 25.81 µs | 37.98 µs | +47.2% | 26.19 µs | 26.04 µs | **−0.6%** | −31.4% |
| `t22_set_authority` | 27.77 µs | 40.03 µs | +44.2% | 27.65 µs | 28.30 µs | **+2.4%** | −29.3% |
| `orca_increase_liquidity` | 2.877 ms | 4.602 ms | +60.0% | 2.898 ms | 2.917 ms | **+0.7%** | −36.6% |
| `jupiter_route` | 2.550 ms | 3.459 ms | +35.7% | 2.550 ms | 2.570 ms | **+0.8%** | −25.7% |

The intermediate's surcharge drops from **+36..60% to within noise**. Note also that the
`convert` columns are unchanged before-to-after — the default signing path
(`include_intermediate_output: false`) neither regresses nor improves, which is the
behaviour-preservation claim above stated numerically.

### Solana — per stage

`intermediate_only` is the new public entry point that stops at IR construction. `decode` is
`from_string`; both `convert` columns are as above.

| fixture | `decode` | `intermediate_only` | `convert` | `convert_with_intermediate` |
|---|---:|---:|---:|---:|
| `spl_mint_to` | 0.76 µs | 8.19 µs | 20.95 µs | 21.01 µs |
| `t22_transfer_checked` | 0.82 µs | 10.01 µs | — (see follow-up 2) | — |
| `t22_burn_checked` | 0.73 µs | 8.48 µs | 26.19 µs | 26.04 µs |
| `t22_set_authority` | 0.81 µs | 8.59 µs | 27.65 µs | 28.30 µs |
| `orca_increase_liquidity` | 1.78 µs | 1.599 ms | 2.898 ms | 2.917 ms |
| `jupiter_route` | 2.03 µs | 0.914 ms | 2.550 ms | 2.570 ms |

The structured decode is roughly half the cost of the full conversion on the two complex
protocols, and decoding itself is negligible everywhere — the cost is in the preset
rendering.

### Ethereum — per stage on the signing path

Context for the `parse_stages` bench this PR also adds. Ethereum emits no intermediate
output, so its `convert` / `convert_with_intermediate` rows are identical by construction —
which is itself a useful control.

| stage | `legacy_transfer` | `eip1559` | `uniswap_v3_swap` |
|---|---:|---:|---:|
| `decode` | 0.69 µs | 8.43 µs | 9.86 µs |
| `convert` | 3.41 µs | 9.82 µs | 13.43 µs |
| `convert_with_intermediate` | 3.50 µs | 9.66 µs | 13.45 µs |
| `validate_charset` | 22.9 µs | 164.8 µs | 215.5 µs |
| `serialize_json` | 9.60 µs | 99.5 µs | 135.8 µs |
| `digest` (borsh + SHA-256) | 9.65 µs | — | — |
| **end-to-end `parse()`** | **466 µs** | **918 µs** | **1.065 ms** |

Two things worth a reviewer's attention, both pre-existing and out of scope here:

- **Parsing is ~2% of end-to-end cost.** `decode` + `convert` on the Uniswap swap is 23 µs
  out of 1.065 ms. The expensive stages are `sign_p256` (~226 µs), `validate_charset`
  (215 µs) and `serialize_json` (136 µs). Optimising decoders would buy almost nothing today.
- **`validate_charset` runs twice.** `registry_dispatch` measures 260 µs against 13.4 µs for
  `convert` alone; the difference is charset validation inside the converter's `from_string`
  path. `parse.rs` then calls `validate_charset()` again as defence-in-depth (deliberately —
  the comment explains why chain overrides can skip the inner one). That is ~20% of
  end-to-end, spent twice.

## Follow-ups (not in this PR)

1. `decode_instructions` and `decode_v0_instructions` still run their own decodes, so instruction-level fields are not yet projections of the shared metadata. That is the larger remaining chunk and the point where rendered output would actually change. The drift test remains the guard until then.
2. **`token_2022/transfer_checked` is a declared capability gap, not drift.** (Corrected — an earlier revision of this description overstated it as a mysterious divergence.) That fixture is the only one carrying `expected_error` rather than `expected_fields`; its description says "UNSUPPORTED - should fail" and `presets/token_2022/tests/fixture_test.rs:288` asserts the rejection. `presets/token_2022/mod.rs:210` allow-lists five `TokenInstruction` variants; `TransferChecked` unpacks fine but falls to the catch-all at `mod.rs:281`. Still worth closing — it is an ordinary mainnet transfer the intermediate decodes and the renderer cannot show — and `presets/spl_token/mod.rs:302` already has a working arm to port. Its error text ("unknown discriminator 0x0c") is also misleading: the discriminator is known, the match arm is missing.

3. Complex-protocol conversion is now the dominant cost in the system — Orca at 2.4 ms and Jupiter at 2.2 ms on a single-instruction transaction, versus ~20 µs for token ops. Worth profiling on its own.


src/chain_parsers/visualsign-solana/src/core/visualsign.rs
```diff
@@ -724,7 +747,13 @@ fn convert_v0_to_visual_sign_payload(
 
     // Process V0 transfer decoding using solana-parser
     if decode_transfers {
-        match decode_v0_transfers(versioned_tx) {
+        // Render from the shared decode when one exists, so the v0 path is a
+        // projection of the same parse the intermediate comes from.
+        let v0_transfers = match shared_metadata {
+            Some(metadata) => Ok(instructions::decode_transfers_from_metadata(metadata)),
```
pepe-anchor, line 753: Question, not a bug claim: is dropping the `V0 ` label prefix here intentional? The same v0 SPL transfer now renders as `"V0 SPL Transfer N"` with `include_intermediate_output=false` and `"SPL Transfer N"` with it `=true`, driven by a flag that's meant to add policy metadata, not change display text. Amount/mint/destination are unaffected, so no fund-loss risk, but the returned `SignablePayload` is no longer independent of that flag. Worth confirming it's deliberate, or keeping the `V0 ` prefix in the shared render path.

---

pepe-anchor, line 753: This is the only test covering `include_intermediate_output: true`, and it's a legacy transfer, not v0. Nothing exercises the v0 + shared-decode branch with the flag on, which is exactly the case that changes the label above. Worth adding a v0 case asserting label parity between flag on/off.

---

pepe-anchor (commented):
Reviewed with [Claude](https://claude.com/claude-code) on behalf of @pepe-anchor. 2 suggestion(s) of 12 drafted; 0 marked as blocking. _Treat findings as suggestions, not directives._

Nice cleanup on the single-decode path. Two small things worth a look, both about `include_intermediate_output` changing v0 transfer labels for the same transaction, see inline.

prasanna-anchorage:
Re-derived on current `main` (`eaeb1c66`) and force-pushed — `46404d5b` → `449b7555`. Now `MERGEABLE`.

**Not a rebase, deliberately.** I tried one first and aborted it. #465 rewrote `intermediate.rs` (+699) and grew `core/visualsign.rs` with simulation support, so git aligned a **126-line block of `main` against 18 lines of this PR**. Hand-merging that on a signing path is how a wrong resolution slips through looking clean — so I re-derived instead and kept the aborted rebase out of the history.

Three things changed versus the original commit, and two of them change what this PR *claims*:

**1. The projection is not pure, by design — and the old code no longer compiled.** #465 replaced `From<&SolanaMetadata> for SolanaIntermediateOutput` with `build_intermediate_output(value, caller_idl_program_ids)`, because each instruction's `registered_source` records whether its program had a registered IDL, which `SolanaMetadata` doesn't carry. The original commit called `SolanaIntermediateOutput::from(&metadata)`; that impl is gone. So the achievable property is **one decode feeding both views**, with the intermediate carrying registry-derived provenance the rendered payload does not. The docstring now says that instead of claiming a pure projection — and anyone writing a divergence test needs it, since field-for-field equality is the wrong expectation.

**2. The scope is narrower than the title, and the code already said so.** Transfer fields become projections of one decode; `instructions::decode_instructions` and `decode_v0_transfers` still run their own decodes, so instruction-level fields stay guarded by the drift test rather than by construction. I'd been describing this PR as "the structural answer" to dual-decoder divergence in scoping notes — that overstated it, and I've corrected those.

**3. `build_intermediate_bytes` is gone, but its behaviour isn't.** #465's simulation merge and the unresolved-instruction count logging are preserved inline in the new single-decode path, along with the best-effort degradation: a message the decoder can't handle still renders, it just carries no policy metadata.

The benches also needed APIs that landed since this was written — `create_registry(&ParserConfig)` and the 3-argument `parse()` (#454), and `ParseRequest::payment_marker` (#448).

**Verification:** `make lint` exit 0, `make test` exit 0 (44 suites / 2222 tests). And I checked the actual point rather than just the build — the only surviving `extract_solana_intermediate_output` call is inside `build_solana_intermediate_output`, the standalone decode-only entry point that is supposed to parse. The conversion path decodes exactly once.

It is `BLOCKED` pending review, not on anything from me.

src/chain_parsers/visualsign-solana/src/core/visualsign.rs
prasanna-anchorage: Not intentional — this was a real bug, and your question found it. Thank you for asking rather than assuming.

The refactor rebased V0 SPL rendering onto the shared projection without carrying the `"V0 "` prefix, so the same V0 transaction rendered:

- `include_intermediate_output: false` -> `label: "V0 SPL Transfer 1"`
- `include_intermediate_output: true`  -> `label: "SPL Transfer 1"`

Amounts, addresses, `fallback_text` and `text_v2` were identical; only the label differed. That matters more than a cosmetic slip: `intermediate_output` is borsh-appended into the signed digest, so an unrelated *request flag* changed the consent artifact a signer sees for identical transaction bytes. A PR whose premise was "decode once so rendering and the intermediate cannot diverge" introduced a divergence of its own.

Fixed in `2608b207` (from #499, merged into this branch): `decode_transfers_from_metadata` now takes an `spl_label_prefix`, so both callers render their own label from the one projection. The prefix is preserved rather than unified away, because that is what `main` renders today and a fix should restore existing behaviour, not change a signed field.

I verified it by mutation rather than by reading: restoring the fix and then flipping only the V0 flag-on call site back to `"SPL Transfer"` reproduces your case exactly, and the new guard fails on it. Post-rebase, all four paths agree per transaction type:

| path | flag off | flag on |
|---|---|---|
| legacy | `instructions.rs:273` -> `"SPL Transfer"` | `visualsign.rs:633` -> `"SPL Transfer"` |
| V0 | `v0.rs:63` -> `"V0 SPL Transfer"` | `visualsign.rs:817` -> `"V0 SPL Transfer"` |

One thing your question exposed that is still open, filed as a follow-up rather than fixed here: SPL `fallback_text` hard-codes `"SPL Transfer"` even when `label` is `"V0 SPL Transfer"`, so one signed field carries two different descriptions of the same transfer. That mismatch predates this PR (the deleted hand-rolled renderer had it too), so it is not a regression — but it is the same class of thing, and worth closing separately.

prasanna-anchorage: You were right, and this is the more important of your two comments — the missing coverage is *why* the `V0 ` label bug shipped unnoticed.

The PR's whole premise is that `payload(flag=false) == payload(flag=true)`, and nothing asserted it. I recorded this PR as "verified" on `make lint` + `make test` passing, but a green suite that never exercises the load-bearing invariant is not verification. That was my error, not a reviewer's.

`2608b207` adds two guards that drive the production converter (`SolanaVisualSignConverter::to_visual_sign_payload`, not an internal helper):

- `payload_is_unchanged_by_include_intermediate_output_legacy_sol`
- `payload_is_unchanged_by_include_intermediate_output_v0_spl` — a real V0 SPL transfer, the case that regressed

The V0 one also asserts the SPL label field is *present* before comparing, so it cannot pass vacuously by matching two payloads that both lack the field.

Mutation-checked: flipping the V0 flag-on call site back to `"SPL Transfer"` makes `..._v0_spl` fail and leaves `..._legacy_sol` green, so the pair catches the specific regression and isolates the V0 path.

On why two tests cover it: the untested combinations cannot diverge by construction — legacy SPL renders `"SPL Transfer"` on both sides, V0 SOL renders `"Transfer {}"` on both. V0 SPL was the only combination where the two literals differed.

Verified on the rebased head: `cargo test -p visualsign-solana --lib` 293 passed (parity tests reporting `running 2 tests`, not 0), `cargo clippy -p visualsign-solana -p parser_app --all-targets -- -D warnings` exit 0, `cargo check --benches` exit 0, `cargo fmt --check` clean.

PR #479, opened 2026-09-02
fix(near): resolve an MT withdraw that wraps a known NEP-141 balance

shahan-khatchadourian-anchorage:
`mt_withdraw` renders every `token_id` as raw base units with no symbol or decimal scaling, unlike its `ft_withdraw`/`transfer`/`token_diff` siblings, which resolve through the token registry. Of the eleven `Intent` kinds, only those three consulted it -- `mt_withdraw` was the odd one out, not a deliberate omission.

## Why it can resolve at all

NEP-245's `token_id` is a fully opaque string (`defuse_nep245::TokenId = String`) -- nothing in the standard lets an arbitrary id be traced back to an underlying asset. `defuse-core`'s own cross-instance MT transfer tests (`tests/src/tests/defuse/tokens/nep245/mod.rs`) exercise a real, code-verified convention: the intents verifier contract represents one of its own held NEP-141 balances as a multi-token by setting `mt_token_id` to that balance's own `TokenId` string verbatim, e.g. `nep245:intents.near:nep141:wrap.near`.

`tokens::mt_underlying_nep141` recognizes exactly this, and only this: it requires the outer `contract` to be `INTENTS_RECEIVER` (`intents.near`) *and* the inner id to parse as `TokenId::Nep141`. The contract check is load-bearing -- `mt_token_id` is a plain string fully controlled by whoever deploys `contract`, so an independent NEP-245 contract could otherwise mint a `token_id` that spells `nep141:wrap.near` and borrow that seeded asset's symbol/decimals for a balance it has no relationship to. Binding to the trusted verifier contract is what makes the borrow self-verifying; the inner id's spelling alone cannot be.

## What changed

- `render_mt_withdraw` builds the same `nep245:<contract>:<token_id>` asset id `TokenDiff`/`Transfer` already produce for an MT entry in their maps, and resolves it through the identical `token_amount_field`/`tokens::resolve` path. No new registry type, no curated table -- it reuses the existing NEP-141 registry/`SEEDS`/signed-metadata machinery entirely.
- `intent_consumes_token_registry`'s `MtWithdraw` arm flips to `true`, so a standalone envelope containing only an `mt_withdraw` now builds the registry too.
- `tokens::seeded_decimals` (the gap-fill check `try_extract_from_chain_metadata` uses to refuse an unattributed override against an already-curated asset) gains the same fallback. Without it, an MT id wrapping a seeded NEP-141 balance would have been exempt from the gap-fill rule even though its direct key is protected -- an unattributed entry could smuggle in the wrong decimals through the MT-shaped key alone.
- `mt_underlying_nep141` binds the alias to `INTENTS_RECEIVER` rather than trusting the inner id's spelling alone.

## Tests

- `tokens.rs`: MT-wraps-a-seeded-balance resolves on the trusted contract; an opaque/unrelated MT token_id doesn't; an inner id that parses but isn't NEP-141 (NFT, nested MT) doesn't; an untrusted contract paired with a seeded-looking inner id doesn't; a registry override keyed by the exact MT asset id wins over the underlying's metadata; `seeded_decimals` sees through the alias.
- `render.rs`: an `mt_withdraw` naming a wrapped `wrap.near` balance on the trusted contract renders a resolved `AmountV2`; one naming an unrelated token_id, or an untrusted contract with a seeded-looking token_id, renders the honest unresolved text; a `token_diff` entry keyed by an untrusted-contract MT id (the shape with no contract visible on screen at all) renders unresolved too.
- `convert.rs`: added to the gate/dispatch parity table (`every_intent_kind_the_registry_gate_recognizes_matches_render_intent`).
- `token_signature.rs`: an unsigned entry keyed by an MT asset wrapping a seeded balance on the trusted contract is rejected by the gap-fill rule, mirroring the existing direct-key regression test.

### DIFF
--- a/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
+++ b/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
@@ -1092,12 +1092,6 @@ mod tests {
             r#"{"intent":"token_diff","diff":{"nep245:evil.near:nep141:wrap.near":"-2000000000000000000000000"}}"#,
         );
         let fields = render_intent(&intent, &empty_reg()).expect("render");
-        assert!(
-            fields
-                .iter()
-                .all(|f| !matches!(f, SignablePayloadField::AmountV2 { .. })),
-            "expected no resolved AmountV2, got {fields:?}"
-        );
         match fields.iter().find(|f| label_of(f) == Some("Send")) {
             Some(SignablePayloadField::TextV2 { text_v2, .. }) => {
                 assert!(
--- a/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
+++ b/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
@@ -338,7 +338,7 @@ fn render_intent_body(intent: &Intent, registry: &Reg) -> Result<Fields, VisualS
         Intent::Transfer(t) => render_transfer(t, registry),
         Intent::FtWithdraw(w) => render_ft_withdraw(w, registry),
         Intent::NftWithdraw(w) => render_nft_withdraw(w),
-        Intent::MtWithdraw(w) => render_mt_withdraw(w),
+        Intent::MtWithdraw(w) => render_mt_withdraw(w, registry),
         Intent::NativeWithdraw(w) => render_native_withdraw(w),
         Intent::AddPublicKey(a) => Ok(vec![
             create_text_field("Add Public Key", &a.public_key.to_string())?.signable_payload_field,
@@ -380,9 +380,11 @@ fn render_intent_body(intent: &Intent, registry: &Reg) -> Result<Fields, VisualS
 /// rather than silently drifting out of step.
 pub(crate) fn intent_consumes_token_registry(intent: &Intent) -> bool {
     match intent {
-        Intent::TokenDiff(_) | Intent::Transfer(_) | Intent::FtWithdraw(_) => true,
+        Intent::TokenDiff(_)
+        | Intent::Transfer(_)
+        | Intent::FtWithdraw(_)
+        | Intent::MtWithdraw(_) => true,
         Intent::NftWithdraw(_)
-        | Intent::MtWithdraw(_)
         | Intent::NativeWithdraw(_)
         | Intent::AddPublicKey(_)
         | Intent::RemovePublicKey(_)
@@ -532,7 +534,13 @@ fn render_nft_withdraw(w: &NftWithdraw) -> Result<Fields, VisualSignError> {
     Ok(fields)
 }
 
-fn render_mt_withdraw(w: &MtWithdraw) -> Result<Fields, VisualSignError> {
+/// Each `token_id` here is scoped to `w.token`, the multi-token contract --
+/// `nep245:<w.token>:<token_id>` is the same asset-id shape `TokenDiff`/
+/// `Transfer` already produce for an MT entry in their maps, so it reaches
+/// `token_amount_field` through the identical registry lookup (including its
+/// `tokens::mt_underlying_nep141` fallback to a wrapped NEP-141 balance)
+/// rather than a bespoke one for this withdraw shape alone.
+fn render_mt_withdraw(w: &MtWithdraw, registry: &Reg) -> Result<Fields, VisualSignError> {
     if w.token_ids.len() != w.amounts.len() {
         return Err(VisualSignError::ValidationError(format!(
             "mt_withdraw token_ids/amounts length mismatch: {} token_ids vs {} amounts",
@@ -546,12 +554,15 @@ fn render_mt_withdraw(w: &MtWithdraw) -> Result<Fields, VisualSignError> {
     ];
     for (id, amount) in w.token_ids.iter().zip(w.amounts.iter()) {
         // As with `nft_withdraw`'s `token_id`, an MT token id is a plain
-        // `String` (`defuse_nep245::TokenId`); filter it before it joins the
-        // composite, so the amount half cannot be pushed onto its own line.
-        fields.push(
-            create_text_field("MT Token", &format!("{} x{}", charset_safe(id), amount.0))?
-                .signable_payload_field,
-        );
+        // `String` (`defuse_nep245::TokenId`), so it carries whatever bytes
+        // the sender chose and must be filtered before it reaches field text.
+        fields.push(create_text_field("MT Token", &charset_safe(id))?.signable_payload_field);
+        fields.extend(token_amount_field(
+            "Amount",
+            &format!("nep245:{}:{id}", w.token),
+            amount.0,
+            registry,
+        )?);
     }
     push_withdraw_call_details(&mut fields, &w.memo, &w.msg, w.storage_deposit)?;
     Ok(fields)
@@ -1334,6 +1345,46 @@ mod tests {
         assert!(labels.contains(&"Storage Deposit"), "labels: {labels:?}");
     }
 
+    /// An MT withdraw whose token_id names a NEP-141 balance a defuse-style
+    /// contract represents as a multi-token (the same convention
+    /// `tokens::mt_underlying_nep141` recognizes) must resolve through the
+    /// registry, not render the raw base-unit amount an unresolved asset
+    /// would.
+    #[test]
+    fn mt_withdraw_resolves_a_wrapped_nep141_balance() {
+        let intent = intent_from(
+            r#"{"intent":"mt_withdraw","token":"defuse.near","receiver_id":"alice.near","token_ids":["nep141:wrap.near"],"amounts":["2000000000000000000000000"]}"#,
+        );
+        let fields = render_intent(&intent, &empty_reg()).expect("render");
+        match fields.iter().find(|f| label_of(f) == Some("Amount")) {
+            Some(SignablePayloadField::AmountV2 { amount_v2, .. }) => {
+                assert_eq!(amount_v2.amount, "2");
+                assert_eq!(amount_v2.abbreviation.as_deref(), Some("wNEAR"));
+            }
+            other => panic!("expected resolved AmountV2, got {other:?}"),
+        }
+    }
+
+    /// An MT withdraw on an independent, unrelated multi-token contract --
+    /// one whose token_id has no relation to any known asset -- must still
+    /// render honestly as unresolved, not silently misattribute an amount.
+    #[test]
+    fn mt_withdraw_with_an_unrelated_token_id_is_unresolved() {
+        let intent = intent_from(
+            r#"{"intent":"mt_withdraw","token":"market.near","receiver_id":"alice.near","token_ids":["gold-sword-42"],"amounts":["5"]}"#,
+        );
+        let fields = render_intent(&intent, &empty_reg()).expect("render");
+        match fields.iter().find(|f| label_of(f) == Some("Amount")) {
+            Some(SignablePayloadField::TextV2 { text_v2, .. }) => {
+                assert!(
+                    text_v2.text.contains("unresolved"),
+                    "expected an unresolved amount, got {text_v2:?}"
+                );
+            }
+            other => panic!("expected an unresolved text field, got {other:?}"),
+        }
+    }
+
     /// A NEP-616 `state_init` attached to an intent, in defuse's JSON shape:
     /// an externally-tagged `StateInit::V1` wrapping a global-contract id.
     const STATE_INIT: &str =
@@ -1547,7 +1598,7 @@ mod tests {
         let fields = render_intent(&intent, &empty_reg()).expect("render");
         assert_eq!(
             text_at(&fields, "MT Token"),
-            "innocent?To: alice.near?Amount: 0.001 SOL x5"
+            "innocent?To: alice.near?Amount: 0.001 SOL"
         );
     }
 
--- a/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
+++ b/src/chain_parsers/visualsign-near/src/presets/intents/render.rs
@@ -1080,6 +1080,35 @@ mod tests {
         assert_eq!(wnear.abbreviation.as_deref(), Some("wNEAR"));
     }
 
+    /// A `token_diff` entry keyed by an MT asset id whose contract is NOT the
+    /// intents verifier, but whose inner token_id is spelled to round-trip as
+    /// a seeded NEP-141 id, must render unresolved -- this shape shows no
+    /// contract on screen at all (unlike `mt_withdraw`'s separate `Token`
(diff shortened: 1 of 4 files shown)


src/chain_parsers/visualsign-near/src/presets/intents/tokens.rs
```diff
@@ -114,26 +115,47 @@ fn usable(meta: TokenMeta) -> Option<TokenMeta> {
...
+/// inner id that isn't NEP-141 (an NFT, or a nested multi-token), does not
+/// match and returns `None`, leaving the amount unresolved exactly as it does
+/// today. `contract` is a NEAR account id, which cannot itself contain `:`, so
+/// splitting on the first `:` after the `nep245:` prefix cannot mistake part
+/// of `mt_token_id` for it even though `mt_token_id` may contain further `:`s.
+fn mt_underlying_nep141(asset_id: &str) -> Option<String> {
+    let mt_token_id = asset_id.strip_prefix("nep245:")?.split_once(':')?.1;
+    matches!(mt_token_id.parse::<TokenId>(), Ok(TokenId::Nep141(_)))
+        .then(|| mt_token_id.to_string())
+}
```
pepe-anchor, line 143: must: mt_underlying_nep141 resolves nep245:<contract>:<mt_token_id> against a curated NEP-141 (e.g. wNEAR) whenever mt_token_id merely parses as nep141:<account>, with no check that contract is an actual trusted defuse vault, so anyone can deploy their own NEP-245 contract, mint a token_id literally spelled nep141:wrap.near, and have render_mt_withdraw (and, since resolve/seeded_decimals are shared, render_transfer and render_token_diff too) show it as a fully verified wNEAR amount with zero unverified-metadata caveat, since it resolves with TokenProvenance::Seed. The PR description calls this 'self-verifying' because an unrelated contract's token_id 'simply fails to parse,' but token_id is an arbitrary string fully controlled by whoever deploys the MT contract, so a malicious deployer can trivially choose one that does parse; this defeats the exact seeded/unattributed-metadata trust boundary the surrounding code (and its recent PR history: #454, #459, #478) was built to enforce. Scope the alias to a known set of trusted defuse instances (or require signed metadata for it) instead of inferring trust from the token_id spelling alone.

prasanna-anchorage (changes requested):
The gap this closes is real — `mt_withdraw` being the one registry-consulting intent that didn't consult the registry is a genuine inconsistency, and the `intent_consumes_token_registry` flip plus the `seeded_decimals` gap-fill extension are both right. `make lint` clean, `make test` 43 suites / 2227 tests, 0 failures.

**But the resolution is keyed on a string the sender fully controls, and the contract half is discarded.** I think this has to be fixed before it lands.

## The finding

`mt_underlying_nep141` throws away the contract:

```rust
let mt_token_id = asset_id.strip_prefix("nep245:")?.split_once(':')?.1;
matches!(mt_token_id.parse::<TokenId>(), Ok(TokenId::Nep141(_))).then(|| mt_token_id.to_string())
```

`render_mt_withdraw` builds the asset id as `format!("nep245:{}:{id}", w.token)` (`render.rs:562`), where **both** `w.token` and `id` come straight out of the intent. So the inner `nep141:...` claim is honoured no matter which contract makes it. Probed on head `cd2a5439`:

```
nep245:defuse.near:nep141:wrap.near       -> wNEAR / 24 decimals / Seed
nep245:evil.near:nep141:wrap.near         -> wNEAR / 24 decimals / Seed
nep245:attacker-mt.near:nep141:wrap.near  -> wNEAR / 24 decimals / Seed
```

End to end through the render path:

```
token=defuse.near -> Token=Text(defuse.near) | To=Text(alice.near) | MT Token=Text(nep141:wrap.near) | Amount=AmountV2(2 wNEAR)
token=evil.near   -> Token=Text(evil.near)   | To=Text(alice.near) | MT Token=Text(nep141:wrap.near) | Amount=AmountV2(2 wNEAR)
```

Anyone can deploy a NEP-245 contract whose `token_id` is the literal string `nep141:wrap.near`. Its balance then renders to the signer as `2 wNEAR`, at wNEAR's 24 decimals, with `provenance: Seed` — the most trusted value, so **no `unverified-token-metadata` diagnostic fires either**. The only thing distinguishing the two lines is a separate `Token` field the signer has to notice and know to distrust, while the `MT Token` field actively corroborates the wNEAR reading.

**And it is not confined to `mt_withdraw`.** `resolve` is the single shared path (`render.rs:120`, `token_amount_field`), so every caller inherits the fallback. In `token_diff` the presentation is strictly worse, because no contract is displayed at all:

```
{"intent":"token_diff","diff":{"nep245:evil.near:nep141:wrap.near":"-2000000000000000000000000"}}
  -> Send=AmountV2(2 wNEAR)
```

That is the whole rendering. The PR body describes this as an `mt_withdraw` fix, so this path looks unintended rather than accepted.

## Why the stated safety argument doesn't hold

The description says an independent contract's convention "simply fails to parse and falls through to unresolved ... rather than needing a curated allowlist to keep it out." That is an honest-mistake analysis: it holds for a contract that happens to use a different convention, and an adversary picks the convention deliberately. Self-verifying would mean the *string* proves the relationship; here it only asserts it.

The file already takes the opposite stance one function up, and names this exact threat — `token_amount_field`'s docstring (`render.rs:110-113`): "Resolution runs against the raw id, because stripping first would let a crafted id collapse onto a seeded one (`nep141:wrap\u{7f}.near` -> `nep141:wrap.near`) and borrow that token's symbol and decimals." A homoglyph that borrows wNEAR's metadata was worth defending against; `nep245:evil.near:` + the real id reaches the same outcome through the front door.

## The tests can't see it, and one docstring claims the opposite

All four positive cases use `defuse.near` and vary only the token_id; the negative case (`nep245:market.near:gold-sword-42`) varies contract *and* token_id together. So nothing distinguishes "we trust the inner id" from "we trust contract + inner id" — the suite passes identically under either. The adversarial cell of that 2x2 (unrelated contract, *related* inner id) is the one that is missing.

`mt_withdraw_with_an_unrelated_token_id_is_unresolved` says in its own docstring: "An MT withdraw on an independent, unrelated multi-token contract ... must still render honestly as unresolved, not silently misattribute an amount." Under the probe above, an independent unrelated contract does silently misattribute. The test documents the intended invariant and then only exercises the case where the token_id gives it away for free.

## What I think it needs

Bind the contract. The convention only holds when a defuse-style contract represents *its own* NEP-141 balance, so the resolvable case is the one where the contract is the intents/verifying contract itself — compare against `INTENTS_RECEIVER` (`convert.rs:332`), or thread the envelope's `verifying_contract` down if you'd rather not hardcode. `nep245:<intents contract>:nep141:wrap.near` resolves; `nep245:anything-else:nep141:wrap.near` stays unresolved, which is exactly today's behaviour and the honest answer. That also makes the existing `mt_specific_override_beats_the_underlying_nep141` precedence safe rather than incidental: a request-scoped override remains the way to name a third-party MT asset, and it carries provenance the signer is told about.

If you'd rather keep it contract-agnostic, then the borrowed metadata shouldn't come back as `Seed` — it's an inference, not curation, and it should carry a provenance that trips `unverified_cause()` so the signer gets the caveat. I'd still prefer the binding.

Either way please add the missing 2x2 cell as a test, at both layers (`resolve` and `render_intent`), plus a `token_diff` case — that path is the one with no contract on screen.

## Smaller things

1. **The body is stale on its own base.** It says "Stacked on #439 (this bases on `near-f-diagnostics`)", but #439 merged 09-03 and the base is now `shahankhatchadourian/near-h-intent-type` (#460's branch). Worth updating since it's the first thing a reviewer checks.
2. **#460, its base, is approved and has been mergeable for a day** — landing that first would let this retarget to `main` and shrink the review surface further (it already dropped from +1146/-95 to +197/-15 after the restack). Consider enabling auto-merge on #460.

src/chain_parsers/visualsign-near/src/presets/intents/tokens.rs
prasanna-anchorage: Credit where it belongs: @pepe-anchor raised this 29 hours before I did (09-09 15:39 vs my 09-10 20:14), and my review doesn't mention it because I posted without reading the thread first. Same mistake I made on #484 two days ago, so it's a habit rather than a slip — reading existing comments before reviewing is the fix.

His comment already had the whole finding, including the two parts I thought were mine to add: that `resolve`/`seeded_decimals` being shared drags `render_transfer` and `render_token_diff` in too, and that it resolves at `TokenProvenance::Seed` so no unverified-metadata caveat fires. His framing of *why* is sharper than mine, too — it defeats the seeded/unattributed-metadata trust boundary that #454, #459 and #478 were built to enforce, which puts it in the context of the whole recent arc rather than treating it as one function's bug.

What my pass adds on top, so this isn't pure duplication:

- **Probes rather than reading.** Empirically, `nep245:evil.near:nep141:wrap.near` and `nep245:attacker-mt.near:nep141:wrap.near` both resolve to `wNEAR / 24 decimals / Seed`, and end-to-end `token=evil.near` renders `Amount=AmountV2(2 wNEAR)`. For `token_diff` the whole rendering is `Send=AmountV2(2 wNEAR)` — no contract on screen at all, which makes that path strictly worse than `mt_withdraw` and is worth stating separately.
- **The file already argues the opposite one function up.** `token_amount_field`'s docstring (`render.rs:110-113`) says resolution deliberately runs against the raw id because stripping first "would let a crafted id collapse onto a seeded one (`nep141:wrap\u{7f}.near`) and borrow that token's symbol and decimals". A homoglyph was worth defending against; `nep245:evil.near:` + the real id reaches the same end through the front door.
- **Why the tests can't catch it.** All four positive cases use `defuse.near` and vary only the token_id; the negative case varies contract *and* token_id together. The suite passes identically whether the code trusts the inner id alone or contract + inner id. And `mt_withdraw_with_an_unrelated_token_id_is_unresolved` states the violated invariant in its own docstring, then only exercises the case where the token_id gives it away for free.

On the fix, pepe's "known set of trusted defuse instances, or require signed metadata" and my "bind to `INTENTS_RECEIVER` / thread `verifying_contract` down" are the same shape; his generalises better if more than one defuse instance ever matters. Either way the fallback-to-`Seed` has to go: if the alias is kept contract-agnostic, the borrowed metadata is an inference rather than curation and must carry a provenance that trips `unverified_cause()`.

Two independent reviewers reaching the same conclusion from different directions is worth something on its own, but it shouldn't read as two findings.

shahan-khatchadourian-anchorage: Fixed in c6c539b7: `mt_underlying_nep141` now also requires the parsed `contract` half to equal `INTENTS_RECEIVER` ("intents.near") before borrowing the inner id's metadata, per your "scope the alias to a known set of trusted defuse instances" ask (going with the single hardcoded contract this codebase already trusts, rather than an allowlist, since that's the only instance anything here currently recognizes).

Added the missing 2x2 cell (untrusted contract + a token_id that *does* round-trip as a seeded NEP-141 id) as a test at both the `resolve`/`seeded_decimals` layer and the `render_intent`/`render_token_diff` layer, plus the existing gap-fill regression test now exercises an MT id on the trusted contract specifically so it still catches an unattributed override smuggling the wrong decimals in.

shahan-khatchadourian-anchorage:
@prasanna-anchorage re your CHANGES_REQUESTED review: fixed in c6c539b7, replied on the inline thread with @pepe-anchor with the details. Went with binding `mt_underlying_nep141`'s `contract` half to `INTENTS_RECEIVER` (your and pepe's shared first preference) rather than threading `verifying_contract` down or downgrading provenance on a contract-agnostic alias.

Re-ran your three probes (`nep245:intents.near:nep141:wrap.near`, `nep245:evil.near:nep141:wrap.near`, `nep245:attacker-mt.near:nep141:wrap.near`) against the new unit tests: only the `intents.near` case resolves to wNEAR now, the other two render unresolved -- across `mt_withdraw` and the `token_diff` shape you flagged as having no contract on screen at all.

Also updated the PR description, which was stale on two counts: it still argued the token_id-alone framing your review corrected, and named a base (`near-h-intent-type`) that's since moved to `main`.

src/chain_parsers/visualsign-near/src/presets/intents/render.rs
```diff
@@ -1069,6 +1080,35 @@
...
+    fn token_diff_with_an_untrusted_mt_contract_is_unresolved() {
+        let intent = intent_from(
+            r#"{"intent":"token_diff","diff":{"nep245:evil.near:nep141:wrap.near":"-2000000000000000000000000"}}"#,
+        );
+        let fields = render_intent(&intent, &empty_reg()).expect("render");
+        assert!(
+            fields
+                .iter()
+                .all(|f| !matches!(f, SignablePayloadField::AmountV2 { .. })),
+            "expected no resolved AmountV2, got {fields:?}"
```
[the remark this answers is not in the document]
shahan-khatchadourian-anchorage: False positive -- CodeQL's taint path runs through account_control_warning's return type (SignablePayloadField), but the actual value at this line was a test's assert! message debug-formatting the parser's own rendered UI Fields output, not any sensitive data. Removed the redundant assert in b28a23b9 anyway (the following narrower match already covered the same check), so this line no longer exists.

shahan-khatchadourian-anchorage:
Ready for re-review @prasanna-anchorage and @pepe-anchor

prasanna-anchorage (approved):
Blocking finding closed, verified by live probe against a built `parser_cli` (not just `cargo test`), head `b28a23b9`.

`mt_underlying_nep141` now binds the parsed `contract` half to `INTENTS_RECEIVER` before borrowing the inner id's metadata — exactly the fix both pepe and I asked for.

Re-ran the three cases from my original finding against the actual CLI:

| Case | Result |
|---|---|
| `mt_withdraw`, `token=evil.near`, `token_ids=["nep141:wrap.near"]` | `2000000000000000000000000 (unresolved nep245:evil.near:nep141:wrap.near)` |
| `mt_withdraw`, `token=intents.near` (positive control), same token_id | `2 wNEAR` -- resolves correctly, so the binding isn't over-strict |
| `token_diff`, `nep245:evil.near:nep141:wrap.near` -- the path with no contract shown at all | `2000000000000000000000000 (unresolved nep245:evil.near:nep141:wrap.near)` |

Before this fix, cases 1 and 3 both rendered `2 wNEAR` at `Seed` provenance with no caveat -- an unrelated NEP-245 deployer could borrow wNEAR's symbol/decimals for their own balance. Both are now honestly unresolved.

Also verified the 2x2 test cell that was missing before (untrusted contract + a token_id that *does* round-trip as a seeded NEP-141 id) is now covered at both the `resolve`/`seeded_decimals` layer and the `render_intent`/`token_diff` layer, per the diff.

`make lint` and `make test` both exit 0, no failures.

PR #481, opened 2026-09-03
test(parser_app): guard the dev Makefile target against argument drift

prasanna-anchorage:
## Why

Nothing in CI runs the dev `make` targets. The `ubuntu` job is `make -C src generated`, `lint`, `test`, `narrow-build-check` — so an argument `parser_app`'s CLI stops accepting leaves `make -C src parser_app` broken with every check green.

That is not hypothetical. The target passed `--usock` for some time after `751118bd` replaced it with `--host-ip`/`--host-port`, and panicked before reaching any flag the existing `cli.rs` tests cover:

```
thread 'main' panicked at parser/app/src/cli.rs:31:33:
Parser: invalid CLI args: found --usock, which was not an expected argument
```

It took someone running the target by hand to find (#454), and then a rebase reintroduced it on another branch (#450). Twice, silently, because no automated check looks.

## What

One test, `parser_app`'s `cli.rs`. It extracts the `parser_app` target's recipe arguments from `src/Makefile`, substitutes `$(VAR)` references, and feeds them to the real `OptionsParser::<ParserParser>` — so it catches drift from either side: a flag removed from the CLI, or a stale flag left in the recipe.

It then resolves the ABI trust posture too. Parsing alone would still pass a target that had lost `--accept-unsigned-abis`, which `Cli::execute` refuses to start without.

Two deliberate choices worth flagging for review:

- **`include_str!` rather than a runtime read**, so the Makefile is a compile-time dependency. Moving or renaming it is a build failure, not a guard that quietly stops checking anything.
- **The structural assertions pin only "some arguments, at least one a flag" — not *which* flags.** They exist because an unreadable recipe yields an empty argument list, which parses cleanly and proves nothing. Pinning specific flag names instead would make a legitimate rename fail with a misleading "the extractor broke" message *before* reaching the parse. I wrote it that way first and it misdiagnosed the `--usock` case, so the ordering here is deliberate: parse first, and let the failure name the offending argument.

## Scope

`parser_app` only, deliberately. Of the five dev run targets, `grpc-server` and `parser_gateway` pass no arguments, `parser_enclave` passes positionals, and `parser_host` is qos_core's binary whose CLI is not ours to test. `parser_app` is the only one with flags this repo owns, and the only one that has actually rotted.

## Verification

Mutated the recipe locally to confirm the guard fails for the right reason in each direction, rather than only that it passes as written:

| Recipe state | Result |
|---|---|
| as written | passes |
| `--usock` restored | fails: `UnexpectedInput("--usock")`, naming the argument |
| `--accept-unsigned-abis` dropped | fails on the posture check |
| target renamed out from under it | fails structurally, not vacuously |

`make -C src fmt`, `lint` (including `-p parser_app --no-default-features`, so `narrow-build-check` is unaffected) and `test` (39 suites, 1674 tests, 0 failures) all green.

## Base

Stacked on #454 because that PR carries the `--usock` fix; on `main` this guard would correctly fail, since `main`'s target still has the broken flag. GitHub will retarget this to `main` when #454 merges. #454's own approval is untouched — this is a child branch, not a push to `prs-556-parser-app`.

### DIFF
--- a/src/parser/app/src/cli.rs
+++ b/src/parser/app/src/cli.rs
@@ -273,4 +273,120 @@ mod tests {
         let opts = parse(&["--accept-unsigned-abis"]).expect("valid args");
         assert_eq!(opts.host_addr().to_string(), "127.0.0.1:3000");
     }
+
+    /// Extract one target's recipe arguments from a Makefile, as the shell would
+    /// receive them after `cargo run --bin <x> --`.
+    ///
+    /// Deliberately simple: find the target line, take the tab-indented recipe
+    /// lines under it, drop everything through the `--` that separates cargo's own
+    /// arguments from the binary's, and substitute `$(VAR)` references. It does not
+    /// try to be `make` -- if the recipe ever grows conditionals or a `$(shell ...)`
+    /// call this stops reflecting reality, which is what the structural assertions
+    /// in the caller are there to catch: they fail on an unreadable recipe rather
+    /// than letting an empty argument list parse cleanly and prove nothing.
+    fn makefile_target_args(makefile: &str, target: &str) -> Vec<String> {
+        let mut recipe = String::new();
+        let mut in_target = false;
+        for line in makefile.lines() {
+            if in_target {
+                // The recipe ends at the first line that is not tab-indented.
+                let Some(body) = line.strip_prefix('\t') else {
+                    break;
+                };
+                recipe.push(' ');
+                recipe.push_str(body.trim_end().trim_end_matches('\\'));
+            } else if line.starts_with(&format!("{target}:")) {
+                in_target = true;
+            }
+        }
+
+        // `--` separates cargo's arguments from the binary's; everything before it
+        // (`cargo run --bin parser_app`) is not the binary's argv.
+        let after_separator = recipe
+            .split_whitespace()
+            .skip_while(|token| *token != "--")
+            .skip(1);
+
+        after_separator.map(substitute_make_vars).collect()
+    }
+
+    /// Replace every `$(VAR)` in one token with a value the CLI will accept. The
+    /// two that have to look real are the ones `host_addr()` parses; anything else
+    /// is only carried as an opaque string.
+    fn substitute_make_vars(token: &str) -> String {
+        let mut out = String::with_capacity(token.len());
+        let mut rest = token;
+        while let Some(start) = rest.find("$(") {
+            out.push_str(&rest[..start]);
+            let after = &rest[start + 2..];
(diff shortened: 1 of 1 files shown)


shahan-khatchadourian-anchorage (dismissed):
Nice, focused addition — single file, a new test plus two small private helpers that guard `make parser_app`'s dev target against argument drift. I traced the Makefile-parsing logic against the actual `parser_app` target in `src/Makefile` line-by-line and it correctly extracts the argv; clippy is clean and the test passes. Good use of a real regression case (the `--usock`/`--host-ip`/`--host-port` flag rename) to justify why this guard is worth having, and it fills a real gap — nothing else in CI exercises this target's actual argv.

One non-blocking note for follow-up, not this PR:

The Makefile-recipe parsing (`makefile_target_args`/`substitute_make_vars`) is a bespoke text scraper — it doesn't handle conditionals, `$(shell ...)`, or more than one `--` separator. If the `parser_app` recipe ever grows any of those, this helper would likely mis-extract tokens silently rather than fail loudly, which undermines the drift-detection this test is meant to provide. Not asking for a rewrite here, just flagging it in case the recipe gets more complex later — a shared `PARSER_APP_ARGS`-style variable referenced by both the recipe and the test would make this airtight.

prasanna-anchorage:
Addressed Copilot's finding in `a352b462`: the guard now calls `host_addr()` after `abi_trust()`, so a malformed `--host-ip`/`--host-port` in the recipe is caught the same way a missing posture flag already was.

Verified by mutation, not just that it compiles:
- As-is: passes.
- Mutated the recipe's `--host-ip $(PARSER_HOST)` to a literal `--host-ip localhost`: now fails with `could not parse ip to IP v4: AddrParseError(Ipv4)` at `cli.rs:39`, confirming the new call actually catches it. Reverted after.

`cargo test -p parser_app --lib` 42/42, `cargo clippy -p parser_app --all-targets -- -D warnings` clean, `cargo fmt -p parser_app -- --check` clean.

No new test needed — this extends the existing guard's own assertion rather than adding a parallel one, since the thing being verified (the target's actual startup path) is the same.

prasanna-anchorage:
@shahan-khatchadourian-anchorage — your approval got dismissed by `a352b462` (the Copilot-thread fix above). Re-approval needed to unblock; the only change since your last review is that one-line `host_addr()` addition, mutation-verified in the comment above. Everything else you already reviewed is untouched.

shahan-khatchadourian-anchorage (approved):
LGTM. Thanks @prasanna-anchorage

Issue #485, opened 2026-09-08
ci: publish parser_cli as a release asset via draft-then-publish (immutable releases block the current flow)


prasanna-anchorage:
Follow-up to the revert of #480's release-asset half. The goal is still worth having — `parser_cli` is usable standalone with no Docker, so a static musl binary on the release page is a real convenience — but the current flow cannot deliver it.

## Why the straightforward version fails

```
HTTP 422: Cannot upload assets to an immutable release.
.../releases/384760827/assets?name=parser_cli-solana-v0.949.0-x86_64-unknown-linux-musl
```

Immutable releases are enabled on this repo. `release.yml` does:

```
gh release create "$TAG" --generate-notes
```

which creates **and publishes** in one call, and publishing freezes the release's assets. stagex then builds the image and runs `gh release upload` against something already frozen. This is structural, not a timing window — every release reports `immutable: true` and carries zero assets, so no release has ever had one.

Note the asymmetry that made the original step look safe: immutability freezes **assets and the tag**, but **not the body**. The adjacent TVC-details step mutates the same release with `gh release edit --notes-file` and works fine — v0.946.0, v0.903.0 and v0.929.0 all carry their mirrored TVC details. So the pattern the asset step copied happens to use the one release mutation that immutability still permits.

## What would actually work

**Draft, attach, then publish.** Sketch:

1. `release.yml`: `gh release create "$TAG" --generate-notes --draft`
2. `release.yml` dispatches stagex (already does)
3. stagex `parser_cli` job: extract the binary from the image and `gh release upload` — permitted, the release is still a draft
4. stagex `parser_app` job: mirror the TVC details with `gh release edit` — wants to land here too, pre-publish, so the published notes are complete on first publish rather than edited afterwards
5. something flips `--draft=false` once both matrix targets are done

## The hard parts, which are why this wasn't done inline

- **Who publishes, and when?** Steps 3 and 4 are different matrix targets in the same job; neither can know the other finished. Needs either a dependent job gated on the whole matrix, or a separate workflow triggered on stagex completion.
- **Failure leaves a draft.** If stagex fails after step 1, the release sits unpublished and the tag exists — `release.yml`'s existing `release_exists` retry logic assumes published releases and would need to understand drafts.
- **A draft release is invisible to consumers** for the duration of the stagex build (tens of minutes), which changes what `gh release view "$TAG"` returns for anything else keying off it.
[...]

PR #486, opened 2026-09-08
Revert "ci(cli): publish parser_cli as a release asset"

prasanna-anchorage:
Reverts the **release-asset half of #480 only**. The narrow feature-build check from the same PR stays — it earned its place within hours by catching a bug latent since #439 (fixed in #484).

## What happened

Immutable releases were enabled on this repo today, on the reasonable expectation that it would be a no-op. It isn't, for this one step: publishing a release now freezes its assets, and `release.yml:114` is

```
gh release create "$TAG" --generate-notes
```

which creates **and publishes** in the same call. By the time stagex has built the image and reaches `gh release upload`, the release can no longer accept an asset:

```
HTTP 422: Cannot upload assets to an immutable release.
.../releases/384760827/assets?name=parser_cli-solana-v0.949.0-x86_64-unknown-linux-musl
```

So this is **not a flaw in #480** — it was written against a repo where publishing didn't freeze assets, and the setting changed underneath it. It is also not the race its comment anticipates: the step says "the tag-dispatched stagex run created afterward picks it up," but that run gets the same 422, because the release is frozen rather than absent. Its guard (`if ! gh release view … skip`) covers absence, and `--clobber` cannot help.

## Why revert rather than fix forward

**Immutability stays** — that is the decision, confirmed. It is a supply-chain guarantee on a parser that sits on a signing path and is heading into external audit, and it is worth more than a download convenience.

Making the asset work under it needs the release created as a **draft**, assets attached, and only then published — which reorders work across `release.yml` and `stagex.yml` and wants the TVC notes edit to land pre-publish too. That is #485, filed with the design and the hard parts (who flips `--draft=false` when the two uploads are different matrix targets; what a mid-flight failure leaves behind given `release.yml`'s `release_exists` retry assumes published releases; draft invisibility for the length of the stagex build). Too much to land inline while `main` is red, hence: revert now, do it properly next.

## Worth recording: why nothing caught this

Immutability freezes **assets and the tag** but **not the body**. The adjacent TVC-details step mutates the same release with `gh release edit --notes-file` and still works fine — v0.946.0, v0.903.0 and v0.929.0 all carry their mirrored TVC details. So the asset step sat next to a release mutation that immutability happens to permit, and nothing in either diff signalled that one of them would break and the other wouldn't.

## Scope

One step removed, 27 lines, nothing else touched:

- `contents: write` stays — it predates #480 and its comment cites the TVC `gh release edit`.
- No docs referenced the asset (grepped `.md`/`.mdx`/`.yml`; only hits were the step's own lines).
- #480's other half — the `parser-cli-narrow-build-check` target, its `main.yml` job, the `cli_test.rs` change — untouched.
- YAML re-parsed after the edit; the `parser` job still has its remaining 7 steps in order.

## Note on `main`

`main` @ `5f44ab68` is red on **two independent** checks. This PR fixes `Build parser_cli (Solana)`. The `ubuntu` failure is a separate bug fixed by **#484**; both need to land.

### DIFF
--- a/.github/workflows/stagex.yml
+++ b/.github/workflows/stagex.yml
@@ -134,33 +134,6 @@ jobs:
           done
           docker image push --all-tags "ghcr.io/anchorageoss/${{ matrix.target.name }}"
 
-      - name: Upload CLI binary to release
-        # parser_cli is also usable standalone (no Docker required), so
-        # attach the static binary we already built as a release asset
-        # alongside the OCI image. images/parser_cli/Containerfile builds
-        # this image with only `--features solana` (see the "parser_cli
-        # (Solana)" matrix label above), so the asset name says so rather
-        # than implying it covers every chain parser_cli can be built with.
-        # Only meaningful on a release build -- SEMVER_TAG is empty for
-        # PR/branch builds, and (per the TVC step's comment above) a
-        # push-to-main run can compute a SEMVER_TAG before release.yml has
-        # actually created that release, so this is a no-op in that race;
-        # the tag-dispatched stagex run created afterward picks it up.
-        if: matrix.target.name == 'parser_cli' && steps.build.outputs.semver_tag != ''
-        env:
-          SEMVER_TAG: ${{ steps.build.outputs.semver_tag }}
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          if ! gh release view "$SEMVER_TAG" --repo "${GITHUB_REPOSITORY}" >/dev/null 2>&1; then
-            echo "No release $SEMVER_TAG yet; skipping CLI binary upload."
-            exit 0
-          fi
-          docker create --name tmp-extract-cli "ghcr.io/anchorageoss/parser_cli:${SEMVER_TAG}" /bin/true
-          asset="parser_cli-solana-${SEMVER_TAG}-x86_64-unknown-linux-musl"
-          docker cp tmp-extract-cli:/usr/local/bin/parser_cli "$asset"
-          docker rm tmp-extract-cli
-          gh release upload "$SEMVER_TAG" "$asset" --clobber --repo "${GITHUB_REPOSITORY}"
-
       - name: TVC deployment details
         if: matrix.target.name == 'parser_app'
         shell: bash


prasanna-anchorage:
Requesting review from @shahan-khatchadourian-anchorage, @pepe-anchor, @rituparna-mandal-anchor — whoever gets to it first. `main` has been red for seven hours and this is the second of the two fixes.

**It's a 27-line deletion of one workflow step, and nothing else.** Rebased onto #484's merge (`ef7cdc63`) so CI here is now genuinely green rather than inheriting `main`'s red. Copilot recommended approval with zero comments.

**Why it needs reverting rather than fixing forward:** immutable releases were enabled on the repo today, on the expectation it'd be a no-op. It isn't for this one step — `release.yml:114` is `gh release create "$TAG" --generate-notes`, which publishes in the same call, and publishing now freezes the release's assets. So by the time stagex has built the image and reaches `gh release upload`, the asset can't be attached:

```
HTTP 422: Cannot upload assets to an immutable release.
```

**This is not a flaw in #480** — that step was correct when written; the repo setting changed underneath it. It's also not the race its own comment anticipates ("the tag-dispatched stagex run created afterward picks it up") — that run gets the same 422, because the release is *frozen*, not absent.

**What to check, if you want to check anything:**
- `contents: write` stays — it predates #480 and the TVC `gh release edit` step needs it.
- #480's other half (`parser-cli-narrow-build-check`, its `main.yml` job) is untouched. That half earned its place: it's what surfaced the fixture bug #484 just fixed.
- Nothing else referenced the asset (grepped `.md`/`.mdx`/`.yml`).

Immutability stays — it's a supply-chain guarantee on a parser sitting on a signing path and heading into audit, and it's worth more than a direct-download convenience. #485 tracks shipping the binary properly via draft-then-publish.

One quirk worth knowing for whoever reviews: immutability freezes **assets and the tag** but **not the body**, which is why the adjacent TVC-details step's `gh release edit --notes-file` still works fine (v0.946.0, v0.903.0, v0.929.0 all carry their mirrored details). That's why nothing in either diff signalled that one of these two release mutations would break and the other wouldn't.

pepe-anchor (approved):

shahan-khatchadourian-anchorage (approved):
LGTM. Thanks @prasanna-anchorage for finding the root cause and also for the fix.

PR #487, opened 2026-09-11
ci: cover the tvc-deploy workspace on pull requests

prasanna-anchorage:
Two CI-coverage gaps found while reviewing #455. Neither was a defect in that PR — both are about what CI runs.

## 1. `tools/tvc-deploy` has no pull-request coverage at all

It is its own cargo workspace, so `src/Cargo.toml`'s workspace lints and `make -C src lint|test` cannot reach it. The repo's only `cargo test --manifest-path tools/tvc-deploy/Cargo.toml` lives inside `tvc-deploy.yml`'s deploy job, gated on:

```
if: github.event_name == 'workflow_dispatch' || github.event.label.name == 'tvc-deploy-test'
```

So a PR touching the deploy helper gets nothing. **Not hypothetical: 56 tests already on `main` never ran in PR CI**, and #455 took the crate to 65 while adding 258 lines of posture logic — the XOR, on-curve pubkey validation, `pivotArgs` composition — that CI never exercised. (#455 carried only the `CI` label; its checks were `ubuntu`/`CodeQL`/`Analyze (*)`/cla/labeler.)

It is the same class as the two gaps this repo has already paid for: the parser_cli narrow build (#480, which surfaced a fixture bug latent for five days the moment it started running) and `make parser_app` argument drift (#481, after that target broke silently twice).

I checked the scope rather than assuming: the repo has three cargo workspaces — `src/` (covered by `main.yml`), `src/chain_parsers/visualsign-solana/fuzz/` (covered by the pull_request-triggered `fuzz-solana.yml`), and `tools/tvc-deploy/`. This is the only uncovered one.

`--locked` is deliberate, for the same reason the narrow-build checks use it: the crate keeps its own `Cargo.lock`, so without it a stale lock gets silently updated in CI instead of failing. Verified it passes both before and after #455, so the step is safe regardless of merge order. Cost is negligible — the crate builds in seconds, tests run in 0.01s.

## 2. The one end-to-end deploy validation always took the permissive posture, silently

#455 added a fail-closed guard for `workflow_dispatch`. But on a `pull_request` labeled `tvc-deploy-test`, the `abi_signer_pubkey` input does not exist — inputs are dispatch-only — so `ABI_SIGNER_PUBKEY` was empty, the guard (ANDed with the event name) did not fire, and the next branch mapped empty to `--accept-unsigned-abis`.

**This was never a production exposure** — `APP_ID` falls back to `vars.TVC_TEST_APP_ID`, the dev test app. What it cost was validation: the label-triggered deploy is the only automated run of tvc-deploy → `pivotArgs` → signed manifest → `parser_app` startup, and it always took the branch that requires no signer. So require-signed — the posture where a wrong key or a bad `pivotArgs` composition makes `parser_app` refuse to start, i.e. the one with a failure mode worth catching — got no end-to-end coverage from the mechanism built to validate deploys. It is also the exact behaviour `abi_signer_pubkey`'s own description gives as the reason it has no default.

Resolution order now:

| event | input | `TVC_TEST_ABI_SIGNER` | result |
|---|---|---|---|
| `workflow_dispatch` | blank | unset **or set** | **exit 1** (unchanged) |
| `workflow_dispatch` | `unsigned` | — | `--accept-unsigned-abis` |
| `workflow_dispatch` | hex key | — | `--accept-signatures-from-pubkey <key>` |
| `pull_request` (label) | blank | set | `--accept-signatures-from-pubkey <test signer>` |
| `pull_request` (label) | blank | unset | `--accept-unsigned-abis` + `::notice::` |
| `pull_request` (label) | `unsigned` / hex key | — | explicit input wins |

The dispatch guard stays **first** on purpose: `app_id` is free text there, so a blank posture must not resolve to anything — not even to the test signer, which would push a test key at whatever app was named. I tested that specific case because I designed for it: dispatch + blank still exits 1 even when the variable is set.

Nothing breaks with no repo variable configured — the label path keeps working, it just announces which posture it is exercising instead of falling through silently. Setting `TVC_TEST_ABI_SIGNER` is what upgrades it to validating the strict path.

## Verification

- Drove the extracted shell through **all eight input × event combinations** (table above, plus the dispatch+blank+variable-set case).
- Both workflows re-parse as YAML; `tvc-deploy.yml`'s triggers and the deploy job's `if` are unchanged.
- The new step's own commands pass: `cargo clippy --locked ... -D warnings` exit 0, `cargo test --locked ...` 65 passed.
- Left `tvc-deploy.yml`'s own test step in place deliberately — it is still the only pre-deploy gate, and running the tests twice on a labeled PR is cheaper than a deploy built from untested code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,3 +64,19 @@ jobs:
         run: make -C src narrow-build-check
       - name: Check parser_cli narrow feature builds
         run: make -C src parser-cli-narrow-build-check
+
+      # `tools/tvc-deploy` is its own cargo workspace, so `src/Cargo.toml`'s
+      # workspace lints and `make -C src lint|test` cannot reach it, and no
+      # other pull_request-triggered workflow builds it. Its tests ran only
+      # inside `tvc-deploy.yml`'s deploy job, which is gated on
+      # `workflow_dispatch` or the `tvc-deploy-test` label -- so a PR touching
+      # the deploy helper had no CI coverage of it at all.
+      #
+      # `--locked` for the same reason the narrow-build checks use it: this
+      # crate keeps its own Cargo.lock, so without it a stale lock is silently
+      # updated in CI instead of failing.
+      - name: Lint and test the deploy helper (separate workspace)
+        run: |
+          cargo clippy --locked --manifest-path tools/tvc-deploy/Cargo.toml \
+            --all-targets -- -D warnings
+          cargo test --locked --manifest-path tools/tvc-deploy/Cargo.toml
--- a/.github/workflows/tvc-deploy.yml
+++ b/.github/workflows/tvc-deploy.yml
@@ -119,15 +119,34 @@ jobs:
           HOST_IP: ${{ inputs.host_ip || '0.0.0.0' }}
           HOST_PORT: ${{ inputs.host_port || '3000' }}
           ABI_SIGNER_PUBKEY: ${{ inputs.abi_signer_pubkey || '' }}
+          TVC_TEST_ABI_SIGNER: ${{ vars.TVC_TEST_ABI_SIGNER }}
         run: |
+          # Resolved in order. The dispatch guard stays first on purpose:
+          # `app_id` is free text there, so a blank posture must not resolve to
+          # anything -- not even to TVC_TEST_ABI_SIGNER below, which would push
+          # a test key at whatever app was named.
           if [ -z "$ABI_SIGNER_PUBKEY" ] && [ "$GITHUB_EVENT_NAME" = "workflow_dispatch" ]; then
             echo "::error::abi_signer_pubkey is required for workflow_dispatch: pass a hex secp256k1 signer pubkey, or the literal 'unsigned' to explicitly deploy --accept-unsigned-abis."
             exit 1
-          fi
-          if [ "$ABI_SIGNER_PUBKEY" = "unsigned" ] || [ -z "$ABI_SIGNER_PUBKEY" ]; then
+          elif [ "$ABI_SIGNER_PUBKEY" = "unsigned" ]; then
             abi_trust_args=(--accept-unsigned-abis)
-          else
+          elif [ -n "$ABI_SIGNER_PUBKEY" ]; then
             abi_trust_args=(--accept-signatures-from-pubkey "$ABI_SIGNER_PUBKEY")
+          elif [ -n "$TVC_TEST_ABI_SIGNER" ]; then
+            # Label-triggered validation deploy with a test signer configured.
+            # Exercise require-signed: that is the posture where a wrong key or
+            # a bad pivotArgs composition makes parser_app refuse to start, so
+            # it is the one worth validating end to end. The permissive path
+            # cannot fail this way, so validating it proves less.
+            abi_trust_args=(--accept-signatures-from-pubkey "$TVC_TEST_ABI_SIGNER")
+          else
+            # Keeps the label path working with no repo variable set, but says
+            # so instead of falling through silently. A blank input previously
+            # took the permissive branch with no signal -- the behaviour
+            # abi_signer_pubkey's own description gives as the reason it has no
+            # default.
+            echo "::notice::no ABI trust posture given and TVC_TEST_ABI_SIGNER is unset; this validation deploy exercises --accept-unsigned-abis only. Set the TVC_TEST_ABI_SIGNER repo variable to validate require-signed instead."
+            abi_trust_args=(--accept-unsigned-abis)
           fi
           ./tools/tvc-deploy/target/release/tvc-deploy deploy \
             --app-id "$APP_ID" \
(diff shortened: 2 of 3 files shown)


prasanna-anchorage:
Replaced the `TVC_TEST_ABI_SIGNER` repo variable with a **derived** signer in `8fa0d385`. The variable was the weak part of this PR: until an admin set it, the label-triggered deploy would keep validating only the permissive posture, and nothing would fail to say so. A follow-up item that gates the value of the change isn't much better than the gap it replaced.

The repo already had a deterministic answer. `visualsign-ethereum`'s `CLI_DEV_SIGNING_KEY_SEED` is a committed constant and `parser_cli` already signs its own ABI files with that key, so its **public** half is derivable with no secret, no variable and no admin action:

```
0424653eac434488002cc06bbfb7f10fe18991e35f9fe4302dbea6d2353dc0ab1c119fc5009a032aa9fe47f5e149bb8442f71f884ccb516590686d8ff6ab91c613
```

Verified it through `tvc-deploy`'s own `validate_signer_pubkey` rather than eyeballing the shape — accepted: 65 bytes, SEC1 `04`, on the curve. So the label path now **always** exercises require-signed, and the `::notice::` apologising for the permissive fallback is gone with it.

**The duplication needed guarding, and that's the part I'd look at closely.** A workflow can't import a Rust constant, so the hex is copied there. `dev_abi_signer_pubkey_is_pinned` in `visualsign-ethereum` fails if the seed is ever rotated, and prints the replacement value to paste in. Without it the drift would be silent *and* green: the post-deploy smoke is a Solana V0+ALT transaction, which no Ethereum ABI posture can touch, so CI would keep passing while deploying a posture that allowlists a key nothing signs with any more. Mutation-checked — changing the seed to `[0x43; 32]` fails it with the correct new value in the message.

Scope is deliberately narrow:

- Only the **label** path takes the derived key, and only because it deploys to the dev TEST app (`vars.TVC_TEST_APP_ID`). The comment says plainly not to point a real app at a well-known dev key.
- `workflow_dispatch` is untouched and still fail-closed **first** — a blank posture there resolves to nothing, not even to this key, because `app_id` is free text. Re-tested that specific case: dispatch + blank still exits 1 even when the override variable is set.
- `TVC_TEST_ABI_SIGNER` survives as an override for anyone who wants a different signer.

| event | input | override var | result |
|---|---|---|---|
| `workflow_dispatch` | blank | unset **or set** | **exit 1** |
| `workflow_dispatch` | `unsigned` | — | `--accept-unsigned-abis` |
| `workflow_dispatch` | hex key | — | that key |
| `pull_request` (label) | blank | unset | **derived dev key** (require-signed) |
| `pull_request` (label) | blank | set | override |
| `pull_request` (label) | `unsigned` | — | `--accept-unsigned-abis` |
| `pull_request` (label) | hex key | set | input wins |

All eight combinations re-verified by extracting the shell and driving it directly. `make lint` 0, `make test` 0 (44 suites / 2223 tests), tvc-deploy clippy 0.

Net effect on this PR: no follow-up action is required of anyone for the require-signed path to start being validated. It starts on merge.

prasanna-anchorage:
@shahan-khatchadourian-anchorage — bumping this, it's had review requested since 09-11 with zero human eyes on it (only Copilot commented). Small scope: CI coverage for the `tools/tvc-deploy` workspace (currently untested in PR CI at all) plus closing the gap where the `tvc-deploy-test` label path always exercised the permissive ABI posture silently. CI green, `MERGEABLE`. Would appreciate a look when you get a chance.

shahan-khatchadourian-anchorage:
Non-blocking: the new `tools/tvc-deploy` clippy/test step in `.github/workflows/main.yml:78-82` doesn't pin a Rust toolchain the way the rest of the repo does. `tools/tvc-deploy/Cargo.toml` declares its own `[workspace]` but there's no `tools/tvc-deploy/rust-toolchain.toml`, and the step runs with cwd at the repo root, which also has none (only `src/rust-toolchain.toml` pins `1.88.0`). Rustup's toolchain-file lookup walks up from cwd, so from the repo root it finds nothing and falls back to whatever `actions-rust-lang/setup-rust-toolchain` installs by default (`stable`) rather than `1.88.0`. `make -C src lint` avoids this because `-C src` puts cargo's cwd inside the directory that owns the pin; this step doesn't.

Consequence: `-D warnings` here can start failing from clippy lint churn on a future stable release, independent of any actual change in `tools/tvc-deploy`, and won't reproduce locally via `make -C src lint` since that stays pinned to `1.88.0`.

Fix: add a `tools/tvc-deploy/rust-toolchain.toml` pinning `1.88.0` to match `src/rust-toolchain.toml`.

shahan-khatchadourian-anchorage (approved):
Approving. This closes a real coverage gap cleanly: `tools/tvc-deploy` had no PR-time lint/test coverage at all (only exercised inside the label/dispatch-gated deploy job), and the label-triggered validation deploy was silently validating only the permissive ABI posture until someone set a repo variable — the derived-signer fix in `8fa0d385` closes that without needing any admin action, and it's backed by a mutation-tested pinning test (`dev_abi_signer_pubkey_is_pinned`) so the hex duplicated into the workflow can't drift silently. The `workflow_dispatch` fail-closed guard staying first in the if/elif chain, ahead of the new label-path fallback, is the right call given `app_id` is free text there.

One non-blocking item, already left as a separate comment: the new `tools/tvc-deploy` clippy/test step in `main.yml` doesn't pin a Rust toolchain the way `make -C src lint` does, so it's exposed to future clippy-version drift independent of this PR.

Heads up for sequencing: #456 (draft, @pepe-anchor) touches the same `tvc-deploy.yml` ABI-posture block but branched off before this PR's changes landed, so it'll need a rebase against whichever of the two merges second.

PR #488, opened 2026-09-16
feat(cli): step a directory of payloads from one URL

shahan-khatchadourian-anchorage:
## What this adds

`serve` gives every file its own link, which works in a browser but not in a viewer that can be handed a URL and nothing else. Walking fourteen payloads through such a client meant pasting fourteen links.

| Route | Returns |
| --- | --- |
| `/next` | one payload per call, advancing a cursor and wrapping at the end |
| `/reset` | cursor back to the first entry |

So refetching a single address steps through the whole directory.

Each response prepends a `diagnostic` field, because a `SignablePayload` carries nothing naming its source:

```json
{ "Diagnostic": { "Domain": "parser_cli-serve", "Level": "ok",
                  "Message": "step 3 of 14: 03-near-register.json",
                  "Rule": "serve::step" },
  "FallbackText": "ok: step 3 of 14: 03-near-register.json",
  "Label": "step", "Type": "diagnostic" }
```

`?bare=true` returns the payload untouched.

`--host` sets the bind address and **defaults to `127.0.0.1`, unchanged**. A client on a separate device needs a non-loopback bind; making it an argument keeps that something the operator states rather than inherits. There is still no auth and no TLS.

## Decisions worth a reviewer's attention

1. **Both routes are GETs that mutate.** The client being served can be given a URL and nothing else, so there is no POST available to it.
2. **The cursor is per server, not per session.** Two viewers stepping at once interleave. Per-session state would need a cookie or token, which the same constraint rules out.
3. **Only entries that decode are in the rotation.** An undecodable file has nothing to return, and a 422 mid-walk would stall a client with no way to skip.
4. **One fetch is one step**, so a client that requests twice per view — a prefetch, a retry, an uncached conditional request — advances twice. The step diagnostic names the file, so a skip is visible rather than silent, and `/reset` recovers.
5. **`/next` and `/reset` are literal segments**, matched ahead of the `/{*path}` wildcard, so files of those names are reachable only through `/api/file?path=`.
6. **`pick_step` returns the payload, not the entry**, so no caller re-checks that the entry decoded.

## Scope

`serve` is behind a non-default cargo feature (`default = ["solana", "ethereum", "near", "tron", "diagnostics"]`) and is not linked by `parser_app` or `parser_grpc_server`. This is a development tool; nothing shipped includes it.

## Testing

`cargo test -p parser_cli --features serve serve::` — 17 pass, 7 of them new: walk order and wrap, undecodable entries skipped and excluded from the count, the empty case, the pick carrying its payload, the diagnostic's shape and position, a non-payload left untouched, and the index advertising both routes.

`cargo clippy -p parser_cli --features serve --all-targets -- -D warnings` clean; `cargo fmt --check` clean.

Verified by hand against a directory of 14 VisualSign renders: the walk covers all 14 in sorted order and wraps to step 1 on the fifteenth call, `?bare=true` omits the step field, `/reset` returns `{"reset":true,"next_step":1}`, an unknown path is a 404, and the default bind still reports `Serving on http://127.0.0.1:47474`.


shahan-khatchadourian-anchorage:
Addressed in `e5da88e0`. All three were real, and the cache one was the most consequential — a cached `/next` would have frozen the walk silently, which is the failure the route exists to avoid.

**`/reset` and `/next` were cacheable.** Both answer `Cache-Control: no-store` now. The review flagged `/reset`; `/next` had the same problem and a worse consequence, so both are covered by one helper whose doc comment says why. Verified on the wire rather than asserted:

```
$ curl -sD - -o /dev/null http://127.0.0.1:47482/next  | grep -i cache-control
cache-control: no-store
$ curl -sD - -o /dev/null http://127.0.0.1:47482/reset | grep -i cache-control
cache-control: no-store
$ curl -sD - -o /dev/null http://127.0.0.1:47482/01-near-verify-wallet.json | grep -i cache-control
(none — the file and index routes are unchanged)
```

**The index linked shadowed names to the walk.** Correct, and I had documented the shadowing without fixing its consequence. `entry_href` now sends a root file named `next` or `reset` through `/api/file?path=`, so every entry in the index stays reachable; that route answers with the `{path, ok, payload}` envelope, which is the visible difference for those two names. Only a root name collides — `sub/next` reaches the wildcard as before, and there is a test asserting that.

**The routes had no test.** Also correct. Coverage stopped at `pick_step`, `with_step_diagnostic` and `render_html`, so the wiring was unverified. Five tests drive the handlers directly:

| test | covers |
| --- | --- |
| `next_walks_in_order_and_wraps` | first, second, wrap, and `no-store` on each |
| `next_prepends_the_step_diagnostic_unless_bare` | `?bare=true` parsing and the field's absence |
| `reset_returns_the_cursor_to_the_first_entry` | the shared cursor and the reset acknowledgement |
| `next_is_not_found_when_nothing_decodes` | 404 and its message on an empty rotation |
| `a_file_named_for_a_walk_route_links_through_api_file` | the reserved-name link, and that nesting does not collide |

22 tests pass, `cargo clippy --features serve --all-targets -- -D warnings` clean, `cargo fmt --check` clean. Clippy caught one thing on the way: a test helper took `&PathBuf` where `&Path` does.

`CLAUDE.md` now records both behaviours next to the route table, so neither has to be rediscovered from the code.

